PR #49, opened 2024-01-05
Fix doc links

ahouseholder:
Addressing a number of 404 links in docs


docs/howto/activitypub/index.md
ahouseholder, line 1: Here I was addressing a previous TODO and adding a brief explanation of how the ActivityStreams Vocabulary allows us to write "sentences" about CVD cases.

Issue #527, opened 2026-05-15
fix(test): integration demo suite takes 17+ minutes — polling helpers not patched


ahouseholder:
## Summary

The full CI test suite (`uv run pytest -m ""`) takes **17+ minutes** on GitHub
Actions (1036s observed in run
[#25924697111](https://github.com/CERTCC/Vultron/actions/runs/25924697111/job/76202676588)).
This makes the CI gate on the `main` branch untenable for merge throughput.

## Evidence

CI log timestamp gaps in the "Run full pytest suite" step:

| Progress | Timestamp | Elapsed since prior |
|---|---|---|
| 54% | 14:58:09 | normal |
| 56% | 14:59:22 | ~73s (2%) |
| 59% | 15:02:56 | ~214s (3%) |
| **62%** | **15:14:52** | **~716s (3%) ← culprit** |
| 65%–100% | 15:14:54–15:14:57 | <3s (rest) |

The entire 3-second burst at the end confirms the problem is localised to the
56–62% window, which corresponds to the integration/demo test modules.

## Root-Cause Hypothesis

`test/demo/conftest.py` patches `demo_utils.DEFAULT_WAIT_SECONDS = 0.0`,
which zeros out the `post_to_inbox_and_wait` delays. **However**, the
polling helpers in `vultron/demo/scenario/two_actor_demo.py` have their
own `time.sleep(poll_interval)` loops that are **not patched**:

| Helper | `timeout_seconds` default | `poll_interval` default |
|---|---|---|
| `wait_for_note_in_case` | 10.0 | 0.5 |
| `wait_for_case_participants` | 5.0 | 0.25 |
| `wait_for_finder_case` | 10.0 | 0.5 |
| `wait_for_finder_log_entry` | 15.0 | 0.5 |
| *(two more polling helpers ~L1201, ~L1268)* | 10.0 | 0.25–0.5 |

If the polled condition does not materialise quickly in the `TestClient`
environment, each call will burn up to its full `timeout_seconds`. With
35 test functions in `test_two_actor_demo.py` many of which call these
helpers, the cumulative wait time can easily reach 10–15 minutes.

The `x` markers visible at 59% and 62% in the CI log are NOTRUN xfail
tests (`test_three_actor_demo`, `test_multi_vendor_demo`,
`test_initialize_participant_demo`); those tests themselves don't run, but
the **surrounding passing tests** in that window are what's slow.

## Files to Investigate

[...]

PR #529, opened 2026-05-15
Add pytest-timeout guardrail: 30s default per-test ceiling

### LINKED ISSUE #528: Add pytest-timeout guardrail: 30s default per-test ceiling
## Summary

Add `pytest-timeout` as a dev dependency and configure a **30-second default
per-test timeout** (thread-based). This catches regressions where a code change
causes a test to explode its runtime, without requiring a suite-level abort timer.

## Motivation

The full test suite has been seen running 10+ minutes, blocking PRs. Any single
test that runs longer than 30 seconds is almost certainly hitting a performance
regression (the current slowest unit test is ~0.5s; the limit is 60× that).

## Approach

- Add `pytest-timeout` to `[dependency-groups] dev` in `pyproject.toml`
- Set `timeout = 30` and `timeout_method = "thread"` in
  `[tool.pytest.ini_options]`
- No proactive audit pass — let the guardrail trip in CI and fix failures as they arise
- Document the policy in `test/AGENTS.md`: fix the slow test first;
  `@pytest.mark.timeout(N)` is a last resort for genuinely unavoidable slow
  tests and requires a comment explaining why

## Acceptance Criteria

- [ ] `pytest-timeout` is in dev deps and `uv.lock` is updated
- [ ] `timeout = 30` and `timeout_method = "thread"` set in
  `[tool.pytest.ini_options]`
- [ ] Unit suite passes with no unexpected timeout trips
- [ ] `test/AGENTS.md` has a timeout policy section explaining the
  fix-first / override-as-last-resort policy

ahouseholder:
Closes #528

## Summary

Adds `pytest-timeout` as a dev dependency and configures a **30-second
default per-test timeout** (thread-based). Any test that runs longer than
30 seconds now fails immediately, bounding the suite even when integration
tests are included.

## Changes

- `pyproject.toml`: add `pytest-timeout>=0.5` to dev deps; add
  `timeout = 30` and `timeout_method = "thread"` to
  `[tool.pytest.ini_options]`
- `uv.lock`: pytest-timeout 2.4.0 installed
- `test/AGENTS.md`: new Per-Test Timeout Guardrail section documenting
  the fix-first policy and `@pytest.mark.timeout(N)` as last resort

## Validation

Unit suite: **2368 passed, 12 skipped in 26.29s** — no unexpected timeout
trips. All four linters (Black, flake8, mypy, pyright) pass clean.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,6 +117,8 @@ minversion = "6.0"
 #
 # See AGENTS.md "Testing" section for full guidance.
 addopts = "-ra -q -m 'not integration'"
+timeout = 30
+timeout_method = "thread"
 testpaths = [
     "test",
 ]
@@ -148,6 +150,7 @@ dev = [
     "pre-commit>=4.6.0",
     "pyright>=1.1.409",
     "pytest>=9.0.3",
+    "pytest-timeout>=0.5",
     "types-networkx>=3.6.1.20260513",
     "types-pyyaml>=6.0.12.20260510",
 ]
--- a/test/AGENTS.md
+++ b/test/AGENTS.md
@@ -160,6 +160,35 @@ behavior across backends (in-memory / tinydb) where reasonable.
 
 ---
 
+### Per-Test Timeout Guardrail
+
+Every test has a **30-second default timeout** enforced by `pytest-timeout`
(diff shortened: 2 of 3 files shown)


sei-ahouseholder:
Merged without waiting for pytest to complete because of #527

Issue #535, opened 2026-05-15
RFC: Replace `_mixins.py` 26-class mixin system with direct `@property` declarations on event subclasses


ahouseholder:
## Problem

`vultron/core/models/events/_mixins.py` (311 lines) contains 26 nearly-identical
mixin classes. Each provides two `@property` aliases that rename a generic slot
on `VultronEvent` to a domain-specific name — for example,
`_ObjectIsReportMixin.report` aliases `event.object_`. The pattern is a
7-slot × N-domain-type matrix encoded as handwritten class boilerplate:

- 7 generic slots (`object_`, `inner_object`, `target`, `context`, `origin`,
  `inner_target`, `inner_context`)
- 9 domain types (`VultronReport`, `VultronCase`, `VultronEmbargoEvent`,
  `VultronNote`, `VultronParticipant`, `VultronParticipantStatus`,
  `VultronCaseStatus`, `VultronActivity`, `VultronObject`)
- 26 classes, each 8–14 lines of identical cast-based `@property` boilerplate

**Coupling friction**: Adding a new domain type requires 2–3 new mixin classes;
adding a new slot requires N new mixin classes. The interface is nearly as
complex as the implementation — a developer must scan 311 lines in a dedicated
file to understand what amounts to a name-alias lookup table.

**Testability**: There are no meaningful behaviors to test in `_mixins.py`
itself — each class is a cast-wrapped `getattr`. The test burden falls entirely
on callers rather than the mechanism.

## Proposed Interface

Replace all 26 mixin classes with direct `@property` declarations on each event
subclass, and delete `_mixins.py`.

```python
# Before — event class inherits from a mixin
class CreateReportReceivedEvent(_ObjectIsReportMixin, VultronEvent):
    semantic_type: Literal[MessageSemantics.CREATE_REPORT] = ...

# After — event class declares its own properties directly
class CreateReportReceivedEvent(VultronEvent):
    semantic_type: Literal[MessageSemantics.CREATE_REPORT] = ...

    @property
    def report_id(self) -> str | None:
        return self.object_id

    @property
    def report(self) -> VultronReport | None:
        return cast(VultronReport | None, self.object_)
```

For event classes with validated invariants (e.g. `ValidateReportReceivedEvent`),
properties can narrow return types to non-optional — consistent with the
existing `@model_validator`:

```python
class ValidateReportReceivedEvent(VultronEvent):
    @property
    def offer_id(self) -> str:          # non-optional — model_validator guarantees it
[...]

Issue #534, opened 2026-05-15
bug: co-located actors in same process share module-level singletons (emitter, dispatcher, DataLayer)


ahouseholder:
## Summary

When multiple Vultron actors are co-located in the same Python process (e.g.,
a report-receiver actor and its spawned Case Actor, or a participant actor and
its replica-manager Case Actor), they share module-level singletons that were
designed for single-actor-per-process deployment:

- `_default_emitter` in `vultron/adapters/driving/fastapi/outbox_handler.py`
- `_DISPATCHER` in `vultron/adapters/driving/fastapi/inbox_handler.py`
- `_shared_instance` / `_actor_instances` in `vultron/adapters/driven/datalayer_sqlite.py`

The last app lifespan to call `configure_default_emitter()` or `init_dispatcher()`
overwrites the others, and all co-located actors share the same underlying SQLite
engine.

## Expected Behaviour

Each co-located actor instance should have its own isolated emitter, dispatcher,
and DataLayer — objects created by Actor A must only become visible to Actor B
after Actor A's outbox delivers a notification to Actor B's inbox.

## Observed Behaviour

In the current architecture, all actors in the same process share a single
module-level `_default_emitter`, `_DISPATCHER`, and SQLite engine.  A Case
Actor spawned during case creation on the same server as the report-receiver
would use the same DataLayer as the report-receiver, bypassing the
outbox → inbox delivery path entirely.

## Components Involved

- `vultron/adapters/driving/fastapi/outbox_handler.py` — `_default_emitter` global
- `vultron/adapters/driving/fastapi/inbox_handler.py` — `_DISPATCHER` global
- `vultron/adapters/driven/datalayer_sqlite.py` — `_shared_instance` / `_actor_instances`
- `vultron/adapters/driving/fastapi/app.py` — no `create_app()` factory; only a module-level singleton

## Root Cause Hypothesis

The outbox handler, inbox dispatcher, and DataLayer were designed for
single-actor-per-process deployment.  Co-located actors in the same process
expose the assumption that one process == one actor.

## Suggested Fix Direction

Move per-app state (emitter, dispatcher, DataLayer) onto `app.state` rather
than module-level globals.  Add a `create_app()` factory so tests and
co-located deployments can spin up isolated app instances.  This also enables
the `outbox_handler` background task to receive the correct per-app emitter
via `request.app.state.emitter` rather than the global `get_default_emitter()`.

## Impact

All co-located deployments (Case Actor + report-receiver on same server,
[...]

Issue #533, opened 2026-05-15
Idea: Interactive ReactFlow Demo UI with CYOA Scenarios for Stakeholder Presentations


ahouseholder:
## Summary

Replace log-scrolling demos with a polished, interactive ReactFlow web UI
that lets non-technical stakeholders experience Vultron's CVD protocol
behavior as a choose-your-own-adventure. The UI visualizes protocol state
in real time across all running actor containers and lets a presenter (or
viewer) make real protocol decisions at key branch points.

## Motivation

Watching Python logs scroll by convinces developers that the system works.
It does nothing for government funders deciding whether to invest, or
industry practitioners deciding whether to adopt Vultron. We need a
presentation artifact that:

- Shows the protocol as a living, interactive story
- Lets stakeholders take the role of a participant and feel the protocol
  working
- Works as a turnkey demo: `git clone; docker compose up; open browser`
- Scales from the two-actor simple CVD case to multi-vendor supply chain
  scenarios without changing the UI architecture

## Design (resolved via grill-me session 2026-05-15)

Full design notes: `notes/demo-interactive-ui.md`

### Three-Scenario Set

All three scenarios run in the same UI, differing only in actor count and
swim lanes:

| Scenario | Actors | Containers | Narrative |
|----------|--------|------------|-----------|
| A — Two-Actor | Finder, Vendor, Case Actor | 2 | Simple CVD: researcher reports a bug to a vendor directly |
| B — Three-Actor | Finder, Vendor, Coordinator, Case Actor | 4 | MPCVD: reporter routes through a neutral coordinator for a widely-used library |
| C — Multi-Vendor | Finder, Coordinator, Vendor1, Vendor2, Case Actor | 5 | Supply chain: shared component used by multiple vendors |

### CYOA Branch Points (priority order)

1. Report validation (validate / invalidate-hold / invalidate-close)
2. Embargo negotiation (accept / reject / counter-propose)
3. Publication (on schedule / early / request extension)
4. Embargo collapse response (re-negotiate / accelerate)
5. Case ownership transfer (retain / offer to coordinator)

The Python demo scripts remain as automated happy-path puppeteers and
can be observed through the same UI.

### UI Architecture

- **Frontend**: ReactFlow (JS/React) in a Docker Compose service
- **Deployment**: All-in-one `docker compose up` — turnkey for demos
- **Layout**: Top = actor network graph with visual state-machine
  scorecards; Bottom = vertical swimlane message timeline, newest at top
[...]

Issue #526, opened 2026-05-15
Idea: Scheduled GitHub Action for automated monthly/quarterly project reports


ahouseholder:
## Summary

Automate the `project-report` skill on a schedule so monthly, quarterly,
and year-end reports are generated without manual intervention, then
published as GitHub Discussions > Announcements so stakeholders receive
email notification automatically.

### Trigger

- **Scheduled**: runs on the 1st of each month (generates the previous
  month's report via the `project-report` skill)
- **Manual `workflow_dispatch`**: accepts a `period` input (e.g.
  `Q1 2026`, `April 2026`) and an optional `suppress_discussion`
  boolean (default: false)

### Auto-detection of period boundaries

When running on a schedule, the action auto-detects whether the previous
month also ended a quarter or year and generates additional reports:

| Run date | Reports generated |
|---|---|
| April 1 | March 2026 + Q1 2026 |
| July 1 | June 2026 + Q2 2026 |
| October 1 | September 2026 + Q3 2026 |
| January 1 | December 2025 + Q4 2025 + Full-year 2025 |
| Any other 1st | Monthly only |

### Report generation

Uses the **Copilot coding agent** to execute the `project-report` skill.
The agent gathers data (git log, GitHub PRs, plan/history files, existing
monthly reports as seed), drafts a capability-focused narrative, and writes
the file without the interactive draft-review loop.

### Output file

Written to:
```
plan/history/<YYMM>/report/project-report-<YYYYMMDD>-<YYYYMMDD>.md
```

Where YYMM is the month containing the end date of the period.
Files in `plan/history/YYMM/report/` are exempt from history
immutability rules and may be overwritten on regeneration.

Committed directly to `main` with a `chore:` commit message
(no PR needed — generated artifact).

### Notification

After the file is committed, post the **full report body** as a new
discussion in **GitHub Discussions > Announcements** using the
`createDiscussion` GraphQL mutation. Repo watchers receive email
automatically.

The `suppress_discussion` workflow input (manual dispatch only)
skips this step.

### Authentication

Uses `GITHUB_TOKEN` with explicit permissions:
```yaml
permissions:
  contents: write       # commit the report file
  discussions: write    # create the Announcements discussion
```

## Out of scope for now

- Slack / webhook notifications (watcher email via Discussions is sufficient)
- PRs for review before merge (generated artifacts don't need review)
- Custom label on the Discussion (not supported by GitHub Discussions API)

## Prerequisites

[...]

PR #536, opened 2026-05-15
fix: isolate per-actor DataLayers in demo integration tests (#530)

### LINKED ISSUE #530: bug(test): demo integration tests share a single DataLayer across actors — delivery path untested
## Summary

Demo integration tests route both the Finder and Vendor actors through a
**single FastAPI app instance** backed by a **single shared `SqliteDataLayer`**.
This means actor DataLayers are not isolated — each actor can see objects
created by the other directly, without any inbox delivery.

## Expected Behaviour

In a real multi-container deployment, each actor has its own separate DataLayer.
An object created by Vendor is only visible to Finder after Vendor's outbox
delivers a notification to Finder's inbox and Finder's inbox handler processes
it. Tests that share a DataLayer bypass this entire path and cannot verify it.

## Observed Behaviour

`test/demo/conftest.py` creates one `TestClient(api_app)` and patches
`DataLayerClient.call` on **the class itself** (not per-instance) so all
`DataLayerClient` instances — whether labelled `finder_client` or
`vendor_client` — route to the same ASGI app with the same in-memory
`sqlite:///:memory:` DataLayer.

As a result:
- `finder_client.get("/datalayer/VulnerabilityCases/")` returns cases
  created by Vendor without any delivery step occurring.
- Cross-actor polling helpers (`wait_for_finder_case`, etc.) succeed
  immediately, not because delivery worked, but because both actors share
  storage.
- The demo integration tests assert that the full workflow "succeeds"
  without ever exercising the outbox → inbox delivery path.

## Components Involved

- `test/demo/conftest.py` — single `client` fixture, single app instance,
  class-level `DataLayerClient.call` patch
- `test/demo/test_two_actor_demo.py` — `_make_client(base)` creates
  clients that appear distinct but share a DataLayer
- `vultron/adapters/driven/datalayer_sqlite.py` — `SqliteDataLayer` is a
  singleton per connection URL; all tests use `sqlite:///:memory:`

## Suggested Fix Direction

Use **two separate FastAPI app instances** (or two separate lifespan
contexts) in demo integration tests, one per actor. Each instance gets its
own `sqlite:///:memory:` DataLayer. Outbox delivery from Actor A to Actor B
should POST to Actor B's `TestClient` via its ASGI transport — i.e., the
`ASGIEmitter` for Actor A's app should be configured to route Actor B's
inbox URL through Actor B's `TestClient`.

Alternatively: introduce a per-actor DataLayer factory keyed by actor ID
that can be injected into tests.

## Impact

All demo integration tests may be passing for the wrong reasons. The
delivery mechanism (`ASGIEmitter` → inbox handler) is effectively untested
in the current suite.

ahouseholder:
Fixes #530

## Summary

Demo integration tests shared a single DataLayer across actors, meaning
the outbox→inbox delivery path was never exercised. This PR introduces
per-actor DataLayer isolation in test infrastructure and fixes global
singleton contamination that caused the combined test suite to deadlock.

### Production code
- **`app.py`**: Refactored `lifespan` into `_make_lifespan(configure_globals)`
  factory. `app_v2` uses `configure_globals=True`; `create_app()` uses `False`
  so isolated test apps don't overwrite module-level singletons. Clears
  `application.state.emitter` on exit to prevent stale emitter leaks.
- **`actors.py`**: Switched from `_shared_dl()` to `Depends(get_shared_dl)`,
  added `Request` parameter to pass per-app emitter to inbox/outbox handlers.
- **`deps.py`**: Uses `Depends(get_shared_dl)` and `dl.clone_for_actor()`.
- **`datalayer.py/datalayer_sqlite.py`**: Added `get_shared_dl()` facade and
  `clone_for_actor()` on DataLayer Protocol.
- **`inbox_handler.py`**: Accepts optional `emitter` param, passes to
  `outbox_handler`.
- **Router fixes**: `datalayer.py`, `health.py`, `info.py` — all use
  `get_shared_dl` instead of removed `_shared_dl()` wrappers.

### Test code
- **`test/demo/conftest.py`**: Added `_TestASGIRouter`,
  `IsolatedActorApp`, `create_isolated_actor_app()` for multi-app test
  infrastructure.
- **`test/demo/test_two_actor_demo.py`**: Added `TestDeliveryIsolation`
  class (3 tests) verifying DataLayer isolation.

## Root cause of the deadlock

`post_actor_inbox` reads `request.app.state.emitter`. When unit tests
ran before demo tests, a stale `ASGIEmitter` (with `DeliveryQueueAdapter`
fallback) persisted on `app_v2.state.emitter` from a prior lifespan cycle.
Demo tests patched the *global* emitter's fallback with
`_NullDeliveryAdapter`, but the stale per-app emitter was used instead,
causing real HTTP delivery attempts that blocked forever.

Also filed: #534 (production co-location risk).

## Incidental changes

These commits landed on this branch but are not part of the bugfix:

- **`.github/lsp.json`**: Added pyright LSP configuration for Python —
  enables code intelligence (symbol lookup, type info) for the Copilot CLI
  agent.
- **`.github/copilot-instructions.md`**: Added four chronicle-derived
  friction rules: `append-history` timestamp rule, default base branch is
  always `main`, demo/integration test requirement when touching
  `vultron/demo/` or `test/demo/`, and DataLayer isolation rule for tests.
- **`plan/PRIORITIES.md`**: Updated after 2026-05-15 review — mark #467 and
  #527 closed in P470, add #534 to P470, wire Epic #539 into P485 with RFC
  issues #535/#537/#538, add new P510 (Interactive Demo UI) with #533, add
  #526 to P99999.


ahouseholder (commented):
All four review comments have been addressed in commit 8272aeab:

1. **mount_prefix mismatch**: `mount_prefix` is now a parameter of `_make_lifespan()`. `app_v2` uses the default `"/api/v2"`; `create_app()` passes `mount_prefix=""` to avoid double-stripping.
2. **init_dispatcher() skipped**: `init_dispatcher()` now runs unconditionally (outside the `configure_globals` guard) so isolated apps can handle inbox delivery.
3. **Stale docstring**: `create_app()` docstring updated to accurately reflect that global singletons are not modified, and that the dispatcher is always initialised.
4. **Unused variables**: Prefixed with `_` (`_vendor_isolated`, `_vendor_tc`, `_finder_tc`).

PR #617, opened 2026-05-21
fix: use /{key:path} catch-all for DataLayer and actors endpoints

### LINKED ISSUE #610: bug: DataLayer endpoint returns HTTP 404 for URL-format CaseParticipant IDs
### Summary

The `GET /api/v2/datalayer/{id}` endpoint returns `404 Not Found` when
the object ID uses a full HTTP URL (e.g.,
`http://vendor:7999/api/v2/actors/case-actor-{uuid}/participant`), even
though the record was successfully stored by `DataLayer.save(CaseParticipant)`.

### Prior acknowledgement — not a fix

`_dl_key()` in `vultron/demo/helpers/seeding.py` already documents this
failure mode in its docstring:

> Note: HTTP URL-based participant IDs (which contain literal slashes)
> cannot be fetched via the single-segment `/{key}` DataLayer route even
> after URL-encoding, because Starlette decodes `%2F` back to `/` before
> path matching. Such IDs must be handled via exception catching at the
> call site.

This comment documents the problem but does **not** fix it. Leaving a
known-broken code path with a comment and no actual resolution is not
acceptable — the real fix was never implemented.

### Observed in

CI run for PR #590 (merged):
https://github.com/CERTCC/Vultron/actions/runs/26242597251

### Reproduction

Two-actor demo integration test — M6 post-failure diagnostics (Phase 5).

### Log evidence

```
# Record created successfully at startup:
vendor-1 | INFO: DataLayer stored CaseParticipant
    'http://vendor:7999/api/v2/actors/case-actor-3af16520-.../participant'

# ... later, demo-runner diagnostic lookup returns 404:
demo-runner-1 | ERROR: HTTP 404 from
    http://vendor:7999/api/v2/datalayer/
    http%3A%2F%2Fvendor%3A7999%2Fapi%2Fv2%2Factors%2F
    case-actor-3af16520-6b5e-450a-bca0-d47afdad572f%2Fparticipant
demo-runner-1 | ERROR: Error response: {"detail":"Not Found"}
```

Same 404 on the finder replica.

### Expected behaviour

`GET /datalayer/{url-encoded-id}` must return the stored `CaseParticipant`
record regardless of whether the ID is a URN or a full HTTP URL.

### Required fix

Either:

1. **Fix the DataLayer endpoint** to correctly handle percent-encoded HTTP
   URL keys — e.g., use a catch-all path parameter (`/{key:path}`) so
   Starlette does not split on decoded slashes; or
2. **Fix every call site** that fetches URL-format IDs to use an alternative
   lookup mechanism (e.g., query by type + filter, or a dedicated
   `/datalayer/by-id` POST endpoint that accepts the raw ID in the request
   body).

Option 1 is preferred as it fixes the root cause once rather than patching
every consumer. The docstring in `_dl_key` should be updated once the fix
lands to remove the "known limitation" language.

ahouseholder:
Fixes #610

## Problem

`GET /api/v2/datalayer/{url-encoded-id}` returned 404 for CaseParticipant IDs
stored with HTTP URL keys (e.g. `http://vendor:7999/.../participant`). Starlette
decodes `%2F` → `/` before path matching, so the single-segment `/{key}` route
never matched. The same bug existed in `GET /actors/{actor_id}`.

## Fix

- **`datalayer.py`**: Replace `/{key}` (first route) with `/{key:path}` registered
  last — specific routes (/Offers/, /Reports/, /Actors/, etc.) continue to
  match before the catch-all.
- **`actors.py`**: Same — `/{actor_id:path}` appended as last route.
- **`seeding.py`**: Remove "known limitation" from `_dl_key()` docstring.

## Tests

- Added unit regression tests for URL-format key lookup in both routers.
- Updated `test_url_based_participant_id_handled_gracefully` integration test
  to assert the now-correct behaviour (fetch succeeds) rather than the old
  workaround (fetch raises).

### DIFF
--- a/plan/BUILD_LEARNINGS.md
+++ b/plan/BUILD_LEARNINGS.md
@@ -6,3 +6,12 @@ insights, issues, and learnings during the implementation process.
 
 Append new items below any existing ones, marking them with the date and a
 header.
+
+### 2026-05-21 — #610: Starlette path parameter type for URL-keyed endpoints
+
+When a FastAPI/Starlette endpoint needs to accept keys that may contain
+forward slashes (e.g., HTTP URL IDs), use `{key:path}` instead of `{key}`.
+Register the catch-all route **last** so specific literal routes (e.g.,
+`/Offers/`, `/Actors/`) are matched first. The `%2F`-decoding-before-routing
+behaviour in Starlette means there is no client-side workaround; only the
+server-side `path` converter fixes the root cause.
--- a/plan/history/2605/README.md
+++ b/plan/history/2605/README.md
@@ -4,6 +4,7 @@
 
 | Date | Time (UTC) | Type | Source | Title |
 |------|------------|------|--------|-------|
+| 2026-05-21 | 19:19 | implementation | 610 | Fix DataLayer and actors endpoints for HTTP URL keys (#610) |
 | 2026-05-21 | 16:33 | learning | CONCERN-593 | Case Actor outbound routing model |
 | 2026-05-21 | 14:58 | implementation | ISSUE-570 | Demo CI: GitHub Actions integration workflow + demo runner hardening |
 | 2026-05-20 | 13:54 | idea | IDEA-582 | Optimize linkchecker.yml: separate build-site and link-check triggers |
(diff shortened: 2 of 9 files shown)


ahouseholder (commented):
Both review comments addressed in commit 422bfb52:

1. **Sub-routes updated to `{actor_id:path}`** — All five actor sub-routes (`/profile`, `/inbox` GET, `/inbox/` POST, `/outbox/` POST, `/cases/{case_id}/action-rules`) now use `{actor_id:path}` so HTTP URL actor IDs with slashes work correctly throughout.

2. **Constants renamed** — `_HTTP_ACTOR_ID`, `_HTTP_CASE_ID`, `_HTTP_PARTICIPANT_ID` renamed to `_URN_ACTOR_ID`, `_URN_CASE_ID`, `_URN_PARTICIPANT_ID` throughout `test_actors.py`.

PR #660, opened 2026-06-01
fix(#659): use append-order for CaseParticipant.participant_status (+ NullPool/WAL hardening)

### LINKED ISSUE #659: Demo Integration CI: intermittent timeout in _phase_fix_lifecycle (wait_for_participant_vfd_state)
## Summary

The Demo Integration CI job ("Two-Actor Demo Integration") intermittently
fails during `_phase_fix_lifecycle` with:

```
AssertionError: Timed out waiting for actor 'http://vendor:7999/api/v2/actors/...'
vfd_state to be in {VFd, VFD}; current=vfd (vendor_unaware/fix_not_ready/fix_not_deployed)
```

The vendor actor has not yet progressed from the initial vfd state by the
time the timeout fires, even though the EM FSM transitions (NONE → PROPOSED
→ ACTIVE) complete normally. This suggests VFD state machine advancement
races with the integration timeout window.

## Evidence

- CI run 26774923893 on branch
  `ingest/concern-502-datalayer-scope-boundaries` (docs-only change, no
  Python code modified) — failure not related to PR content.
- `vultron/demo/scenario/two_actor_demo.py` line 577 — `wait_for_participant_vfd_state`.
- `vultron/demo/utils.py` line 126 — `demo_check` context manager.

## Impact if Ignored

Flaky CI blocks docs-only and unrelated PRs, erodes confidence in the test
suite, and makes genuine failures harder to spot.

## Suggested Action

1. Investigate whether VFD state advancement depends on an outbound activity
   delivery that races with the timeout.
2. Consider increasing the timeout for `wait_for_participant_vfd_state` or
   adding a retry/backoff.
3. Add a debug log before the assertion capturing the actor's current VFD
   and EM state to improve future failure diagnostics.

ahouseholder:
Fixes #659

## Root cause

`CaseParticipant.participant_status` selected the "current" status via
`max(participant_statuses, key=(updated or published, index))`. The
`init_participant_status_if_empty` validator constructs the initial vfd
with `published=now_utc()` *at local construction time*. When the BT
later appended a sender-authored `ParticipantStatus` (e.g. vendor's VFd
from an M3 `Add` activity), that wire timestamp could be *earlier* than
the finder's local default on the initial entry. The tiebreaker then
selected the stale index-0 initial vfd despite a newer append at index 1.

Result: M4 (`wait_for_participant_vfd_state`) timed out polling for
`VFd` even though the adapter correctly persisted and returned both
statuses.

## Fix

Return `self.participant_statuses[-1]` from the property. Append order
is this replica's authoritative chronology, robust to sender clock skew,
and aligns wire with core (`VultronParticipant` already uses `[-1]`).

## Also in this PR

- **NullPool + WAL** for file-backed SQLite in
  `vultron/adapters/driven/datalayer_sqlite.py` (necessary hardening
  against read-after-write staleness on shared file-backed engines, even
  though it wasn't the root cause of #659).
- **Diagnostic DEBUG logs** in `datalayer_sqlite.py`
  (`_participant_status_summary` on every `read`/`save` of a
  CaseParticipant) and `wait_for_participant_vfd_state` (per-poll
  observed state). Kept in tree — cheap, useful for future demo
  flakiness.
- M4 polling default reverted to 10s.

## Validation

- New unit tests in `TestParticipantStatusProperty` (3 tests; first
  fails on old property, all pass on new).
- Full pytest suite: 2725 passed, 12 skipped, 3 xfailed.
- **Demo Integration CI: 4 consecutive runs passed** on `b80c0587`
  (compared with 3 consecutive failures on the prior diagnostic commit).

## Related

- #663 (`BroadcastStatusToPeersNode` self-loop) — separate concern,
  filed during this investigation, out of scope here.

### DIFF
--- a/plan/history/2606/README.md
+++ b/plan/history/2606/README.md
@@ -4,6 +4,7 @@
 
 | Date | Time (UTC) | Type | Source | Title |
 |------|------------|------|--------|-------|
+| 2026-06-01 | 19:11 | implementation | 659-vfd-timeout-increase | fix: increase wait_for_participant_vfd_state timeout to 30 s (#659) |
 | 2026-06-01 | 18:47 | learning | CONCERN-502 | Actor-scoped vs shared DataLayer scope boundaries |
 | 2026-06-01 | 18:33 | implementation | ISSUE-437 | Enforce spec vs. ADR delineation guidelines (MS-11) |
 | 2026-06-01 | 18:04 | implementation | ISSUE-584 | Rename linkchecker.yml to docs-build-check.yml with narrowed triggers |
--- a/plan/history/2606/README.md
+++ b/plan/history/2606/README.md
@@ -4,6 +4,7 @@
 
 | Date | Time (UTC) | Type | Source | Title |
 |------|------------|------|--------|-------|
+| 2026-06-02 | 16:38 | implementation | BUG-659-FIX2 | Bug #659 follow-up — participant_status append-order fix |
 | 2026-06-02 | 15:22 | implementation | BUG-659 | BUG-659 — Fix SQLite read-after-write staleness causing M4 timeout |
 | 2026-06-01 | 19:11 | implementation | 659-vfd-timeout-increase | fix: increase wait_for_participant_vfd_state timeout to 30 s (#659) |
 | 2026-06-01 | 18:47 | learning | CONCERN-502 | Actor-scoped vs shared DataLayer scope boundaries |
--- a/plan/history/2606/README.md
+++ b/plan/history/2606/README.md
@@ -4,6 +4,7 @@
 
 | Date | Time (UTC) | Type | Source | Title |
 |------|------------|------|--------|-------|
+| 2026-06-02 | 15:22 | implementation | BUG-659 | BUG-659 — Fix SQLite read-after-write staleness causing M4 timeout |
 | 2026-06-01 | 19:11 | implementation | 659-vfd-timeout-increase | fix: increase wait_for_participant_vfd_state timeout to 30 s (#659) |
 | 2026-06-01 | 18:47 | learning | CONCERN-502 | Actor-scoped vs shared DataLayer scope boundaries |
 | 2026-06-01 | 18:33 | implementation | ISSUE-437 | Enforce spec vs. ADR delineation guidelines (MS-11) |
--- a/plan/history/2606/implementation/659-vfd-timeout-increase.md
+++ b/plan/history/2606/implementation/659-vfd-timeout-increase.md
@@ -1,0 +1,8 @@
+---
+source: 659-vfd-timeout-increase
+timestamp: '2026-06-01T19:11:41.913314+00:00'
+title: 'fix: increase wait_for_participant_vfd_state timeout to 30 s (#659)'
+type: implementation
+---
+
+fix: increase wait_for_participant_vfd_state timeout to 30 s to accommodate cross-container async delivery pipeline latency under CI load. Also add DEBUG diagnostic log before AssertionError.
--- a/plan/history/2606/implementation/BUG-659-FIX2.md
+++ b/plan/history/2606/implementation/BUG-659-FIX2.md
@@ -1,0 +1,58 @@
+---
+source: BUG-659-FIX2
+timestamp: '2026-06-02T16:38:58.835298+00:00'
+title: 'Bug #659 follow-up — participant_status append-order fix'
+type: implementation
+---
+
+## Context
+
+Follow-up to the initial BUG-659.md entry. The NullPool + WAL adapter
+hardening (commit aafaa868) was necessary but did not eliminate the
+M4 timeout. Diagnostic logging added in commit 7e3fc52b localised the
+real root cause to the wire-layer `CaseParticipant.participant_status`
+property.
+
+## Symptoms
+
+`vultron-demo two-actor` flaked in CI: M4
+(`wait_for_participant_vfd_state`) timed out waiting for the vendor's
+vfd_state to reach `VFd` on the finder replica, even though
+adapter-side diagnostic logs confirmed the `VFd` ParticipantStatus
+was correctly persisted and read back from SQLite on every poll.
+
+## Root cause
+
+`CaseParticipant.participant_status` selected the "current" status by
+`max(participant_statuses, key=(updated or published, index))`. The
+`init_participant_status_if_empty` model validator constructs the
+initial vfd locally with `published=now_utc()` *at construction time*.
+When the M3 `Add(VFd)` activity arrived and the BT appended a new
+`ParticipantStatus`, that status carried sender-authored timestamps
+that could be *earlier* than the finder's local default on the initial
+entry (any clock skew or processing gap is enough). The tiebreaker
+then selected index-0 initial vfd despite a freshly appended VFd at
+index 1.
+
+## Fix
+
+Return `self.participant_statuses[-1]` from the property. Append order
+IS this replica's authoritative chronology; it is robust to sender
+clock skew and matches what `VultronParticipant` already does in the
+core layer.
+
+## Validation
+
+- Local repro via `jsonable_encoder` round-trip demonstrated the bug
+  deterministically and confirmed the fix.
+- New unit tests in `test/wire/as2/vocab/test_case_participant.py`
+  (`TestParticipantStatusProperty`): 3 tests, first one fails on old
+  property, all pass with the fix.
+- Full pytest suite: 2725 passed, 12 skipped, 3 xfailed.
+- Kept diagnostic DEBUG logs in `datalayer_sqlite.py` and
+  `wait_for_participant_vfd_state` — cheap insurance for next time.
+
+## Related
+
+- #663 (`BroadcastStatusToPeersNode` self-loop) still open and out of
+  scope here.
--- a/plan/history/2606/implementation/BUG-659.md
+++ b/plan/history/2606/implementation/BUG-659.md
@@ -1,0 +1,57 @@
+---
+source: BUG-659
+timestamp: '2026-06-02T15:22:31.481660+00:00'
+title: BUG-659 — Fix SQLite read-after-write staleness causing M4 timeout
+type: implementation
+---
+
+## Summary
+
+Fixed root cause of intermittent CI timeout in
+`_phase_fix_lifecycle` / `wait_for_participant_vfd_state` (#659). The
+previous fix (PR #660, commit 364a8de3) bumped the polling timeout from
+10 s to 30 s; CI run 26776168609 confirmed the 30 s timeout still
+expired, ruling out a pure timing issue.
+
+## Root cause
+
+`SqliteDataLayer._make_engine` used SQLAlchemy's default pool for
+file-backed SQLite (`SingletonThreadPool`), which reuses a single
+DB-API connection per thread. In the FastAPI driving adapter, a
+`BackgroundTask` (BT save of a `CaseParticipant`) and a concurrent GET
+handler can end up sharing the same connection. Combined with SQLite's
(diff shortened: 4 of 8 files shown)


ahouseholder:
## 🔄 Bugfix Session Interrupted — Handoff Notes

**Branch**: `bug/659-vfd-timeout-increase`
**Issue**: #659 / #660 — M4 `wait_for_participant_vfd_state` timeout (VFd propagation to finder replica)
**Interrupted at phase**: Phase 3 — Root cause investigation (diagnosis complete, no fix implemented yet)
**In progress at interrupt**: Tracing why `CaseParticipant de553587` GET returns stale `vfd` state for 30+ seconds after a confirmed `save()` at log line 1145, after exhausting all static-analysis paths.

---

### What was tried

| Attempt | Outcome |
|---|---|
| Identified failing CI job from PR #660 logs | Demo Integration / Two-Actor Demo Integration, run 26776168609, exit code 1 |
| Audited all June 1 merged PRs for code changes | PR #645 (`_is_case_owner` fail-closed) is the only code change; confirmed it does NOT affect VFd flow |
| Downloaded and analyzed CI artifact `demo-container-logs` | M3 passes (vendor VFd at 19:14:54.808); M4 times out after exactly 30s |
| Traced Add(ParticipantStatus) delivery from vendor → finder | Vendor sends to finder (HTTP 202 confirmed at vendor log line 1857); finder stores status at line 1109, BT saves participant at line 1145 — all before any M4 poll |
| Checked for intermediate overwrites of `de553587` between save (line 1145) and first poll (line 1190) | None found — only CaseLogEntry, Announce, Person saves between those lines |
| Round-trip serialization test (local) | `CaseParticipant` → record → `CaseParticipant` with valid `ParticipantStatus` round-trips correctly; VFd preserved |
| Local integration tests (`uv run pytest test/demo/ -m integration`) | All 210 pass — local tests use in-process TestClient, not real HTTP |
| Ran Docker compose test locally (2× runs) | Both passed! The failure appears to be flaky under CI load, not deterministic |
| Investigated `BroadcastStatusToPeersNode` "2 peer(s)" anomaly on finder | Finder broadcasts to `vendor_actor` AND `finder_actor` (itself) — message loop — but idempotency should prevent corruption |
| Checked `ParticipantStatus` validation requirements | Wire format from vendor IS correct (context present, booleans as bool, not strings) — earlier hypothesis eliminated |
| Checked `actor_participant_index` mapping | `vendor_actor → de553587` confirmed at finder log line 177; M4 IS polling the right participant |
| Verified `_scoped()` / actor-scoped DL logic | BT uses shared DL (no actor filter); GET endpoint also uses shared DL — both see same rows |
| Checked `save()` does not update `actor_id` on existing row | Confirmed — existing rows keep their original `actor_id`; PK ensures single row per `id_` |
| Verified `participant_status` property tiebreaker | Initial `vfd` status has `None` timestamps → `_EPOCH`; VFd status has `published=2026-06-01T19:14:54Z` → VFd always wins |

---

### Root cause analysis

**Confirmed facts** (evidence in code, test output, or logs):

- M4 polls `de553587` on the finder for 30s and ALWAYS sees `vfd` — the GET succeeds (200 OK) but returns stale data
- BT saves `de553587` at finder log line 1145 (confirmed by "DataLayer saved CaseParticipant" log), which is BEFORE the first M4 poll at line 1190
- No intermediate write to `de553587` between line 1145 and the first poll
- Local Docker runs pass reliably; only CI shows the failure
- The `BroadcastStatusToPeersNode` at the finder incorrectly broadcasts to 2 peers (vendor_actor + finder_actor) instead of 1 — creates a message loop, but idempotency prevents data corruption
- The second save of `de553587` at line 1884 is for a DIFFERENT status (`2444d657`, VFD/public) triggered by the M5/M6 flow, arriving AFTER the 30s M4 timeout

**Active hypotheses** (plausible but not yet confirmed):

- **H1: Race condition under CI load** — The BT BackgroundTask and the M4 GET run concurrently in uvicorn's event loop. Under CI resource pressure, the BackgroundTask may not have committed by the time M4 polls, even though the log ordering suggests otherwise. SQLAlchemy `SingletonThreadPool` reuses one connection per thread; if the GET and the BackgroundTask share a thread, a dirty-read or uncommitted-read scenario could explain 30s of stale data. Supporting evidence: passes consistently locally (less load), fails in CI; the 30s window matches exactly the M4 timeout (not a brief gap).
- **H2: `BroadcastStatusToPeersNode` self-loop overwrites the save** — The finder broadcasts to itself (`finder_actor`), queuing a second inbox processing of the same Add activity. If this second invocation reads `de553587` BEFORE the first save commits (due to async concurrency), it appends the same status and saves, but might overwrite with an in-memory object that still has the old state. Supporting evidence: "2 peer(s)" log at line 1150; second processing would produce the save at line 1884 for a different status though — so timing may not match.
- **H3: SQLAlchemy session isolation under asyncio** — In FastAPI with asyncio (single-thread event loop), `SingletonThreadPool` reuses ONE SQLite connection. If a read transaction was started before the commit, SQLite's DEFERRED isolation might return a snapshot that predates the write. The `with Session(engine) as session:` idiom in `read()` should begin a fresh transaction, but the connection's isolation level under `pysqlite` is non-trivial.

**Ruled out**:

- `ParticipantStatus` validation failure (context missing, wrong bool types) — the actual wire format has correct types; round-trip works
- Wrong participant ID in M4 poll — `actor_participant_index[vendor_actor] = de553587` confirmed
- PR #645 (`_is_case_owner`) causing the regression — only used in embargo triggers, not VFd flow
- Intermediate overwrite between save and first poll — log analysis shows none
- `init_participant_status_if_empty` reinitializing to `vfd` — would only fire if `participant_statuses` is empty; stored data has 2 statuses

---

### Current scope

The problem appears to be localized to:

- `vultron/core/behaviors/status/nodes.py` (~lines 151–282) — `AppendParticipantStatusNode`: confirms save succeeds but state not visible to concurrent reads under load
- `vultron/adapters/driven/datalayer_sqlite.py` (~lines 430–480, 557–590) — `read()` and `save()`: SQLite session/connection lifecycle under asyncio concurrency
- `vultron/core/behaviors/status/nodes.py` (~lines 283–390) — `BroadcastStatusToPeersNode`: "2 peer(s)" anomaly creating a self-loop

Components / layers involved:

- **SQLite persistence layer**: SQLAlchemy `SingletonThreadPool` + asyncio may produce read-after-write visibility issues under CI load
- **Status BT**: AppendParticipantStatus → save → BroadcastStatusToPeers → self-loop may create race
- **M4 polling loop**: `wait_for_participant_vfd_state` in `vultron/demo/helpers/polling.py` polls every 0.25s for up to 30s

---

### Open questions and blockers

1. **Does `SingletonThreadPool` cause stale reads in asyncio?** If the BackgroundTask and GET handler run in the same asyncio thread (single-event-loop uvicorn), and `SingletonThreadPool` gives them the same SQLite connection, can a read transaction started before the commit see stale data?
2. **Does `BroadcastStatusToPeersNode` on the finder actually deliver to `finder_actor`'s inbox?** Does the finder's outbox monitor pick up activities queued for `vendor_case_actor` (the broadcast actor), or only for finder's own actors? If delivery never happens, the self-loop theory is wrong.
3. **Is the "2 peers" broadcast itself the bug?** If the finder sends an Add to its own inbox, this triggers another BackgroundTask. Could this second BackgroundTask run concurrently with the M4 polls and somehow clobber the saved state?
4. **What does the CI runner's resource profile look like?** Is the CI runner known to be slow enough that a BackgroundTask could take >30s to execute?

---

### Suggested next steps

1. **Instrument `read()` with a log line** — Add `logger.debug("read(%s) -> data=%s", object_id, row.data)` in `datalayer_sqlite.py` before `_from_row` to confirm what data is actually in the DB when M4 polls. This would distinguish "data not committed" from "data committed but wrong value".
2. **Fix `BroadcastStatusToPeersNode` self-loop** — In `vultron/core/behaviors/status/nodes.py` line 344, add `self_actor_id` (the currently-processing actor) to the exclusion filter: `if a_id != self.sender_actor_id and a_id != case_manager_id and a_id != self_actor_id`. The current_actor_id is available from the BT blackboard. This is a correctness bug regardless of whether it's the root cause of M4 failure.
3. **Switch to `NullPool` for file-based SQLite** — In `datalayer_sqlite.py::_make_engine`, add `poolclass=NullPool` for non-memory databases. `NullPool` gives a fresh connection per `with Session(engine)` block, eliminating any connection-reuse visibility issues. This is the most likely fix for H1/H3.
4. **Add WAL mode pragma** — If staying with `SingletonThreadPool`, enable WAL mode: `engine.execute("PRAGMA journal_mode=WAL")` at startup. WAL allows concurrent readers to see committed writes immediately.
5. **Run CI multiple times on the 30s-timeout branch** to confirm flakiness rate before investing more in root cause. If it passes 3/5 times, H1 (load-sensitive race) is confirmed.

---

### Relevant code locations

| File | Lines | Why it matters |
|---|---|---|
| `vultron/adapters/driven/datalayer_sqlite.py` | ~104–125 | `_make_engine`: pool configuration — `NullPool` for `:memory:`, default (`SingletonThreadPool`) for file |
| `vultron/adapters/driven/datalayer_sqlite.py` | ~430–470 | `read()`: opens `Session`, applies `_scoped()`, calls `_from_row` |
| `vultron/adapters/driven/datalayer_sqlite.py` | ~557–590 | `save()`: `session.get()` by PK + `session.commit()` |
| `vultron/core/behaviors/status/nodes.py` | ~151–282 | `AppendParticipantStatusNode`: reads participant, appends status, saves |
| `vultron/core/behaviors/status/nodes.py` | ~283–390 | `BroadcastStatusToPeersNode`: "2 peer(s)" anomaly — broadcasts to self |
| `vultron/demo/helpers/polling.py` | ~272–335 | `wait_for_participant_vfd_state`: M4 poll loop |
| `vultron/demo/helpers/verification.py` | ~43–70 | `_fetch_participant`: GETs case → extracts participant_id → GETs participant |
| `/tmp/demo-logs-660/finder.log` | lines 1145–1200 | BT save then immediate M4 polls all returning stale `vfd` |

---

### Test status at interrupt

```text
Local unit tests:      uv run pytest --tb=short  → all pass (baseline)
Local integration:     uv run pytest test/demo/ -m integration  → 210 passed
Local Docker compose:  2 runs, both passed (less load than CI)
CI run 26776168609:    FAILED — M4 timed out after 30s, vfd_state never reached VFd

No code changes have been made on this branch beyond the 30s timeout increase
from PR #659 (commit 364a8de3).
```

---

### Notes for the next agent

- **This branch (`bug/659-vfd-timeout-increase`) already has the 30s timeout commit merged** (PR #659, commit `364a8de3`). PR #660 was opened to investigate whether the timeout is treating a symptom. The working tree is clean.
- The CI artifact `demo-container-logs` from run 26776168609 is available at `/tmp/demo-logs-660/` on the session machine (ephemeral — re-download with `gh run download 26776168609 -n demo-container-logs -D /tmp/demo-logs-660/`).
- The most tractable fix is probably **`NullPool` for file SQLite** (step 3 above) — it removes the connection-reuse ambiguity entirely and is a low-risk change. Alternatively, adding `WAL mode` is even lower risk.
- The **`BroadcastStatusToPeersNode` self-loop** (finder broadcasting to itself) is a separate correctness bug that should be fixed regardless. The finder should not re-broadcast a status it just received; only the Case Manager should broadcast, and only to OTHER participants.
- Key demo artifact IDs for tracing (from CI run): Case `c9fd96f9`, Vendor participant at finder `de553587`, VFd status `4938dd48`.
- Local tests pass because they use FastAPI `TestClient` (synchronous, in-process) — they never exercise the asyncio + real SQLite + HTTP concurrency that triggers the race in CI.

---

*Handoff generated by the `bugfix-handoff` skill.*

PR #733, opened 2026-06-03
Merge main into feature/demo-ui

sei-ahouseholder:
Catches `feature/demo-ui` up with the 239 commits on `main` since the branch diverged. Only conflict is a trivial `.gitignore` addition on each side (both append non-overlapping lines at the end).


sei-ahouseholder:
closing. will resolve this later.

PR #886, opened 2026-06-10
Consolidate report prioritization sender-side BT

### LINKED ISSUE #753: Consolidate EmitEngageCaseActivity and EmitDeferCaseActivity into SenderSideBT pattern
## Summary

`EmitEngageCaseActivity` and `EmitDeferCaseActivity` in
`vultron/core/behaviors/report/nodes.py` are near-duplicates (~130 combined lines).
Both nodes do the same four things, differing only in the wire factory function called:

1. Read case + resolve addressees
2. Call factory to create an Accept/Reject activity
3. Persist idempotently
4. Append to `outbox.items` + `record_outbox_item`

This is exactly the pattern already implemented in `sender/nodes.py` as `SenderSideBT`:
```
SenderSideBT (Sequence)
├── ResolveCaseManagerNode
├── ConstructActivitiesNode   ← factory injected here
└── QueueToOutboxNode
```

## Acceptance Criteria

- [ ] AC-1: `EmitEngageCaseActivity` replaced by `SenderSideBT` with
  `make_engage_case_activity` factory injected
- [ ] AC-2: `EmitDeferCaseActivity` replaced by `SenderSideBT` with
  `make_defer_case_activity` factory injected
- [ ] AC-3: ~130 lines of duplicated code removed from `report/nodes.py`
- [ ] AC-4: Existing tests pass; new tests confirm factory injection works

## Reference

Notes: `notes/bt-integration.md`
Related: #738 (report/nodes.py split — should be done first)

sei-ahouseholder:
- Closes #753

## Summary

Consolidate report engage/defer emission into a shared sender-side BT sequence so the prioritization tree uses one reusable routing pattern.

## Changes

- vultron/core/behaviors/report/prioritize_tree.py: replaced the duplicate emit-node logic with a shared sender-side sequence that resolves report addressees, emits the activity, and queues it.
- vultron/core/behaviors/report/nodes/emit.py: removed the obsolete engage/defer emit node classes.
- vultron/core/behaviors/report/nodes/__init__.py: dropped the removed exports.
- test/core/behaviors/report/test_prioritize_tree.py: updated tree-structure assertions and added defer-path coverage.
- test/core/behaviors/report/nodes/test_emit.py: removed the obsolete unit test module.

## Verification

- 3139 passed, 12 skipped, 219 deselected, 5633 subtests passed.
- Black, flake8, mypy, and pyright are clean.


sei-ahouseholder (commented):
## Code Review — PR #886

Reviewed against: issue #753, `notes/case-communication-model.md`, `specs/participant-case-replica.yaml` (PCR-08), `specs/behavior-tree-integration.yaml` (BT-06), and `vultron/core/behaviors/sender/`.

---

### 🔴 Critical: PCR-08-001 / PCR-08-002 violated — `case_addressees()` used on sender side

`ResolveCaseAddresseesNode.update()` calls `case_addressees(case, self.actor_id)`, which returns **all participant actor IDs except the sender**. Addressing outbound participant activities to that full list directly violates:

- **PCR-08-001**: "a participant actor MUST address all case-scoped outbound messages … exclusively to the Case Actor."
- **PCR-08-002**: "Sending to `case_addressees()` … is a violation of this rule."

`notes/case-communication-model.md` § "Known Implementation Gaps" explicitly documents this exact pattern (`triggers/case.py:84,132`) as an open bug that this refactor was supposed to **fix**. Instead, the PR perpetuates it in new nodes.

The correct resolver is `_resolve_case_manager_id(case, dl)` (already in `use_cases/_helpers.py`), which is exactly what `ResolveCaseManagerNode` in `sender/nodes.py` uses.

---

### 🔴 Critical: Canonical `SenderSideBT` not reused — parallel (divergent) implementation created

Issue #753 AC-1 and AC-2 say:

> "replaced by `SenderSideBT` with `make_engage_case_activity` factory injected"

The canonical SenderSideBT already exists as `sender_side_bt()` in `vultron/core/behaviors/sender/send_tree.py`. It composes `ResolveCaseManagerNode` + `ConstructActivitiesNode` + `QueueToOutboxNode`. The trigger-side use cases (`SvcEngageCaseUseCase`, `SvcDeferCaseUseCase`) and `engage_defer_trigger_tree.py` already use it correctly via a closure-captured `activity_builder`.

Instead, this PR:
1. Creates two new node classes (`ResolveCaseAddresseesNode`, `EmitCasePriorityActivityNode`) in `prioritize_tree.py`
2. Names the resulting sequence `"SenderSideBT"` — the same name as the canonical pattern but with different internals

A developer seeing `SenderSideBT` in a tree visualization will expect `ResolveCaseManagerNode → ConstructActivitiesNode → QueueToOutboxNode`, not the nodes defined here.

The correct fix is to call `sender_side_bt(case_id, activity_builder)` with a closure, similar to how `SvcEngageCaseUseCase.execute()` builds `_build_activities`. Since `trigger_activity_factory` is available on the `DataLayerAction` blackboard at `update()` time but not at construction time, the closure approach captures it lazily — but `ConstructActivitiesNode` can be adapted to support this, or the factory can be threaded in through the use-case layer.

---

### 🟡 High: BT node classes defined in a tree-composition module

`ResolveCaseAddresseesNode` and `EmitCasePriorityActivityNode` are defined in `prioritize_tree.py`, which is a tree-composition module. Per the project conventions, BT leaf nodes belong in `nodes/` subpackages (AGENTS.md: "every process area under `vultron/core/behaviors/` SHOULD use a `nodes/` subpackage"). If these nodes were correct in design, they should live in `vultron/core/behaviors/report/nodes/emit.py` or a new `report/nodes/prioritize.py`, not inline in the tree factory file.

---

### 🟡 High: String-based method dispatch instead of `Callable` injection

`EmitCasePriorityActivityNode.__init__` accepts `activity_method: str` and in `update()` calls `getattr(factory, self._activity_method, None)`. This is less type-safe than the canonical `ConstructActivitiesNode` approach, which takes `activity_builder: Callable[[str], list[str]]` at construction time. A mistyped string (e.g., `"engage_cases"`) fails only at runtime; a bad callable fails at call sites that can be statically checked.

---

### 🟡 High: Docstring/tree comment mismatch

In `create_prioritize_subtree()`, the docstring still says:

```
├─ SenderSideBT              # resolve case manager and emit Join
```

But the actual `ResolveCaseAddresseesNode` resolves _all addressees_ (via `case_addressees()`), not the case manager. The comment was correct for what the implementation _should_ do but is factually wrong for what it _does_.

---

### 🟡 Medium: `create_prioritize_subtree()` could reuse existing trigger-side BT factories

`engage_case_trigger_bt()` and `defer_case_trigger_bt()` in `engage_defer_trigger_tree.py` already compose `TransitionParticipantRM*` + `sender_side_bt()` for exactly the send-side workflow. The new `create_prioritize_subtree()` rebuilds that same pattern inside a Selector. After the critical issues above are resolved, it would be worth checking whether `create_prioritize_subtree()` can delegate to those factories rather than duplicating the sequence structure.

---

### 🟢 Positive notes

- The test added for the defer path (`test_prioritize_subtree_defers_when_engage_path_fails`) is a real improvement — the defer path had no coverage before.
- `QueueToOutboxNode` from `sender/nodes.py` is correctly reused.
- Deletion of `EmitEngageCaseActivity`/`EmitDeferCaseActivity` from `nodes/emit.py` is the right direction.
- The history entry and PR body are well-formed.

---

### Summary

The PR removes the old duplicate nodes (good) but replaces them with a new parallel implementation that still routes outbound messages incorrectly (`case_addressees` instead of `_resolve_case_manager_id`). This means the known PCR-08 routing bug tracked in `notes/case-communication-model.md` is not fixed — it's moved to new code. The fix needs to use `sender_side_bt()` from `sender/send_tree.py` with a proper `activity_builder` closure, which is how every other trigger-side use case in this codebase handles the same pattern.

sei-ahouseholder:
## 🔄 Bugfix Session Interrupted — Handoff Notes

**Branch**: `task/753-sender-side-bt-pattern`
**PR**: #886 — Consolidate report prioritization sender-side BT
**Interrupted at phase**: Phase 2 — Root-cause analysis complete, no code changes made
**In progress at interrupt**: Deep-tracing whether `case_participants` survive the
DataLayer serialisation round-trip as full objects vs. bare ID strings, to determine
whether fixing Root Cause 1+2 alone is sufficient for AC-1 to pass.

---

### What was tried

| Attempt | Outcome |
|---|---|
| Retrieved CI failure logs for both failing jobs | Got full tracebacks; identified the two test failures and the logged error string |
| Read `SvcValidateReportUseCase` in `triggers/report.py` | **Confirmed** it has no `trigger_activity` param (Root Cause 1) |
| Read `TriggerService.validate_report` in `triggers/service.py` | **Confirmed** it calls `SvcValidateReportUseCase(self._dl, req).execute()` without `trigger_activity` (Root Cause 2) |
| Read `two_app_setup` fixture + `_make_lifespan` + `ASGIEmitter` | Determined isolated test apps use `configure_globals=False` → no config pollution of `_default_emitter` |
| Traced `Create(VulnerabilityCase)` delivery path from step 1 | Concluded reporter SHOULD get owner's `CaseParticipant` via `_store_embedded_participants` during `CreateCaseReceivedUseCase._handle_bootstrap` |
| Read `VulnerabilityCase.to_core()` and `Record.from_obj()` serialisation path | In progress at interrupt — was determining if `case_participants` survive the round-trip as full objects |

---

### Root cause analysis

**Confirmed facts**:

- `SvcValidateReportUseCase.__init__` only accepts `(dl, request)` — no `trigger_activity`.
  `execute()` calls `BTBridge(datalayer=dl)` and `create_validate_report_tree(...)` both without
  `trigger_activity`. Inside `create_prioritize_subtree` the closures `_build_engage`/`_build_defer`
  check `if factory is None: raise RuntimeError(...)`. This causes `ConstructActivitiesNode` to
  return `FAILURE`, failing the entire `ValidateReportBT`.
- `TriggerService.validate_report` (`service.py` line ~159) calls
  `SvcValidateReportUseCase(self._dl, req).execute()` — it has no `trigger_activity=` kwarg.
  All other similar methods (`invalidate_report`, `submit_report`, etc.) already pass
  `trigger_activity=self._trigger_activity`.
- The CI error string `"Activity construction failed: create_prioritize_subtree: no TriggerActivityPort; cannot build engage_case activity"` matches exactly.

**Active hypotheses**:

- **H1**: After fixing Root Cause 1+2, AC-1 (reporter has owner's `CaseParticipant`) may already
  pass from the `Create(VulnerabilityCase)` delivery in the SubmitReport step — because
  `dl.hydrate()` in `outbox_handler` expands `case_participants` and
  `_store_embedded_participants` is called in `CreateCaseReceivedUseCase._handle_bootstrap`.
  This would make the PCR test config URL fix (Root Cause 3, below) unnecessary for AC-1.
  — *Supporting evidence*: `CreateCaseActivity.update()` runs AFTER `CreateCaseOwnerParticipant.update()`
  saves the case with the participant, so the case object stored in the Create activity should
  already have the owner as a participant entry.
- **H2**: Config cache pollution causes CaseActor ID to use `https://custom.vultron.example/`
  which is not registered in `_TestASGIRouter`, so CaseActor deliveries are dropped.
  This matters for the `Join(VulnerabilityCase)` path (engage-case → CaseActor), but only if
  H1 is wrong and AC-1 actually needs the CaseActor to forward the joined state to the reporter.
  — *Supporting evidence*: `test/adapters/test_utils.py` monkeypatches `VULTRON_SERVER__BASE_URL`
  to `https://custom.vultron.example/` and does NOT call `reload_config()` after restoring env.
  Since `_config_cache` is module-level, it persists across tests unless explicitly cleared.
  `CreateCaseActorNode._resolve_server_base_url()` calls `get_config().server.base_url` at runtime.

**Ruled out**:

- The isolated test apps `configure_globals=False` in `_make_lifespan` — this means the
  lifespan does NOT call `configure_default_emitter`, so the module-level `_default_emitter`
  is NOT overwritten by the app startup and stays as `router` (set by `two_app_setup`).
- The `_TestASGIRouter` can't route the reporter — it IS registered via
  `create_isolated_actor_app(base_url=_REPORTER_BASE, router=router)`.

---

### Current scope

- `vultron/core/use_cases/triggers/report.py` (lines ~95–150) — `SvcValidateReportUseCase`:
  needs `trigger_activity` param added to `__init__` and wired to `BTBridge` and
  `create_validate_report_tree`
- `vultron/core/use_cases/triggers/service.py` (~line 159) — `TriggerService.validate_report`:
  needs `trigger_activity=self._trigger_activity` passed to `SvcValidateReportUseCase`
- `test/demo/test_pcr_engage_case.py` (`two_app_setup` fixture, lines 63–114) — may need
  `monkeypatch.setenv("VULTRON_SERVER__BASE_URL", ...)` + `reload_config()` if H2 is the
  remaining blocker after fixing Root Causes 1+2.

---

### Open questions and blockers

1. **Does AC-1 pass from the `Create(VulnerabilityCase)` step alone, or does it need the
   CaseActor routing fix (H2)?** After fixing Root Cause 1+2, run
   `uv run pytest test/demo/test_pcr_engage_case.py -v --tb=short -m ""` to find out.
   If AC-1 still fails, H2 is the remaining blocker.
2. **Does `VulnerabilityCase.model_dump(mode="json")` serialize `case_participants` as full
   dicts or as ID strings?** The investigation was interrupted here. If the DataLayer round-trip
   stores participants as bare ID strings rather than full dicts, `dl.hydrate()` must look them
   up from the shared DL — which works if the participant was stored before the Create activity
   was queued.
3. **Are there other use cases missing `trigger_activity`?** Compare
   `SvcSubmitReportUseCase`, `SvcCloseReportUseCase`, `SvcInvalidateReportUseCase` as the
   reference pattern for what `SvcValidateReportUseCase` should look like after the fix.

---

### Suggested next steps

1. **Apply Fix 1** — add `trigger_activity: "TriggerActivityPort | None" = None` to
   `SvcValidateReportUseCase.__init__`, store as `self._trigger_activity`, then in `execute()`
   change line 123 to `BTBridge(datalayer=dl, trigger_activity=self._trigger_activity)` and
   add `trigger_activity=self._trigger_activity` to the `create_validate_report_tree(...)` call.
2. **Apply Fix 2** — in `TriggerService.validate_report` (`service.py` ~line 159), change
   `SvcValidateReportUseCase(self._dl, req).execute()` to
   `SvcValidateReportUseCase(self._dl, req, trigger_activity=self._trigger_activity).execute()`.
3. **Run targeted tests** to see if fixes 1+2 are sufficient:
   `uv run pytest test/demo/test_pcr_engage_case.py test/core/use_cases/triggers/ -v --tb=short -m ""`
4. **If AC-1 still fails**, apply Fix 3: add `monkeypatch` param to `two_app_setup`, call
   `monkeypatch.setenv("VULTRON_SERVER__BASE_URL", f"{_OWNER_BASE}/api/v2")` + `reload_config()`
   before creating apps, and `reload_config()` in teardown. Import `reload_config` from
   `vultron.config`.
5. **Run full suite** (demo files were touched):
   `uv run pytest -m "" --tb=short 2>&1 | tail -10`
6. **Format, lint, commit, push**.

---

### Relevant code locations

| File | Lines | Why it matters |
|---|---|---|
| `vultron/core/use_cases/triggers/report.py` | ~95–150 | `SvcValidateReportUseCase` — root cause of both CI failures |
| `vultron/core/use_cases/triggers/service.py` | ~149–165 | `TriggerService.validate_report` — must pass `trigger_activity` |
| `vultron/core/behaviors/report/validate_tree.py` | whole file | `create_validate_report_tree` — already accepts `trigger_activity` (no change needed) |
| `vultron/core/behaviors/report/prioritize_tree.py` | ~185–210 | `create_prioritize_subtree` — where `factory is None` causes FAILURE |
| `test/demo/test_pcr_engage_case.py` | 63–114, 289–332 | `two_app_setup` fixture + AC-1 test — may need config fix |
| `vultron/config.py` | ~198–220 | `_config_cache`, `get_config()`, `reload_config()` — relevant for Fix 3 |
| `vultron/core/behaviors/case/nodes/case_setup.py` | ~285–320 | `CreateCaseActorNode._resolve_server_base_url()` — reads `get_config()` at runtime |
| `vultron/core/use_cases/triggers/case/engage.py` | whole file | `SvcEngageCaseUseCase` — reference for why `SvcValidateReportUseCase` needs the same treatment |
| `test/demo/conftest.py` | 78–200 | `_TestASGIRouter` + `create_isolated_actor_app` — routing details for multi-app tests |

---

### Test status at interrupt

```text
No tests run locally yet — analysis only.

From CI logs (failing):
- test/demo/test_pcr_engage_case.py::TestEngageCaseParticipantExpansion::test_engage_case_stores_owner_participant_in_reporter_dl
  FAILED — assertion: owner CaseParticipant not found in reporter's DataLayer
  Logged error: "Activity construction failed: create_prioritize_subtree: no TriggerActivityPort; cannot build engage_case activity"

- test/demo/test_two_actor_demo.py::TestRunTwoActorDemo::test_full_workflow_succeeds
  FAILED — UnboundLocalError: cannot access local variable 'final_case'
  Caused by AssertionError in _phase_notes_exchange: "Vendor participant RM state did not transition to ACCEPTED"
  Root: same PrioritizeBT failure prevents RM → ACCEPTED transition
```

---

### Notes for the next agent

- **The fix is small and localized.** Root Causes 1+2 are in two files, ~4 lines total to change.
  Model the fix on `SvcInvalidateReportUseCase` (lines 153–200 in `triggers/report.py`) which
  already has the correct `trigger_activity` pattern.
- **AC-2 is already passing** per the WIP commit notes. Only AC-1 and the two-actor demo
  fail. The two-actor demo failure is a direct consequence of PrioritizeBT not running.
- **The `Join(VulnerabilityCase)` goes to the CaseActor, not the reporter.** The test comment
  "delivered to reporter" in `_bootstrap_and_engage` is misleading — the Join is sent to
  `case_manager_id` (the CaseActor). For the reporter to learn the owner's participant, it
  must come from the `Create(VulnerabilityCase)` path (step 1 of `_bootstrap_and_engage`) OR
  from a CaseActor broadcast. Investigate whether AC-1 is satisfied by step 1 alone.
- **Fix 3 (config URL) is speculative.** Apply it only if AC-1 fails after fixes 1+2. The
  config pollution issue is real (`test_utils.py` doesn't call `reload_config()`), but it
  may not block AC-1 if the `Create(VulnerabilityCase)` path carries the participants correctly.
- **Do NOT rebase or force-push.** The branch is attached to an open PR with review history.

---

*Handoff generated by the `bugfix-handoff` skill.*

sei-ahouseholder:
## Root-cause analysis: validate→engage cascade is a business logic error

**Session outcome** from a grill-me interview on the in-progress bugfix.

### Key finding

The difficulty fixing this bug arises from a **process/business logic design error**, not only a missing parameter.

The Vultron RM protocol has two distinct steps:
- `validate-report` advances RM to **VALID** — the receiver decides the report is credible
- `engage-case` (or `defer-case`) advances RM to **ACCEPTED**/**DEFERRED** — a separate, independent decision

These are intentionally distinct. A receiver can validate a report (RM:VALID) and still choose to defer it (RM:DEFERRED) without ever engaging. **Valid does not imply engaged.**

### What went wrong

1. **D5-7-AUTOENG-2** added a procedural `_auto_engage()` helper to `SvcValidateReportUseCase` for demo convenience — validate automatically cascaded to engage.
2. **D5-7-BTFIX-1/2** removed the procedural helper and replaced it with a `PrioritizeBT` subtree appended to `create_validate_report_tree`. This preserved the cascade in BT form.
3. The fact that `SvcValidateReportUseCase` was **never given `trigger_activity`** in either iteration was a latent signal that the cascade was architecturally misplaced.
4. **PR #886** changed how `create_prioritize_subtree` receives its factory (from closure argument instead of BT blackboard), which made the latent failure visible as a hard error.

The fix is not to wire `trigger_activity` through — it's to **remove the cascade**.

### Fix plan

1. Remove the `if case_id and actor_id: append(create_prioritize_subtree(...))` block from `create_validate_report_tree`, and drop the `case_id`/`actor_id`/`trigger_activity` params from that function.
2. Clean up the dead forwarding args in `SvcValidateReportUseCase.execute()`.
3. Add a `coordinator_engages_case` helper to `vultron/demo/helpers/workflow.py`.
4. Update `two_actor_demo.py` and `multi_vendor_demo.py` to explicitly call `engage-case` after `validate-report` (the correct two-step sequence already modelled by `trigger_demo.py` and `demo/cli.py`).
5. Update `test_validate_tree.py` to remove cascade assertions.
6. File a separate issue for the `AcceptInviteActorToCaseReceivedUseCase` PrioritizeBT concern (case owner running PrioritizeBT on the invitee's behalf — a likely PCR-model violation, but not blocking this PR).

Implementation proceeding on this branch.

Issue #878, opened 2026-06-10
Split vultron/core/behaviors/case/nodes/participant.py (816 lines) into focused submodules


sei-ahouseholder:
## Summary

`vultron/core/behaviors/case/nodes/participant.py` is 816 lines and mixes
multiple participant-management concerns. It emerged from the split of
`case/nodes.py` (closed #514) but is itself a candidate for further
decomposition (CS-18-002 — recursive rule, BTND-07-001).

Split into focused submodules by semantic concern (e.g., lookup, role
assignment, status updates, broadcast) within the existing `nodes/`
subpackage. This preserves the tree/node structural split already in place
for `case/` — only the `nodes/` subpackage internals change.

## Acceptance Criteria

- [ ] AC-1: `participant.py` is replaced by two or more focused submodules,
  each under 500 lines (CS-18-002)
- [ ] AC-2: All previously public names re-exported from `nodes/__init__.py`
  (CS-18-003); no call-site changes required
- [ ] AC-3: Test coverage mirrors new submodule structure (CS-18-004)
- [ ] AC-4: All existing tests pass

## Reference

Source: #882, #504
Prior split: #514 (closed)
Spec: `specs/behavior-tree-node-design.yaml` (BTND-07-001, BTND-07-002,
BTND-07-003)
Spec: `specs/code-style.yaml` (CS-18-001 through CS-18-004)

sei-ahouseholder:
Claimed on branch `task/878-split-case-participant-nodes`.

Issue #612, opened 2026-05-21
Production Source Refactoring: Module Splitting and Complexity Reduction


ahouseholder:
## Overview

Split large production source files into focused submodules and reduce cyclomatic
complexity in high-churn areas. Goals: eliminate files exceeding 500 lines that mix
multiple responsibilities, improve code navigability, reduce merge-conflict risk, and
make tests more targeted.

Work spans four layers:

- **Behavior trees** (`vultron/core/behaviors/`): standardize all process areas to
  use `nodes/` subpackages with semantically grouped leaf-node modules
- **Wire layer** (`vultron/wire/as2/`): split pattern definitions, registry, and
  extraction logic in `extractor.py`
- **Adapters** (`vultron/adapters/driven/`): decompose the SQLite DataLayer into
  focused schema, CRUD, and query submodules
- **Use cases** (`vultron/core/use_cases/received/`): split large received
  use-case modules by semantic cluster
- **Outbox handler** (`vultron/adapters/driving/fastapi/outbox_handler.py`):
  reduce cyclomatic complexity and add targeted delivery-path unit tests

## Open Tasks

- #873 — Add unit tests for outbox delivery-path scenarios (recipient extraction,
  dehydration, validation)
- #874 — Refactor `outbox_handler.py` to reduce cyclomatic complexity (CC ≤ 10)
- #876 — Split `behaviors/status/nodes.py` (1281 lines) into `nodes/` subpackage
- #877 — Split `behaviors/embargo/nodes.py` (920 lines) and merge
  `trigger_nodes.py` into `nodes/` subpackage
- #878 — Split `behaviors/case/nodes/participant.py` (816 lines) into focused
  submodules within the existing `nodes/` subpackage
- #879 — Split `adapters/driven/datalayer_sqlite.py` (1296 lines) into focused
  submodules
- #880 — Split `wire/as2/extractor.py` (826 lines) into focused submodules;
  preserve registry order and import-time guard
- #881 — Split `use_cases/received/case.py` (819 lines) and
  `received/actor.py` (803 lines) into focused submodules
- #883 — Convert `behaviors/note/nodes.py` and `behaviors/sender/nodes.py`
  to `nodes/` subpackages following the standard pattern

## Completed Work

- #504 — Initial survey: identified oversized mixed-responsibility files ✓
- #514 — Split `behaviors/case/nodes.py` (1502 lines) into `nodes/` subpackage ✓
[...]

PR #887, opened 2026-06-10
Decompose CreateCaseActor and owner-participant BT nodes

### LINKED ISSUE #714: God node decomposition: split CreateCaseActorNode, InitializeDefaultEmbargoNode, CreateCaseOwnerParticipant, and BroadcastStatusToPeersNode into composed subtrees
## Summary

Two BT leaf nodes in `vultron/core/behaviors/case/` have `update()` methods with
3–6 distinct responsibilities that should each be decomposed into a named BT subtree
of simple leaf nodes (per BTND-07-001 god-node rule).

Affected nodes (both now in the `nodes/` subpackage after #736):

- **`CreateCaseActorNode.update()`** (~83 lines): idempotency check, ID slug
  construction, server URL read, CaseActor creation, Participant creation, case
  registration → ~4 leaf nodes
- **`CreateCaseOwnerParticipant.update()`** (~58 lines): status build, participant
  creation, case attach, RM advancement, event recording → ~4 leaf nodes

### CreateCaseActorNode → CreateCaseActorBT
```
CreateCaseActorBT (Sequence)
├── ResolveCaseActorUrlsNode          # derive slug + base URL → blackboard
├── CheckCaseActorParticipantAbsentNode  # idempotency condition
├── CreateCaseActorServiceNode        # create VultronCaseActor → blackboard
└── RegisterCaseActorParticipantNode  # create participant, attach to case, save
```

### CreateCaseOwnerParticipant → CreateCaseOwnerParticipantBT
```
CreateCaseOwnerParticipantBT (Sequence)
├── ResolveOwnerAcceptedStatusNode    # find or create accepted CaseParticipantStatus
├── CreateOwnerParticipantNode        # create VultronCaseParticipant → blackboard
├── AttachOwnerParticipantToCaseNode  # case.participants.append + save
└── RecordOwnerJoinedEventNode        # case.record_event(...)
```

## Acceptance Criteria

- [ ] AC-1: `CreateCaseActorNode` decomposed into ≥3 named leaf nodes in a
  factory subtree; original monolithic `update()` removed
- [ ] AC-2: `CreateCaseOwnerParticipant` decomposed into ≥3 named leaf nodes;
  boolean `advance_to_accepted` replaced with a conditional node
- [ ] AC-3: All new nodes have unit tests; existing integration tests pass

## Reference

Notes: `notes/bt-integration.md`
Spec: `specs/behavior-tree-node-design.yaml` BTND-07-001, BTND-07-002
Related: #736 (case/nodes.py split — prerequisite, already done)
See also: New issue for BroadcastStatusToPeersNode (status/), new issue for InitializeDefaultEmbargoNode (blocked by #538)

sei-ahouseholder:
- Closes #714

## Summary

This decomposes the remaining in-scope god nodes from #714 into explicit composed BT subtrees while preserving idempotent behavior and existing case-creation call paths.

## Changes

- vultron/core/behaviors/case/nodes/case_setup.py: replaced monolithic CreateCaseActorNode.update() with a composed subtree of named leaves for URL resolution, idempotency reuse, service creation, and participant registration.
- vultron/core/behaviors/case/nodes/participant.py: replaced monolithic CreateCaseOwnerParticipant.update() with a composed subtree and explicit conditional RM-advance branch nodes.
- test/core/behaviors/case/nodes/test_case_setup.py: added subtree-composition assertions for CreateCaseActorNode.
- test/core/behaviors/case/nodes/test_participant.py: added subtree-composition assertions and advance_to_accepted behavior coverage for CreateCaseOwnerParticipant.
- plan/BUILD_LEARNINGS.md: recorded ISSUE-714 decomposition pitfall about preserving alternate context seams.

## Verification

- 3148 passed, 14 skipped, 219 deselected, 5633 subtests passed (uv run pytest --tb=short 2>&1 | tail -5)
- Black, flake8, mypy, and pyright clean

### DIFF
--- a/plan/BUILD_LEARNINGS.md
+++ b/plan/BUILD_LEARNINGS.md
@@ -173,3 +173,10 @@ header.
   collect context, derive index/recipients, then perform side effects.
 - Condition+action hybrid nodes are clearer and safer as `Selector` composites:
   a pure condition leaf first, then side-effect action fallback.
+
+### 2026-06-10 ISSUE-714 — Decomposed BT nodes must preserve alternate context seams
+
+- When replacing a god node with leaf-node sequences, preserve all original
+  input seams (`case_id` from blackboard and `case_obj`-derived context).
+- If downstream leaves rely on blackboard keys, add explicit fallback reads
+  from staged objects/status context to avoid regressing valid call paths.
--- a/plan/history/2606/implementation/ISSUE-714.md
+++ b/plan/history/2606/implementation/ISSUE-714.md
@@ -1,0 +1,12 @@
+---
+source: ISSUE-714
+timestamp: '2026-06-10T18:14:57.614505+00:00'
+title: Decompose CreateCaseActor and owner participant BT nodes
+type: implementation
+---
+
+## Issue #714 — God node decomposition: split CreateCaseActorNode, InitializeDefaultEmbargoNode, CreateCaseOwnerParticipant, and BroadcastStatusToPeersNode into composed subtrees
+
+Completed the in-scope decomposition for `CreateCaseActorNode` and `CreateCaseOwnerParticipant` as composed BT subtrees with named leaf nodes, preserving idempotency and call-path compatibility. Added unit coverage for subtree structure and owner RM advancement behavior.
+
+PR: [#887](https://github.com/CERTCC/Vultron/pull/887)
--- a/test/core/behaviors/case/nodes/test_case_setup.py
+++ b/test/core/behaviors/case/nodes/test_case_setup.py
@@ -405,3 +405,28 @@ class TestCreateCaseActorNodeBlackboard:
             # No case_id supplied
         )
         assert result.status == py_trees.common.Status.FAILURE
+
+
+# ---------------------------------------------------------------------------
+# RegisterCaseActorParticipantNode — precondition failure tests
+# ---------------------------------------------------------------------------
+
+
+class TestRegisterCaseActorParticipantNode:
+    """RegisterCaseActorParticipantNode returns FAILURE when case is absent."""
+
+    def test_fails_when_case_not_in_datalayer(
+        self,
+        bt_scenario: BTTestScenario,
+        actor: VultronCaseActor,
+        actor_id: str,
+    ) -> None:
+        """Node returns FAILURE (not SUCCESS) when the case record is missing."""
+        result = bt_scenario.run(
+            RegisterCaseActorParticipantNode(),
+            actor_id=actor_id,
+            case_id="https://example.org/cases/nonexistent",
+            case_actor_id=f"{actor_id}/case-actor",
+            case_actor_participant_id=f"{actor_id}/case-actor/participant",
+        )
+        assert result.status == py_trees.common.Status.FAILURE
--- a/test/core/behaviors/case/nodes/test_case_setup.py
+++ b/test/core/behaviors/case/nodes/test_case_setup.py
@@ -39,6 +39,12 @@ from vultron.core.behaviors.case.nodes import (
     RecordOfferReceivedEventNode,
     UpdateActorOutbox,
 )
+from vultron.core.behaviors.case.nodes.case_setup import (
+    CreateCaseActorServiceNode,
+    RegisterCaseActorParticipantNode,
+    ResolveCaseActorUrlsNode,
+    ReuseExistingCaseActorParticipantNode,
+)
 from vultron.core.behaviors.helpers import (
     UpdateActorOutbox as UpdateActorOutboxHelper,
 )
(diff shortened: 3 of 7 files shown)


sei-ahouseholder (commented):
## Review: PR #887 — Decompose CreateCaseActor and owner-participant BT nodes

Reviewed against issue #714 ACs, `specs/behavior-tree-node-design.yaml` (BTND-07-001–BTND-07-003), `specs/code-style.yaml` (CS-18-002), and `notes/bt-integration.md`.

---

### ✅ AC compliance

| AC | Status | Notes |
|----|--------|-------|
| AC-1: CreateCaseActorNode decomposed into ≥3 named leaves, monolithic `update()` removed | ✅ | 4 named leaf classes + idempotency Selector |
| AC-2: boolean `advance_to_accepted` replaced with a conditional node | ✅ | `ShouldAdvanceOwnerToAcceptedNode` condition leaf |
| AC-3: new nodes have unit tests; existing integration tests pass | ✅ | Subtree-composition assertions + `advance_to_accepted` behavioral test; 3148 passed |

---

### 🔴 Critical: file-size violations (CS-18-002 / BTND-07-002)

```
participant.py   1163 lines   ← well over 500-line threshold
case_setup.py     506 lines   ← marginally over threshold
```

The PR adds ~275 lines to `participant.py`, pushing it to 1163 lines — more than twice the SHOULD-split threshold. Per CS-18-002 and BTND-07-002, any `nodes/` subpackage file approaching or exceeding 500 lines MUST be further split into sub-subpackages grouped by semantic concern. A follow-up issue should be opened before merging (or this PR should include the split). The `case_setup.py` overage is minor but also warrants noting.

---

### 🟡 BTND-07-003 — Composite subtree classes in leaf-node files

`CreateCaseActorNode` (`case_setup.py`) and `CreateCaseOwnerParticipant` (`participant.py`) are `py_trees.composites.Sequence` subclasses that perform tree assembly/composition. BTND-07-003 says tree-factory/composition classes belong in `*_tree.py` modules at the area root, not in the `nodes/` subpackage:

> "Tree factory modules (files named `*_tree.py` at the area root) define BT composition and structure. Leaf node modules (files in the `nodes/` subpackage) define BT `Behaviour` subclasses."

Note: `RecordCaseCreationEvents` and `CreateCaseParticipantNode` have the same issue (pre-existing), so this PR adds two more violations to an existing pattern. Both this PR and the pre-existing instances are BTND-07-003 violations that should be tracked — either move the composite classes to a `*_tree.py` file or open a follow-up issue.

---

### 🟡 `RecordOwnerJoinedEventNode` — proposed by issue but not implemented

Issue #714 explicitly proposed this as leaf #4 of `CreateCaseOwnerParticipantBT`:

```
└── RecordOwnerJoinedEventNode        # case.record_event(...)
```

The PR does not include it. By contrast, the non-owner path (`CreateCaseParticipantNode`) does include `RecordParticipantAddedEventNode`, which calls `case.record_event(participant_id, "participant_added")`. The owner-participant path has no equivalent event recording in the BT.

Per `notes/bt-integration.md`: "All Protocol-Significant Behavior MUST Be in the BT" — case history event recording is protocol-significant. This is a pre-existing gap (the original monolithic `CreateCaseOwnerParticipant.update()` also didn't call `record_event`), so this PR doesn't introduce a regression. However, the issue proposed closing that gap and the PR defers it silently. This should either be implemented here or tracked with an explicit follow-up issue.

---

### 🟡 `RegisterCaseActorParticipantNode` returns SUCCESS when case not found

```python
# case_setup.py ~line 449
if not is_case_model(case):
    self.logger.warning(...)
    return Status.SUCCESS   # ← case absent but tree continues
```

The original `_register_case_actor_participant` helper silently returned `None` here (void), so this preserves backward-compatible behavior. But returning SUCCESS when the case doesn't exist means a real precondition failure is swallowed. Since this node runs only after the case was just created, a missing case is unexpected and probably warrants FAILURE (or at minimum a comment justifying the SUCCESS). The original silence was arguably a bug; the PR has an opportunity to fix it here.

---

### 📝 Minor observations

- `_derive_case_slug` and `_resolve_server_base_url` correctly promoted to module-level helpers (was previously `self._derive_case_slug` / `self._resolve_server_base_url` methods — cleaner as pure functions). ✅
- Logging format strings converted from f-strings to `%s`-style throughout — correct. ✅
- `AdvanceOwnerRmToAcceptedNode` always returns SUCCESS whether RM advance succeeded or not, consistent with the original `_save_owner_case` warning-but-continue behavior. The Selector fallback `Success` node fires only when `advance_to_accepted=False`, so the tree semantics are sound. ✅

---

### Summary

ACs are met and the core decomposition work is solid. The primary concern is the **CS-18-002 file-size violation** in `participant.py` (1163 lines) — this should be addressed (either here or via an immediately-opened follow-up issue) before merging. The missing `RecordOwnerJoinedEventNode` should also be tracked explicitly to prevent the pre-existing gap from being silently forgotten. The BTND-07-003 composite-in-leaf-file pattern is worth tracking but is pre-existing and lower priority.

sei-ahouseholder:
## Review response (cbfe74f8)

### 🔴 RegisterCaseActorParticipantNode returns SUCCESS when case not found — **Fixed**

Changed to return `FAILURE` in cbfe74f8. A missing case after creation is
an unexpected precondition failure, not a silent no-op.
Test added: `TestRegisterCaseActorParticipantNode.test_fails_when_case_not_in_datalayer`.

---

### 🟡 RecordOwnerJoinedEventNode not implemented — **Fixed**

Added `RecordOwnerJoinedEventNode` and wired it into `CreateCaseOwnerParticipant`
as child index 4 (after `PersistOwnerCaseNode`, before the RM-advance Selector)
in cbfe74f8. Records `'owner_joined'` on the case, analogous to
`RecordParticipantAddedEventNode` in the non-owner path. Tests updated/added
in `test_participant.py`.

---

### 🔴 CS-18-002 — participant.py file-size violation — **Tracked: #878**

`participant.py` at 1163 lines is already tracked by **#878**
("Split `vultron/core/behaviors/case/nodes/participant.py` into focused submodules"),
a child of the #612 refactoring epic. That issue predates this PR; the additional
275 lines added here make it higher priority. `case_setup.py` at 506 lines
will also be addressed as part of that work.

---

### 🟡 BTND-07-003 — composite Sequence/Selector subclasses in nodes/ — **Tracked: #889**

Opened **#889** ("Move composite Sequence/Selector subclasses from nodes/ to
`*_tree.py` modules (BTND-07-003)") as a child of #612 to track
`CreateCaseActorNode`, `CreateCaseOwnerParticipant`, `CreateCaseParticipantNode`,
and the pre-existing `RecordCaseCreationEvents`.

PR #960, opened 2026-06-15
Add emConsentState and cvdRole to ParticipantStatus snapshots

### LINKED ISSUE #931: Add emConsentState and cvdRole fields to ParticipantStatus schema and populate at write sites
## Summary

Concern #923 Finding 12 + 13:

- **Finding 12:** Every `ParticipantStatus` object carries `embargoAdherence: true` (a boolean) but the 5-state embargo consent machine (NO_EMBARGO → INVITED → SIGNATORY → LAPSED → DECLINED) is never recorded. No `emConsentState` field appears on any status object in any log.
- **Finding 13:** `cvdRole` is recorded on the enclosing `CaseParticipant` within the `Add` activity, not on `ParticipantStatus` itself. When a status snapshot is read in isolation (e.g., during replay from the canonical log), there is no way to know the participant's role at that point in time.

Both fields are required for a participant's protocol-significant state at any given log position to be self-contained.

## Acceptance Criteria

- [ ] AC-1: Add `emConsentState` field (typed by the existing 5-state consent enum) to `ParticipantStatus`. Required if a value is present (per "if present, then non-empty" rule); MAY be absent only if not applicable.
- [ ] AC-2: Add `cvdRole` field to `ParticipantStatus` (typed by `CVDRole` enum) and populate it at every `ParticipantStatus` construction site (BT node + trigger + received use cases).
- [ ] AC-3: Update all `ParticipantStatus` constructors and builders to require `cvdRole`; update existing call sites to pass it.
- [ ] AC-4: Existing `embargoAdherence` boolean is preserved (or deprecated explicitly) — document the choice in the PR description.
- [ ] AC-5: Unit tests verify both fields are populated correctly across the full assertion→commit→broadcast→replica-replay path.
- [ ] AC-6: Flips #925 xfail invariant #9 (emConsentState + cvdRole present on every ParticipantStatus) to passing.

## Reference

Source concern: #923
Parent epic: #788
Test harness: #925 (this PR MUST flip the named xfail to passing)
Docs PR: https://github.com/CERTCC/Vultron/pull/924
ADR: `docs/adr/0019-separate-case-ledger-from-process-log.md`
Spec: `specs/case-ledger-processing.yaml` (CLP-07)
Notes: `notes/case-ledger-authority.md`

sei-ahouseholder:
- Closes #931

## Summary

Adds emConsentState and cvdRole to ParticipantStatus and populates them across core and wire write sites so participant status snapshots are self-contained for canonical ledger replay.

## Changes

- vultron/core/models/participant_status.py and vultron/wire/as2/vocab/objects/case_status.py: add typed em_consent_state and cvd_role fields with emConsentState and cvdRole wire aliases and enum coercion.
- vultron/core/models/case_participant.py and vultron/wire/as2/vocab/objects/case_participant.py: populate new fields when seeding and appending statuses, and keep latest status metadata aligned when participant roles change.
- vultron/core/behaviors/case/nodes/participant/*, vultron/core/behaviors/report/nodes/rm_transitions.py, vultron/core/use_cases/received/case/_helpers.py, vultron/wire/as2/extractor/_builders.py, vultron/demo/exchange/*.py: populate new fields at status write paths and preserve existing consent snapshots on idempotent reuse.
- test/ci/test_case_ledger_invariants.py: promote invariant #9 by removing xfail and validating presence plus enum-valid values for both fields.
- test/ci/README-case-log-ratchet.md: mark invariant #9 as passing.
- test/core/behaviors/case/nodes/participant/test_participant_add.py: add regression test that existing accepted-status consent is preserved.
- plan/BUILD_LEARNINGS.md: add build observation for consent preservation in accepted-status reuse.

## Verification

- 3221 passed, 34 skipped, 225 deselected, 2 xfailed, 5633 subtests passed


test/ci/test_case_ledger_invariants.py
```diff
@@ -549,21 +552,10 @@ def test_invariant_8_late_joiner_has_full_history(
...
-)
 def test_invariant_9_participant_status_schema_completeness(
     case_ledger_replicas: dict[str, list[dict]],
 ) -> None:
-    """Every ParticipantStatus snapshot includes emConsentState and cvdRole (AC-4.9).
-
-    When this xfail is unexpectedly promoted to XPASS, remove the
-    ``xfail`` decorator to make it a permanent regression guard.
-    """
+    """Every ParticipantStatus snapshot includes emConsentState and cvdRole (AC-4.9)."""
```
sei-ahouseholder: This may actually require a spec update too, but: a CaseParticipant can have more than one cvdRole at a time, and the expectation is that this ParticipantStatus would reflect the full set of roles they have when the status is created. So the datastructure assumption in the current changeset that cvd_role is a singular value is incorrect. It needs to be a list to parallel CaseParticipant

vultron/core/models/participant_status.py
```diff
@@ -21,10 +21,45 @@
...
+    if isinstance(value, PEC):
+        return value
+    if isinstance(value, str):
+        return PEC[value]
+    raise TypeError(
+        f"Unsupported em_consent_state type: {type(value).__name__}"
+    )
+
+
+def primary_cvd_role(roles: object) -> CVDRole:
```
sei-ahouseholder: We shouldn't be doing this. The value should support multiple roles, not prioritize them this way.

vultron/core/models/participant_status.py
```diff
@@ -49,6 +84,16 @@ class ParticipantStatus(CoreObject):
     vfd_state: CS_vfd = CS_vfd.vfd
     case_engagement: bool = True
     embargo_adherence: bool = True
+    em_consent_state: PEC | None = Field(
+        default=None,
+        validation_alias="emConsentState",
+        serialization_alias="emConsentState",
+    )
+    cvd_role: CVDRole = Field(
```
sei-ahouseholder: this should parallel whatever is in CaseParticipant. I think that's: `cvd_roles: list[CVDRole]` or something adjacent. Point being that this is a multi-valued attribute, not singular.

sei-ahouseholder (commented):
main concern is that `cvd_role` shouldn't be singular, it should be a list like that on `CaseParticipant` since it's snapshotting the participant's role(s) at the time of the status object's creation.

test/ci/test_case_ledger_invariants.py
sei-ahouseholder: Addressed in 3e96fbd7: invariant 9 now treats `cvdRole` as a required non-empty list and validates every role entry against `CVDRole` values.

vultron/core/models/participant_status.py
sei-ahouseholder: Addressed in 3e96fbd7: removed role-priority reduction and introduced `coerce_cvd_roles()` so ParticipantStatus carries the full role set instead of a single prioritized role.

vultron/core/models/participant_status.py
sei-ahouseholder: Addressed in 3e96fbd7: `ParticipantStatus.cvd_role` is now `list[CVDRole]` (serialized as `cvdRole` list) to parallel CaseParticipant multi-role semantics.

PR #1355, opened 2026-07-10
fix: suppress context and pytest griffe false-positive warnings

### LINKED ISSUE #1329: [Docs] mkdocs build: suppress `context` and `pytest` griffe false-positive warnings
## Summary

`mkdocs build --strict` reports two recurring griffe false-positive warnings
that are not suppressed by `.github/scripts/mkdocs-build-strict.sh`:

```
WARNING -  Inline reference to unknown key context
WARNING -  Inline reference to unknown key pytest
```

These appear twice each per build (4 warnings total). They follow the same
pattern as already-suppressed false positives (`@dataclass`, `@main`, `@prefix`,
`@base`) where griffe's docstring parser misinterprets Python decorators or
parameter names as bibliography citations.

`context` and `pytest` appear in docstrings (e.g., `@pytest.fixture`,
`context=...` parameter docs) and are parsed by griffe as unknown citation keys.

## Acceptance Criteria

- [ ] Add `context` and `pytest` to the false-positive suppression grep pattern
  in `.github/scripts/mkdocs-build-strict.sh`
- [ ] `mkdocs build --strict` (via the wrapper script) no longer reports these
  as real warnings
- [ ] No legitimate warnings are newly suppressed

## Fix

In `.github/scripts/mkdocs-build-strict.sh`, update the suppression pattern:

```bash
# Before
FALSE=$(grep "^WARNING -  Inline reference to unknown key" "$TEMP_OUTPUT" | \
  grep -cE "(petterogren7535|main|v4|dataclass|prefix|base)" || echo 0)

# After
FALSE=$(grep "^WARNING -  Inline reference to unknown key" "$TEMP_OUTPUT" | \
  grep -cE "(petterogren7535|main|v4|dataclass|prefix|base|context|pytest)" || echo 0)
```

sei-ahouseholder:
- Closes #1329

## Summary

Adds `context` and `pytest` to the false-positive suppression pattern in
`.github/scripts/mkdocs-build-strict.sh`. These two keywords are
misinterpreted as bibliography citation keys by `mkdocs_bibtex` when
griffe encounters `@context` and `@pytest` in docstrings (e.g.,
`context=...` parameter docs, `@pytest.fixture` decorator references).

## Changes

- `.github/scripts/mkdocs-build-strict.sh`: extend `grep -cE` alternation
  to include `context` and `pytest` alongside existing patterns
  (`dataclass`, `prefix`, `base`, etc.)

## Verification

- `bash .github/scripts/mkdocs-build-strict.sh` now exits 0 with
  "Suppressed N false-positive warnings" (previously exited 1 with
  "5 real warning(s)")
- All 4486 unit tests pass (0 new)

### DIFF
--- a/.github/scripts/mkdocs-build-strict.sh
+++ b/.github/scripts/mkdocs-build-strict.sh
@@ -22,7 +22,7 @@ uv run mkdocs build --strict 2>&1 | tee "$TEMP_OUTPUT" || true
 TOTAL=$(grep -c "^WARNING -" "$TEMP_OUTPUT" || echo 0)
 
 # Count false-positive warnings (known decorator/keyword names)
-FALSE=$(grep "^WARNING -  Inline reference to unknown key" "$TEMP_OUTPUT" | grep -cE "(petterogren7535|main|v4|dataclass|prefix|base)" || echo 0)
+FALSE=$(grep "^WARNING -  Inline reference to unknown key" "$TEMP_OUTPUT" | grep -cE "(petterogren7535|main|v4|dataclass|prefix|base|context|pytest)" || echo 0)
 
 # Count print-site warnings (version-specific false positive)
 PRINT=$(grep -c "^WARNING -  \[mkdocs-print-site\]" "$TEMP_OUTPUT" || echo 0)
--- a/plan/history/2607/implementation/ISSUE-1329.md
+++ b/plan/history/2607/implementation/ISSUE-1329.md
@@ -1,0 +1,12 @@
+---
+source: ISSUE-1329
+timestamp: '2026-07-10T20:52:42.316781+00:00'
+title: suppress context and pytest griffe false-positive warnings in mkdocs build
+type: implementation
+---
+
+## Issue #1329 — suppress context and pytest griffe false-positive warnings
+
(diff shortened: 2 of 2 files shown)


sei-ahouseholder:
## Code Review — Advisory Finding

**[ADVISORY]** `.github/scripts/mkdocs-build-strict.sh` line 25: unanchored substring patterns

The `context` and `pytest` entries in `grep -cE` are unanchored substrings. If a future docstring or `.md` file adds an inline `@`-reference containing these as a substring (e.g., `@context_id`, `@pytest_plugin`), the `mkdocs_bibtex` warning would be counted as a false positive and silently suppressed.

**Current risk:** None — verified that no compound `@context_*` or `@pytest_*` inline references exist in the current codebase.

**Suggested long-term fix:** When the list grows further, consider suppressing ALL `Inline reference to unknown key` warnings at the `mkdocstrings` layer (via `griffe_warnings_filters` in `mkdocs.yml`) rather than maintaining a growing keyword allowlist in the shell script.

Also note: `plan/incoming/learnings/20260710-mkdocs-nav-omitted-files-guard.md` mentions `context` and `pytest` as "genuine pre-existing warnings" that cause the strict wrapper to exit non-zero — that learning file is now stale after this fix and should be updated on the next `learn` skill run.

Issue #1446, opened 2026-07-15
Migration guard: detect already-persisted entries with unapplied effects (SYNC-12)


sei-ahouseholder:
## Context

After #1447 (SYNC-13), the ingress adapter no longer writes `CaseLedgerEntry` to the DataLayer. A stored entry now reliably implies its effects have been applied — the stale `{entry: stored, effects: not-applied}` state this issue guarded against cannot be created by current code.

There are no extant running nodes or cases holding stale state (this was pre-production code). A repair path is therefore unnecessary.

## Work Remaining

- AC-2: Document the upgrade path in `notes/sync-ledger-replication.md`: nodes that ran pre-SYNC-13 code and have stale `{entry: stored, effects: not-applied}` state must wipe their local DataLayer and re-sync from the CaseActor.

## References

- PR #1447 / SYNC-13 — enforces core-only ledger writes
- `specs/sync-ledger-replication.yaml` SYNC-12-003, SYNC-13
- `vultron/core/behaviors/sync/announce_tree.py`

sei-ahouseholder:
Context update from the SYNC-13 fix in PR #1447 (commit b088f4a3): the effects-before-persist invariant is now enforced with the ledger written **only** by core (`PersistReceivedLogEntry`), never by the ingress adapter (SYNC-13-002). This removes the class of stale `{entry: stored, effect: not-applied}` state that this issue guards against for the *newly-fixed* code path — a persisted entry now genuinely implies its effects ran.

This issue's migration concern still stands for **nodes that ran the old persist-before-effects code** and already hold stale derived state; SYNC-13 does not retroactively repair those. AC-1 (accept vs. guard) is still open. Recommend keeping this open and deciding the migration approach separately from #1447.

sei-ahouseholder:
Claimed on branch `task/1446-migration-guard-accept`.

Issue #1394, opened 2026-07-13
Architecture Hardening: compliance, migration, and test-coverage ratchets


sei-ahouseholder:
## Summary

Epic tracking architecture compliance gaps and migration work surfaced during the
2026-07-13 gap analysis. These are correctness and structural hygiene items —
no behavioral changes — that tighten ratchets and close known violations.

## Scope

- [ ] #1385 — Use `CasePersistence` in `AcceptCaseProposalReceivedUseCase` and `RejectCaseProposalReceivedUseCase` (DL-03-001/DL-03-002 gap)
- [ ] #1386 — Add dedicated `execute()`-path tests for `SvcValidateReportUseCase` and `SvcSubmitReportUseCase`
- [ ] #1387 — Clear ARCH-14-001 `KNOWN_VIOLATIONS` — apply `as_` prefix to 16 remaining wire vocab objects (follow-up after #1056 closed)
- [ ] #1388 — Replace direct `case_roles` `getattr()` reads with `.roles` property in core (PRM-01-003, 9 violations)
- [ ] #1401 — Forward Offer `content` in AC-6 Note to fully satisfy CM-16-008 (surfaced by PR #1398 review)
- [ ] #1402 — Split `suggest_actor_tree.py` (778 lines) into `suggest_actor/` subpackage per BTND-07-004/CS-18-002 (surfaced by PR #1398 review)

## Motivation

The glossary defines "Priority 473 — Architecture Hardening" as an initiative to
deepen module design, reduce coupling, and improve testability. This epic
operationalises that initiative for the gaps found in the July 2026 gap analysis.
- [ ] #1474 — Migrate `core/services/embargo_lifecycle.py` guards to BT nodes
- [ ] #1475 — Migrate `vultron/demo/` guards to staged types and predicates

PR #1447, opened 2026-07-15
feat(#1324): enforce effects-before-persist invariant in AnnounceLogEntryReceivedBT

### LINKED ISSUE #1324: Spec + impl: effects-before-persist ordering in Announce(CaseLedgerEntry) receive tree
## Summary

In `create_announce_log_entry_tree()`, side effects that mutate derived local state (EM state, participant list, notes) are applied **after** the ledger entry is persisted. This violates the invariant that entry presence in the local ledger should be evidence that its effects were successfully applied.

Discovered during review of PR #1317 (BTND-08 refactor). The PR is correct — the existing persist-before-effects ordering predates it. The advisory at https://github.com/CERTCC/Vultron/pull/1317#issuecomment-4935836384 identified an `AlwaysSuccess` masking concern; this issue captures the root cause.

## The Invariant

If an actor has a ledger entry stored locally, it MUST have also applied that entry's protocol-significant side effects. Entry presence is not just a hash-chain record — it is a commitment that the actor has integrated the effect.

## Current (incorrect) ordering in `announce_tree.py`

```text
ParticipantGate (Sequence)
  CheckIsNotOwnCaseActorNode
  ParticipantSubtree (Selector)
    CheckLedgerEntryAlreadyStored   <- if stored, skip persist AND effects
    ValidateAndPersistFlow (Seq)    <- 1. hash-check  2. PERSIST ENTRY
  LogEntryEventEffects (Sequence)   <- 3. effects run AFTER persist
    Selector(IsRemoveEmbargo, ApplyEmbargoTeardown, AlwaysSuccess)
    ... (3 more slots)
```

If any `Apply*` node fails, `AlwaysSuccess` fires, the outer Sequence sees SUCCESS, and the entry is already persisted with wrong derived state.

## Correct ordering

```text
ParticipantGate (Selector)
  CheckLedgerEntryAlreadyStored      <- if stored -> SUCCESS (effects already applied)
  ProcessAndStore (Sequence)
    ValidateChain                    <- hash-check only (no persist yet)
    LogEntryEventEffects (Sequence)  <- effects first; FAILURE propagates
      Seq(IsRemoveEmbargo, ApplyEmbargoTeardown)   <- no AlwaysSuccess fallback
      ... (3 more slots, no AlwaysSuccess)
    PersistReceivedLogEntry          <- LAST: only runs if all effects succeeded
```

Because all `Apply*` nodes are idempotent, a failed delivery can be replayed safely.

`AlwaysSuccess` is retained only for the routing no-op path (`IsFoo` returns FAILURE = wrong event type). It must NOT mask `Apply*` FAILURE.

## Acceptance Criteria

- [ ] AC-1: Add a spec entry to `specs/sync-ledger-replication.yaml` (SYNC-12 or addendum to SYNC-02) stating the effects-before-persist invariant.
- [ ] AC-2: Restructure `create_announce_log_entry_tree()` so effects run before persist.
- [ ] AC-3: Remove `AlwaysSuccess` as the fallback for `Apply*` FAILURE in each effect slot — `Apply*` FAILURE must propagate and block persist.
- [ ] AC-4: `CheckLedgerEntryAlreadyStored` must gate the entire entry processing (effects + persist), not just the persist path.
- [ ] AC-5: All existing tests pass; add tests verifying that `Apply*` FAILURE prevents the entry from being persisted.

## Note on BTND-08

`AlwaysSuccess` in BTND-08 is correct as a routing no-op (fires when `IsFoo` returns FAILURE = wrong type). The spec is not wrong. The Precondition-Sequence pattern in BTND-08 is unaffected by this fix — the issue is the tree-level ordering, not the slot-level pattern.

## Reference

- PR #1317 (BTND-08 refactor, correct; this ordering issue predates it)
- `vultron/core/behaviors/sync/announce_tree.py`
- `specs/sync-ledger-replication.yaml` SYNC-02, SYNC-08-003
- `specs/behavior-tree-node-design.yaml` BTND-08

sei-ahouseholder:
- Closes #1324

## Summary

Restructures `create_announce_log_entry_tree()` so domain effects run **before** `PersistReceivedLogEntry`, and replaces `Success("...Skipped")` fallbacks with `Inverter(IsX)` so any `Apply*` FAILURE propagates and blocks persist. Adds the SYNC-12 spec group codifying the invariant.

This PR is now scoped to the #1324 behavior-tree work **only**. An adapter-layer changeset (rehydrate/parse-store) that had been added mid-branch to chase an FVV demo regression was **reverted** (commit `41840381`) — it regressed 6 demo/integration tests and did not fix its target scenario in the integration CI jobs. That work is tracked separately in **#1472**.

## Changes

- **`vultron/core/behaviors/sync/announce_tree.py`**: Reorder effects before persist inside a new `ProcessAndStore` Sequence; `CheckLedgerEntryAlreadyStored` now gates the entire path (effects+persist) via the `EntryProcessing` Selector; replace `Success("Skipped")` fallbacks with `Inverter(IsX)` to propagate `Apply*` FAILURE (SYNC-12-001).
- **`specs/sync-ledger-replication.yaml`**: Add SYNC-12 group (Effects-Before-Persist Invariant) with three MUST specs.
- **`vultron/core/behaviors/sync/nodes/conditions.py`**: Update `IsX` node docstrings to document the Inverter pattern (replaces stale `AlwaysSuccess("Skipped")` references).
- **`test/core/behaviors/sync/test_announce_tree.py`**: Update already-stored tests; add `TestAnnounceLogEntryAppliesParticipantStatus`; add `TestEffectsFailureBlocksPersist` (4 tests).

## Verification

- Full pytest suite (`uv run pytest -m ""`, including demo/integration tests) passes locally with **0 failures** (exit 0) on the reverted branch — the 6 demo tests that were red before the revert now pass.
- Black + flake8 clean (pre-commit).
- [x] AC-1: SYNC-12 spec group added to `specs/sync-ledger-replication.yaml`
- [x] AC-2: `create_announce_log_entry_tree()` restructured — effects before persist
- [x] AC-3: `Apply*` FAILURE propagates (Inverter replaces Success fallback)
- [x] AC-4: `CheckLedgerEntryAlreadyStored` gates entire processing path
- [x] AC-5: New `TestEffectsFailureBlocksPersist` tests verify invariant

## Follow-ups

- **#1472** — the rehydration/semantic-routing bug for `Announce(CaseLedgerEntry)` (root-caused during this PR's review). The correct fix belongs in the canonical DataLayer rehydration pipeline, not an ad-hoc adapter bypass.
- **#1446** — migration guard for nodes that ran the old persist-before-effects code and hold `{entry: stored, effect: not-applied}` state.


sei-ahouseholder (commented):
## PR Review — `feat(#1324): enforce effects-before-persist invariant` (updated after fixes applied)

Reviewed against issue #1324, project specs/notes, AGENTS.md rules, and CI. Focus was placed on `inbox_orchestration.py` as requested. **Findings were addressed in-session; this PR is now green and on-scope.**

### What was wrong (now fixed)

**FAIL — CI red / Verification inaccurate.** The original body claimed "All 4770 unit tests pass" and clean CI, but 3 checks were failing (Tests pytest, FVV Demo Integration, Two-Actor Demo Integration). The "4770" run used the default marker filter that excludes the demo/integration tests — precisely the ones that broke.

**FAIL — adapter changeset caused a 6-test regression.** By bisecting (reverting files to `origin/main` and re-running), I confirmed the branch contained two separable change-sets:
1. The **#1324 BT work** (`announce_tree.py`, `conditions.py`, `sync-ledger-replication.yaml`, `test_announce_tree.py`) — matches the issue's ACs and proposed tree structure exactly; passes all demos on its own.
2. An **adapter changeset** (`inbox_orchestration.py` rehydrate rewrite + `_inbox.py` in-place mutation + `CaseLedgerEntry` skip) added mid-branch to chase an FVV regression.

The adapter changeset was the **sole** cause of these failures:
```
test/demo/test_sync_ledger_replication.py::test_sync_single_peer_happy_path_replication
test/demo/test_pcr_engage_case.py::...test_engage_case_stores_owner_participant_in_reporter_dl
test/demo/test_two_actor_demo.py::TestRunTwoActorDemo::test_full_workflow_succeeds
test/demo/test_two_actor_demo.py::TestCaseLedgerInvariants::test_add_participant_status_entries_present
test/demo/test_two_actor_demo.py::TestCaseLedgerInvariants::test_all_participants_rm_closed_at_scenario_end
test/demo/test_two_actor_demo.py::TestCaseLedgerInvariants::test_required_event_types_present_in_case_actor_log
```
Worse, the adapter fix **did not achieve its own goal**: the FVV Demo Integration job failed with the very same `payloadSnapshot.target must be an inline object, not a bare ID string` error it was written to eliminate.

### Architecture smells in `inbox_orchestration.py` (the specific ask)

The `rehydrate()` rewrite exhibited real smells, which is why it was reverted rather than patched:

1. **Reimplements a canonical pipeline by hand.** `main`'s `rehydrate()` delegated to the DataLayer `read()` pipeline (`record_to_object → _rehydrate_fields → _coerce_to_semantic_class`). The rewrite bypassed `read()` and hand-rebuilt a *subset* — missing `target`/`origin` scalar-ref expansion and semantic coercion — which is exactly why `target` arrived as a bare ID string.
2. **Load-bearing hidden mutation.** `_store_nested_inbox_object` (a "store" helper) silently mutated the caller's `activity.object_`, and `rehydrate()` was rewritten to *depend* on that having happened — a hidden temporal contract coupling two adapter functions (confirmed: neither file could be reverted independently).
3. **Domain special-casing in a generic helper.** `type_ == "CaseLedgerEntry"` branching leaked SYNC-12 idempotency knowledge into a generic wire/adapter storage utility.
4. **Knowingly-broken replay path** shipped without a tracked issue (per the branch's own learning doc).

### Resolution applied

- **Reverted** the adapter changeset (source, its tests, and adapter-specific learning docs) in commit `41840381`, keeping the clean #1324 BT work. Full `uv run pytest -m ""` suite (incl. demos) now passes locally with **0 failures**.
- **Corrected the PR body** Verification section.
- **Opened #1472** to track the rehydration/semantic-routing bug with full root-cause analysis and ACs that direct the fix into the canonical rehydration pipeline (not an adapter bypass), wired under epic #1394.

### Other phases

- **Issue linkage** — BT portion satisfies #1324 AC-1..AC-5 precisely. Follow-ups #1446 and #1472 referenced.
- **Spec conformance** — SYNC-12 well-formed; conditions.py docstrings correctly updated.
- **Lint** — Black/flake8/mypy/pyright green.

**Recommendation:** merge once CI reruns green on the reverted HEAD.

sei-ahouseholder:
## Root cause found and fixed — SYNC-13 (core-only ledger writes)

The repeated CI failures were not flaky or cosmetic. They were the visible symptom of a real underspecification in #1324, now resolved.

**The conflation:** the SYNC-12 already-stored gate (`CheckLedgerEntryAlreadyStored`) treats *"a `CaseLedgerEntry` is present in the DataLayer"* as proof that *"its domain effects were already applied → skip `ProcessAndStore`"*. But the FastAPI ingress adapter used that **same** DataLayer presence for an unrelated purpose: `_store_nested_inbox_object` pre-stored the inline entry during parse so that `rehydrate()` (a by-ID re-read) could re-expand `object_` for semantic routing.

So the two layers overloaded one signal with contradictory meanings:

- Pre-store present → routing works, but the gate fires SUCCESS on first delivery → effects skipped → FVV `participant count 4; found 3`.
- Pre-store removed → gate works, but bare-string `object_` mis-routes to `announce_vulnerability_case` → the log-entry BT never runs.

Every prior attempt (#1446, #1472, and the reverted commits here) oscillated between those two corners. There is no green in that coordinate system.

**Resolution — ledger writes are a core responsibility; ingress never writes the ledger** (new spec group **SYNC-13**):

1. `_store_nested_inbox_object` refuses to persist a `CaseLedgerEntry` (SYNC-13-002).
2. `rehydrate` hydrates the *in-memory* parsed `Announce(CaseLedgerEntry)` instead of re-reading by ID, so the typed inline entry survives routing without a pre-store (SYNC-13-003/004). Non-ledger activities keep the proven by-ID path.
3. Inline nested-object subtype fields are preserved end-to-end (`serialize_as_any=True` in `Record.from_obj` + both emitters; inline `CaseLedgerEntry` kept inline in `_dehydrate_data` per SYNC-02-004; `Record.to_obj` re-types inline refs on read → replay reconstructs the full entry, #1472 AC-4).
4. The wire parser no longer recurses into opaque `payload_snapshot` blobs (that coercion down-cast the entry to base `as_Object`).

**Verification:** Docker isn't available in this environment, so the FVV/Two-Actor Docker jobs can't run locally — but the in-process isolated-app path (`test/demo/test_sync_ledger_replication.py`, `TestDeliveryIsolation`) drives the real `Announce(CaseLedgerEntry)` cross-app flow and reproduced the bug. The **full pytest suite now passes (exit 0)**; black/flake8/mypy/pyright clean. New `test_ingress_ledger_entry_routing.py` pins first-delivery routing and the replay path with no pre-store.

Note: `test_sync_ledger_replication` was updated to seed the case (with genesis hash) on the peer — a participant must hold the case replica before it can validate a replicated entry (CLP-08-005). That precondition was previously masked by the spec-violating pre-store.

PR #1815, opened 2026-07-29
Skill process improvements: reflect-cycle macro, review-priorities consolidation, board-id resolver

sei-ahouseholder:
## Summary

Three related improvements to the process-improvement skill suite, plus a fix
for a recurring class of bug.

### 1. `board-id.sh` — resolve board IDs at runtime, never hardcode them

GitHub board IDs (project/field/Schedule-option/issue-type node IDs) are
server-generated and **rotate** whenever a ProjectV2 single-select field's
options are edited. Copies pasted into skills drift stale and silently
mis-schedule issues — this has bitten the project before (the stale Schedule
option IDs fixed in `b8d609ab`/`cfe2a14a` were the same root cause).

New `.agents/skills/shared/board-id.sh` resolves any board ID **by name**,
backed by a TTL-cached `board-ids.json` (24h default; `--refresh` forces a
re-fetch; `BOARD_ID_TTL` overrides). The only hardcoded facts are the stable
bootstrap inputs (`owner=CERTCC`, `repo=Vultron`, project `#24`); everything
opaque is derived live. Every consumer — the shared scripts, `create_epic.sh`,
and all skill docs — now resolves IDs through it. `shared/README.md` documents
it as the single source of truth. Result: **zero raw board IDs** anywhere
except the resolver's bootstrap and the regenerable cache.

### 2. `review-priorities` — delegate instead of duplicate

`review-priorities` had re-implemented the tier-move mutation and constants
inline, duplicating `check-priority-status` (audit) and `update-priorities`
(mutations). Rewritten as a thin coordinator — the `pr-ship` pattern for the
board: **audit → review → update**, delegating each phase to its owning skill
(`check-priority-status`, `update-priorities`, `calve-epics`) and never
inlining a mutation. Adds a "when to use which board skill" table to resolve
the "which of these do I run?" confusion.

### 3. `reflect-cycle` — the reflection-half macro

New guided router over the reflection half of the dev cycle:
`learn → update-plan → decision-audit → review-priorities`. Unlike `pr-ship`,
it **stops between phases** because each contains a human gate (a `grill-me`
interview and/or its own PR) — it dispatches one phase, then re-confirms before
the next. `dev-status` now recommends it via a **macro-vs-atomic rule** (prefer
`reflect-cycle` over `learn`, and `pr-ship` over individual `pr-*` phases, when
the macro covers the same input).

## Verification

- All migrated scripts run correctly against the live board; bad inputs fail
  cleanly.
- `board-id.sh` verified: fresh-cache offline hit (no `gh` call), TTL expiry
  re-fetch, all categories return correct live values.
- `grep` for every raw board-ID pattern across `.agents/skills` → clean
  (resolver bootstrap + cache only).
- `markdownlint-cli2` on all 18 changed/new markdown files → 0 issues.
- Pre-commit hooks (markdownlint, flake8) pass on every commit.

## Commits

1. `chore(skills)`: board-id.sh resolver + migrate all consumers
2. `refactor(review-priorities)`: delegate to granular skills
3. `feat(reflect-cycle)`: reflection-half macro + dev-status routing

### DIFF
--- a/.agents/skills/bugfix/SKILL.md
+++ b/.agents/skills/bugfix/SKILL.md
@@ -29,10 +29,11 @@ bug is being fixed and why.
    (freeform), synthesize a title, and create the issue:
 
    ```bash
+   BUG_TYPE_ID=$(bash .agents/skills/shared/board-id.sh issue-type Bug)
    ISSUE_NUMBER=$(.agents/skills/manage-github-issue/manage_github_issue.sh \
      --title "${BUG_TITLE}" \
      --body "${BUG_BODY}" \
-     --issue-type-id "IT_kwDOAjf0s84AcFLq")
+     --issue-type-id "${BUG_TYPE_ID}")
    ```
 
 4. Fetch the issue body and comments. Use the content as bug context
@@ -83,6 +84,7 @@ Ask one targeted question before locking in scope:
   file each remaining concern as a new Bug-type issue:
 
   ```bash
+  BUG_TYPE_ID=$(bash .agents/skills/shared/board-id.sh issue-type Bug)
   .agents/skills/manage-github-issue/manage_github_issue.sh \
     --title "<short bug title>" \
     --body "## Symptoms
@@ -93,7 +95,7 @@ Ask one targeted question before locking in scope:
 
   ## Components involved
   - \`path/to/module.py\`" \
-    --issue-type-id "IT_kwDOAjf0s84AcFLq"
+    --issue-type-id "${BUG_TYPE_ID}"
   ```
 
   Confirm the narrowed scope before proceeding.
--- a/.agents/skills/check-priority-status/REFERENCE.md
+++ b/.agents/skills/check-priority-status/REFERENCE.md
@@ -9,15 +9,15 @@ status check.
 
 ## Project Board Constants
 
-| Name | Value |
+| Name | Resolver |
 |---|---|
-| Project node ID | `PVT_kwDOAjf0s84BZnre` |
-| Schedule field ID | `PVTSSF_lADOAjf0s84BZnrezhUlFOM` |
-| Schedule: Focus | `6bca50d7` |
-| Schedule: Now | `22e6679d` |
-| Schedule: Next | `1c1ed63d` |
-| Schedule: Later | `520032ef` |
-| Schedule: Someday | `a890eacc` |
+| Project node ID | `bash .agents/skills/shared/board-id.sh project` |
+| Schedule field ID | `bash .agents/skills/shared/board-id.sh schedule-field` |
+| Schedule: Focus | `bash .agents/skills/shared/board-id.sh schedule Focus` |
+| Schedule: Now | `bash .agents/skills/shared/board-id.sh schedule Now` |
+| Schedule: Next | `bash .agents/skills/shared/board-id.sh schedule Next` |
+| Schedule: Later | `bash .agents/skills/shared/board-id.sh schedule Later` |
+| Schedule: Someday | `bash .agents/skills/shared/board-id.sh schedule Someday` |
 
 ## Data Model
 
@@ -153,9 +153,12 @@ unblocked open sub-issue with no open PR.
 
 ### All items in Project #24 with Schedule tier
 
+Resolve `$PROJECT_ID` first via `bash .agents/skills/shared/board-id.sh project`,
+then substitute it into the query:
+
 ```graphql
 {
-  node(id: "PVT_kwDOAjf0s84BZnre") {
+  node(id: "$PROJECT_ID") {
     ... on ProjectV2 {
       items(first: 100) {
         nodes {
--- a/.agents/skills/check-priority-status/SKILL.md
+++ b/.agents/skills/check-priority-status/SKILL.md
@@ -58,8 +58,9 @@ Run the check-priority-status skill. The skill will:
 ### Querying Project #24 by Schedule Tier
 
 ```bash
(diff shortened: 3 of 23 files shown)


sei-ahouseholder (commented):
## 🔍 PR Triage — 4 findings (1 FAIL, 3 IMPROVE)

All 11 inspection phases ran. CI is green (Analyze, CodeQL, lint all pass). This is a skills/docs-only PR; the risk surface is the shell scripts. No spec/notes/ADR conformance concerns (nothing under `vultron/` or `specs/` touched). Grep confirms **zero raw board IDs** remain outside the resolver bootstrap + cache. Live smoke tests of `board-id.sh` (project, schedule Now, issue-type Epic, bad inputs) pass.

### 🔴 FAIL

**1. `board-id.sh:125` — self-heal retry is unreachable dead code**
`resolve()` uses `jq -er '... // empty'`, which exits non-zero (4) on a cache miss. So `if ! val=$(resolve "$@")` **always** takes the `exit 1` branch, and the refresh-once-and-retry block (lines 128–135) never runs. A name absent from a *still-fresh* cache hard-fails silently — no refresh, no message. This defeats the resolver's stated core purpose: recovering from board drift within the TTL window. When a new Schedule option or issue type is added to the board, `board-id.sh schedule <NewName>` will exit 1 instead of auto-refreshing. *Verified empirically.*
Fix: detect a miss as empty-output-with-success (drop `-e`, keep `// empty`), or split the helper so 'unknown category' (exit 1) is distinguished from 'valid category, name not found' (→ trigger refresh-retry).

### 🟡 IMPROVE

**2. `board-id.sh:119` — offline claim vs. TTL hard-fail**
README says the committed cache makes "first-run/offline use work," but once the cache is >24h stale, `refresh_cache()` runs, and offline it `return 2` → under `set -e` the script exits 2 rather than falling back to the present-but-stale cache. Offline use only works within 24h of the committed `fetched_at_epoch`. Fix: on refresh failure with a cache present, warn and fall back to stale.

**3. `board-id.sh:19` — bare `--refresh` contradicts its own doc**
Header says `--refresh [<args...>]` with "args optional," but a bare `--refresh` passes an empty category to `resolve()` → "unknown category ''" → exit 1. Fix: treat empty args after `--refresh` as a refresh-only no-op, or correct the doc.

**4. `reflect-cycle/SKILL.md:26` — dangling `plan-issue` gate reference**
The "why this stops between phases" note cites `plan-issue` as a gated phase, but the sequence is only learn → update-plan → decision-audit → review-priorities; `plan-issue` is never dispatched. Fix: drop the bullet or clarify it is out-of-macro.

---
*Artifact: `.claude/pr-1815-triage.json`. Next: `/pr-execute 1815` applies the fixes.*

sei-ahouseholder (commented):
## ⚙️ PR Execute — 4/4 findings fixed (commit `ff094956`)

All findings from triage were `fix-now`; each is resolved and verified empirically. No new issues filed, no deferred-ask items. Skills/docs-only PR — no pytest suite applies; shell behavior verified with targeted smoke tests.

### 🔴 FAIL — fixed

**1. `board-id.sh` self-heal was dead code** → dropped `-e` from `resolve()`'s `jq` calls. A valid-category-but-absent name now yields empty output + exit 0 (instead of `jq -e`'s exit 4), so the `if [ -z "$val" ]` refresh-retry block is finally reachable. *Verified:* fresh cache + absent name now refreshes and reports "no match even after refresh" instead of silently exiting 1.

### 🟡 IMPROVE — fixed

**2. Offline hard-fail past TTL** → added `ensure_cache()`: on refresh failure with a cache present, warn and fall back to the stale cache (exit 0) rather than propagating `return 2` under `set -e`. All refresh call sites now route through it. *Verified:* forced GraphQL failure with a stale cache resolves from it and exits 0 — the README "offline use works" claim now holds.

**3. Bare `--refresh` errored** → added a refresh-only branch: `board-id.sh --refresh` with no category now rebuilds the cache and exits 0, matching the "args optional" doc. *Verified:* exits 0; unknown category still exits 1.

**4. Dangling `plan-issue` gate** → removed the `plan-issue` bullet from `reflect-cycle/SKILL.md`; it is never dispatched by the sequence. markdownlint clean.

---
*Artifact: `.claude/pr-1815-execute.json`. Next: `/pr-verify 1815`.*

sei-ahouseholder (commented):
## ✅ PR Verify — all findings CONFIRMED · verdict: PENDING-CI

Integrity check passed (4 triage findings = 4 execute results). Each fix spot-checked against HEAD (not just the commit diff) at `ff094956`.

| # | Severity | Finding | Verdict |
|---|---|---|---|
| 1 | 🔴 FAIL | board-id.sh self-heal dead code | **CONFIRMED** — `resolve()` now `jq -r` (no `-e`) at lines 112/116; refresh-retry reachable |
| 2 | 🟡 IMPROVE | offline hard-fail past TTL | **CONFIRMED** — `ensure_cache()` warns + falls back to stale cache |
| 3 | 🟡 IMPROVE | bare `--refresh` errored | **CONFIRMED** — refresh-only branch exits 0 |
| 4 | 🟡 IMPROVE | dangling `plan-issue` gate | **CONFIRMED** — bullet removed from reflect-cycle/SKILL.md |

**CI:** `lint` ✅ passed; `Analyze (actions)`, `Analyze (python)`, `CodeQL` ⏳ pending.

**Overall: PENDING-CI** — all fixes verified; CI must finish green before merge. Re-run `/pr-verify 1815` (or `/pr-ship 1815`) after CI completes for the final `READY-TO-MERGE` confirmation.

Issue #1937, opened 2026-08-03
Test fidelity & code hygiene


sei-ahouseholder:
## Design idea

Raise reference-implementation test fidelity and clear code-hygiene debt so
tests are self-contained, spec-traceable, and free of stale markers:

- **Spec traceability** — audit/decorate `@pytest.mark.spec` coverage and
  ratchet it (#1806); per-event-type phrase regression tests (#1905).
- **Test self-containment** — dual-DataLayer setup for case-proposal tests
  (#1749); make `test_queues_offer_and_writes_activity_id` self-contained
  (#1904).
- **Code hygiene** — clear 15 TODO/FIXME/XXX markers (#505); migrate inline
  `em_state` accesses to Read/WriteEmStateNode (#1554); clear BTND-04-003
  import-ratchet KNOWN_VIOLATIONS (#1906).
- **Docs hygiene** — split notes/bt-pitfalls.md (#1899); amend ADR-0026
  Consequences with the stored-Offer trust invariant (#1760).

## Rationale

Calved 2026-08-03 from **#1794 Reference-Implementation Fidelity &
Correctness**. Janitorial fidelity/hygiene debt with no CM-23 dependency;
Someday tier — pull individual items forward opportunistically.

sei-ahouseholder:
Dissolved as part of epic restructuring. Children redistributed:
- #2321, #1904, #1848, #1899, #1760, #2150 → #2618 Test & code hygiene: quick wins (Now)
- #2099, #2414, #1993, #1877, #1978, #1906 → #2619 Test correctness: investigations & deep work (Later)
See #2617, #2618, #2619 for the successor epics.

PR #2119, opened 2026-08-08
fix(ci): repair demo-integration startup failure; add actionlint gate

### LINKED ISSUE #2118: Bug: demo-integration.yml startup failure — matrix context in job-level if: silently disabled all demo CI
## Summary

`.github/workflows/demo-integration.yml` has been failing at workflow **startup** since 2026-08-06 — 116 consecutive runs across every branch, including `main`. No demo or invariant-harness job has executed in that window.

## Symptom

Every run: `conclusion=failure`, `0s` duration, **zero jobs**, no logs, no annotations. The API reports `run.name` as the file path (`.github/workflows/demo-integration.yml`) instead of the workflow name (`Demo Integration`) — the signature of a workflow GitHub could not parse.

Failures appeared on **every branch**, not just `main`, even though the `push` trigger is filtered to `branches: ["main"]`. A startup failure is registered before the branch filter is evaluated, so the red X showed up on branches the workflow was never meant to run on. That made the signal read as CI noise rather than a real regression, which is why it went unnoticed for two days.

## Root cause

Both the `demo` and `invariant-harness` jobs had a job-level `if:` referencing the `matrix` context:

```yaml
if: >-
  !startsWith(github.head_ref, 'dependabot/') &&
  (github.event_name != 'pull_request' || matrix.full_suite_only != true)
```

GitHub evaluates job-level `if:` **before** matrix expansion, so the `matrix` context does not exist at that point. `actionlint` reports it directly:

```
demo-integration.yml:101:96: context "matrix" is not allowed here.
   available contexts are "github", "inputs", "needs", "vars". [expression]
demo-integration.yml:219:108: (same)
```

This is valid YAML — `yaml.safe_load()` accepts the file — so nothing in the existing test or lint suite could detect it.

Introduced in aa7b7929 (DEMOCI-06), which added the `full_suite_only` matrix field and these gating conditions. The same commit added the `push: branches: [main]` trigger, which is what spread the visible failure to every branch.

## Fix

Resolve the matrix **before** expansion instead of gating expanded entries:

- Scenario list moves to `.github/demo-scenarios.json` (single source of truth).
- A new `scenarios` job filters it by `github.event_name` and emits it as a JSON output.
- `demo` and `invariant-harness` consume `include: ${{ fromJSON(needs.scenarios.outputs.matrix) }}`.

This also removes the verbatim 9-entry matrix that was duplicated across both jobs, where the two copies could silently drift.

## Prevention

Add `actionlint` as a CI gate and pre-commit hook. It flags this exact error in one line. Notably it must **not** go in `python-app.yml`, whose path filter is scoped to `vultron/**` and `test/**` — a workflow-only change would never trigger it. A dedicated `actions-lint.yml` scoped to `.github/**` is required.

## Acceptance criteria

- [ ] AC-1: `demo-integration.yml` parses and schedules jobs; no `matrix` reference in any job-level `if:`.
- [ ] AC-2: `pull_request` runs the 4-scenario minimum PR set (DEMOCI-06-002); `push`/`workflow_dispatch` run all 9 (DEMOCI-06-003).
- [ ] AC-3: Dependabot PRs skip both jobs without spurious failures (DEMOCI-02-004, DEMOCI-04-005).
- [ ] AC-4: `actionlint` runs in CI on `.github/**` changes and as a pre-commit hook, pinned by version + SHA256.
- [ ] AC-5: The scenario JSON is structurally validated (field types, duplicate names, non-empty PR set, referenced test files exist).
- [ ] AC-6: Spec updated to prohibit the anti-pattern and record the actionlint gate.

## Notes

Nothing alerted on 116 failures over two days. A required status check on `main` would have surfaced this on day one — worth tracking separately.

sei-ahouseholder:
Fixes #2118

## What was broken

`demo-integration.yml` had been failing at workflow **startup** since 2026-08-06 — **116 consecutive runs** across every branch, including `main`. Zero demo or invariant-harness jobs executed in that window.

Both the `demo` and `invariant-harness` jobs carried a job-level `if:` referencing the `matrix` context:

```yaml
if: >-
  !startsWith(github.head_ref, 'dependabot/') &&
  (github.event_name != 'pull_request' || matrix.full_suite_only != true)
```

GitHub evaluates job-level `if:` **before** matrix expansion, so `matrix` does not exist there and the entire file is rejected. The result is a near-silent failure: 0s duration, zero jobs, no logs, no annotations, and `run.name` reported as the file path rather than `Demo Integration`.

The file is valid YAML — `yaml.safe_load()` accepts it — so nothing in the existing test or lint suite could catch it. Introduced in aa7b7929 (DEMOCI-06). That same commit added the `push: branches: [main]` trigger, which is what made the failure visible on *every* branch: a startup failure is registered before the branch filter is evaluated, so the red X appeared on branches the workflow never targeted. That made it read as CI noise rather than a regression, which is why it survived two days.

## The fix

Resolve the matrix **before** expansion rather than gating expanded entries:

- **`.github/demo-scenarios.json`** (new) — single source of truth for the 9 scenarios and their `full_suite_only` flags.
- **`scenarios` job** (new) — filters that list by `github.event_name`, emits it as a JSON output.
- **`demo` / `invariant-harness`** — consume `include: ${{ fromJSON(needs.scenarios.outputs.matrix) }}`.

This also eliminates the verbatim 9-entry matrix that was duplicated across both jobs, where the copies could silently drift.

Two subtleties worth a reviewer's eye:

- The Dependabot guard moved onto `scenarios` (a job with no matrix), so it's enforced **once** instead of three times. On push events `github.head_ref` is empty, so the guard is a no-op there and the on-main run is never skipped (DEMOCI-05-004).
- `invariant-harness` uses `always() && needs.scenarios.result == 'success'` rather than a bare `always()`. A bare `always()` runs the job even when an upstream job was **skipped** — which would resurrect exactly the spurious Dependabot failures that DEMOCI-04-005 exists to prevent.

## Prevention

`actionlint` added as both a CI gate (`.github/workflows/actions-lint.yml`) and a pre-commit hook, pinned by version and verified against a pinned SHA256.

It deliberately does **not** live in `python-app.yml`: that workflow's path filter is scoped to `vultron/**` and `test/**`, so a workflow-only change would never have triggered the check. The new workflow is scoped to `.github/**`.

It also structurally validates `demo-scenarios.json` — field types, duplicate scenario names, non-empty PR set, referenced test files exist — because the matrix is now *data*, which degrades quietly rather than loudly. A string `"false"` instead of a boolean would silently drop a scenario from the PR set; an empty PR set would report a vacuous green.

## Spec changes

- **DEMOCI-06-004** (new) — selection MUST resolve pre-expansion; `matrix.*` in a job-level `if:` is prohibited, with the failure mode documented.
- **DEMOCI-08** (new) — the actionlint gate and the JSON validation requirements.
- **DEMOCI-04-005** (amended) — it *mandated* the literal `!startsWith(github.head_ref, 'dependabot/')` predicate on the invariant-harness job. Hoisting the guard upstream contradicts its letter while preserving its intent, so I restated it as the required **outcome** and recorded why. Flagging for review since it's the one place I changed a MUST rather than adding one.
- **DEMOCI-06-003** — fixed a stale "8-scenario" count (it enumerates 9).
- **DEMOCI-02-003** — added the new JSON file to the enumerated path filter.

## Verification

- `actionlint` (with shellcheck) and `spec-lint` both exit 0; all five files parse.
- **The fix is verified against the actual bug, not just lint-clean**: restoring the original buggy file into a scratch workflow reproduces the error at both original line numbers, and the pre-commit hook catches it.
- Both matrices resolve correctly — 4 scenarios on `pull_request`, 9 on `push`/`dispatch`.
- All 9 referenced invariant test files exist.
- The `GITHUB_OUTPUT` write is single-line and `fromJSON`-parseable for both event types.
- The JSON validator rejects all four malformed-input cases (string-boolean, duplicate name, empty PR set, missing field) and accepts valid input.

**Caveats:**

- The demos themselves could not be run locally — no Docker in the devcontainer. The first on-main run after merge is the real proof. Note this PR's own `pull_request` run will exercise the new 4-scenario path, since the workflow's path filter includes itself.
- `actionlint` cannot lint the composite actions under `.github/actions/` — it parses any file handed to it as a workflow, so an `action.yml` yields spurious "on section is missing" errors. The path filters still include them so consuming workflows get re-linted.

## Follow-up worth tracking separately

Nothing alerted on 116 failures over two days. A required status check on `main` would have surfaced this on day one.

### DIFF
--- a/.github/demo-scenarios.json
+++ b/.github/demo-scenarios.json
@@ -1,0 +1,47 @@
+[
+  {
+    "demo": "fv",
+    "test_file": "test/ci/invariants/test_fv_invariants.py",
+    "full_suite_only": false
+  },
+  {
+    "demo": "fvcv-handoff",
+    "test_file": "test/ci/invariants/test_fvcv_handoff_invariants.py",
+    "full_suite_only": false
+  },
+  {
+    "demo": "fcvcv",
+    "test_file": "test/ci/invariants/test_fcvcv_invariants.py",
+    "full_suite_only": false
+  },
+  {
+    "demo": "fcv-reject",
+    "test_file": "test/ci/invariants/test_fcv_reject_invariants.py",
+    "full_suite_only": false
+  },
+  {
+    "demo": "fvv",
+    "test_file": "test/ci/invariants/test_fvv_invariants.py",
+    "full_suite_only": true
+  },
+  {
+    "demo": "fvcv-extension",
+    "test_file": "test/ci/invariants/test_fvcv_extension_invariants.py",
+    "full_suite_only": true
+  },
+  {
+    "demo": "fccv-extension",
+    "test_file": "test/ci/invariants/test_fccv_extension_invariants.py",
+    "full_suite_only": true
+  },
+  {
+    "demo": "fccv-handoff",
+    "test_file": "test/ci/invariants/test_fccv_handoff_invariants.py",
+    "full_suite_only": true
+  },
+  {
+    "demo": "fcv",
+    "test_file": "test/ci/invariants/test_fcv_invariants.py",
+    "full_suite_only": true
+  }
+]
--- a/.github/workflows/actions-lint.yml
+++ b/.github/workflows/actions-lint.yml
@@ -1,0 +1,113 @@
+# Workflow linter — validates the GitHub Actions workflow files themselves.
+#
+# Kept separate from python-app.yml because that workflow's path filter is
+# scoped to vultron/ and test/, so workflow-only changes would never trigger
+# it. This workflow's filter is scoped to .github/ for the same reason.
+#
+# actionlint catches errors that YAML parsing cannot: invalid context
+# references, malformed `if:` expressions, unknown action inputs, and shell
+# problems in `run:` blocks (via shellcheck).
+#
+# Motivating regression (#2118, DEMOCI-06-004): demo-integration.yml carried a job-level
+# `if:` referencing the `matrix` context. That is valid YAML — yaml.safe_load
+# accepts it — but the `matrix` context does not exist until after matrix
+# expansion, so GitHub rejected the file with a startup failure: zero jobs,
+# no logs, no annotation, and a bare red X. It went unnoticed for 116
+# consecutive runs across every branch. actionlint flags it in one line.
+
+name: Actions Lint
+
+on:
+  push:
+    branches: ["main"]
+    paths:
+      - ".github/workflows/**"
+      - ".github/actions/**"
+      - ".github/demo-scenarios.json"
+  pull_request:
+    branches: ["main"]
+    paths:
+      - ".github/workflows/**"
+      - ".github/actions/**"
+      - ".github/demo-scenarios.json"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  actionlint:
+    name: Lint (actionlint)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      # Pinned by version + SHA256 so a compromised or retagged release cannot
+      # silently change what runs in CI.
+      - name: Install actionlint
+        env:
+          ACTIONLINT_VERSION: "1.7.7"
+          ACTIONLINT_SHA256: "023070a287cd8cccd71515fedc843f1985bf96c436b7effaecce67290e7e0757"
+        run: |
+          set -euo pipefail
+          TARBALL="actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz"
+          curl -fsSL --retry 3 \
+            "https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VERSION}/${TARBALL}" \
+            -o "$TARBALL"
+          echo "${ACTIONLINT_SHA256}  ${TARBALL}" | sha256sum --check --strict
+          tar xzf "$TARBALL" actionlint
+          install -m 0755 actionlint /usr/local/bin/actionlint
+          actionlint -version
+
+      # Default discovery walks .github/workflows/. shellcheck is preinstalled
+      # on ubuntu-latest runners, so `run:` blocks are linted too.
+      #
+      # Composite actions under .github/actions/ are NOT linted: actionlint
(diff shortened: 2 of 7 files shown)


sei-ahouseholder:
## CI status — the red demo checks are pre-existing, not caused by this PR

The workflow **now starts correctly**, which was the point of this PR. [Run 31230964426](https://github.com/CERTCC/Vultron/actions/runs/31230964426) scheduled **5 jobs** (previously 0), `Select scenarios` succeeded, and it picked exactly the 4-scenario minimum PR set:

```
success   Select scenarios
success   fv Demo Integration          success   fv Invariant Harness
failure   fcvcv Demo Integration       failure   fcvcv Invariant Harness
failure   fvcv-handoff Demo Integration failure  fvcv-handoff Invariant Harness
success   fcv-reject Demo Integration  failure   fcv-reject Invariant Harness
```

The 3 failures are **genuine application regressions that landed unvalidated** while demo CI was dark for ~2 days. Evidence:

1. This PR touches **no** files under `vultron/`, `docker/`, `test/`, or `integration_tests/` — CI config and specs only.
2. Parsing both workflow versions confirms the `demo` and `invariant-harness` **step lists are byte-identical** (8 and 4 steps). The only structural change is the new upstream `scenarios` job.
3. `fcvcv` and `fvcv-handoff` both **passed** on 2026-08-06 13:28Z (run 31106053880), before the startup failure began.
4. `fcv-reject` was added in 37b1edf1 *during* the dark window, so it has never had a green CI run — it merged on a PR whose demo checks reported a 0s startup failure instead of a real result.

Filed as sub-issues of Epic #1937:

- **#2120** — `fcvcv` + `fvcv-handoff`: `ReconstructChainTail` cannot bootstrap an unanchored ledger chain (CLP-08-005). Includes candidate commits for bisection.
- **#2121** — `fcv-reject` invariant 15: `vfd_state == 'VFd'` never observed (34/35 invariants pass; likely a copy-pasted expected-transition set, since a vendor that rejects an invitation plausibly never produces a fix).

This is the fix working as intended: it converted two days of silent green-by-absence into three actionable failures. Worth deciding explicitly whether to merge with those demo checks red — the alternative is leaving the workflow broken, which is strictly worse.

`Actions Lint` is green after 3c2a27e2, which fixed two pre-existing shellcheck SC1007 warnings the new gate caught in `spec-check.yml` and `docs-build-check.yml` — its first real catches.

PR #2123, opened 2026-08-08
fix: point scenario ratchet at demo-scenarios.json

sei-ahouseholder:
- Closes #2122

## Summary

Re-points the CI scenario ratchet at `.github/demo-scenarios.json` after PRs
#2118/#2119 moved the scenario matrix there, leaving `test_valid_scenarios_matches_ci`
and `test_at_least_one_scenario_in_ci` broken on `main`.

## Changes

- **`test/demo/test_integration_script_scenarios.py`**: Replace regex-based YAML
  scraper of `DEMO=` lines with a JSON reader; `_ci_scenarios()` now parses
  `.github/demo-scenarios.json` and extracts the `"demo"` key from each entry.
  Add `_CI_SCENARIOS_JSON` path constant. Add `test_ci_scenarios_json_exists`
  and `test_ci_scenarios_json_is_valid` guards (DEMOCI-08-002).
  Update module docstring and assertion messages to name the new source file.

## Verification

- [x] `test_valid_scenarios_matches_ci` passes (was FAILED)
- [x] `test_at_least_one_scenario_in_ci` passes (was FAILED)
- [x] All 5 original tests + 2 new JSON guards pass (7 total)
- [x] All four linters clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1994, opened 2026-08-05
test_demo in test_initialize_participant_demo.py references closed #464 — needs new tracking issue


sei-ahouseholder:
## Summary

`test/demo/test_initialize_participant_demo.py::test_demo` is `xfail(run=False)` referencing #464 (Two-Actor Demo Redesign, now closed). The test is never run.

The `run=False` means the test is skipped entirely — not even attempted. This hides whether the underlying issue (pre-CBT case seeding pattern incompatible with CBT trust setup) was resolved by #464's work or is still broken.

## Required action

1. Remove `run=False` and run the test to determine current status.
2. If it passes: remove the `xfail` marker entirely.
3. If it still fails: update the `reason` to reference this issue number and remove `run=False` so failures are visible.

## Acceptance criteria

- `xfail(run=False)` is gone
- Test either passes cleanly or has a live `reason=#N` referencing this issue

sei-ahouseholder:
Claimed on branch `bug/1994-xfail-run-false-test-demo`.

sei-ahouseholder:
**History: implementation — fix: discover canonical case in initialize_participant_demo; remove xfail**

Issue #1994 tracked that test_demo in test_initialize_participant_demo.py had xfail(run=False) referencing closed issue #464. The run=False had been removed by a prior commit; this PR removed the xfail marker and fixed the underlying demo protocol bug. Root cause: setup_case_precondition was creating a vendor-local case that the protocol skipped (no ReportCaseLink). Fix: discover the CaseActor's canonical case via GET /datalayer/VulnerabilityCases/ after validate-report. Filed #2356 for run_exchange_demos missing assert_demo_success. PR: https://github.com/CERTCC/Vultron/pull/2357

Issue #1993, opened 2026-08-05
CM-22 derives-from-non-accepted-adr signal not firing in decision-audit inventory


sei-ahouseholder:
## Summary

`test/metadata/test_decision_audit_inventory.py::TestBuildInventory::test_known_landmine_surfaces` is `xfail` with no issue reference. The test verifies that CM-22 (which derives from superseded ADR-0015) fires a `derives-from-non-accepted-adr` signal in the decision-audit inventory. It currently fails — the signal is not firing.

This is the ISSUE-1272 class of defect: a spec entry that cites a superseded ADR should surface as a candidate in the audit inventory.

## Acceptance criteria

- `decision_audit_inventory.py` emits a `derives-from-non-accepted-adr` signal for CM-22
- `test_known_landmine_surfaces` passes without `xfail` marker

sei-ahouseholder:
**Triage audit — 2026-08-26**

Confirmed still live on `origin/main`.

`test/metadata/test_decision_audit_inventory.py:117` is still `@pytest.mark.xfail(reason="CM-22 derives-from-non-accepted-adr signal not yet firing. Tracked in #1993.")`

Root cause unchanged: CM-22's specs carry `adr: [ADR-0023, ADR-0041]` — no ADR-0015 in the structured field. `_group_signals()` checks `spec.adr` edges only, not prose mentions. The Aug 17 decision-audit cleanup (commit `552ee7af`) fixed CM-15 and LST-02 dangling ADR-0015 refs but explicitly left CM-22 alone.

Issue #1992, opened 2026-08-05
Remove core-layer actor classes from wire VOCABULARY registry (ARCH-12-003)


sei-ahouseholder:
## Summary

`test/architecture/test_hierarchy_invariants.py::TestWireVocabularyHierarchy::test_all_vocabulary_are_as_base_subclasses` is `xfail` referencing #800 (closed without fixing this violation).

The following core-layer classes are incorrectly registered in the wire `VOCABULARY` dict, violating ARCH-12-003:

- `CoreActor` (`'Actor'`), `VultronOfferRecord` (`'OfferRecord'`), `PendingCaseInbox`, `PendingCreateCaseActivity`, `ReplicationState`, `ReportCaseLink`

`VOCABULARY` must contain only wire-layer `as_Base` subclasses.

## Acceptance criteria

- All listed classes removed from `VOCABULARY`
- `test_all_vocabulary_are_as_base_subclasses` passes without `xfail` marker
- No regression in wire-layer deserialization tests

sei-ahouseholder:
Claimed on branch `bug/1992-vocab-registry-arch-violation`.

sei-ahouseholder:
**History: implementation — Remove core-layer types from wire VOCABULARY registry (ARCH-12-003)**


Fixes ARCH-12-003 violation (#1992): 7 core-layer types were incorrectly
registered in the wire `VOCABULARY` dict —
`CoreActor` (keys "Actor" and "CoreActor"), `VultronOfferRecord` ("OfferRecord"),
`VultronPendingCaseInbox` ("PendingCaseInbox"), `PendingCreateCaseActivity`
("PendingCreateCaseActivity"), `VultronReportCaseLink` ("ReportCaseLink"),
`VultronReplicationState` ("ReplicationState").

Root cause: `Record.to_obj()` (and similar DB deserialization paths) called
`find_in_vocabulary(type_string)` which required these types to appear in the
wire vocabulary.

Fix: Added `CORE_TYPE_MAP` in `vultron.core.models.registry` with dual-key
auto-registration in `VultronObject.__init_subclass__` (by class name and by
Literal `type_` value). Modified `find_in_vocabulary` to fall back to
`CORE_TYPE_MAP`. Removed all direct `VOCABULARY` registrations of core-layer
types. `VOCABULARY["Actor"]` now correctly points to `as_Actor` again.

Key discovery during implementation: `model_fields` is NOT populated at
[...]

PR #2126, opened 2026-08-08
fix(test): stop demo config cache leaking across tests

### LINKED ISSUE #1897: Concern: test_queues_offer_and_writes_activity_id is order-dependent on ActorConfig env
## Summary

`test/core/behaviors/case/nodes/test_communication.py::TestSendOfferCaseManagerRoleNode::test_queues_offer_and_writes_activity_id` fails when run in isolation or a small subset with:

```
ResolveCaseActorUrlsNode: case_actor_service_url is not configured in ActorConfig (set VULTRON_ACTOR__CASE_ACTOR_SERVICE_URL)
AssertionError: CreateCaseActorNode must write case_actor_id
```

The test passes in a full-suite run. First reproduced on a clean stashed base (unrelated to any current change), so it is a pre-existing condition.

## Why It Matters

The test depends on `VULTRON_ACTOR__CASE_ACTOR_SERVICE_URL` being set by some other test's fixture rather than establishing it itself. This makes subset runs — the normal workflow when iterating on a change in this file — report a spurious failure, and forces developers to rule out their own diff before concluding the failure is pre-existing.

## Suggested Fix

The test should set `VULTRON_ACTOR__CASE_ACTOR_SERVICE_URL` (and any other required `ActorConfig` fields) in its own fixture or `monkeypatch` call rather than relying on cross-test environment state.

## File

- `test/core/behaviors/case/nodes/test_communication.py`
- `TestSendOfferCaseManagerRoleNode::test_queues_offer_and_writes_activity_id`

### LINKED ISSUE #2086: Flaky CI: TestBootstrapSequence.test_announce_creates_case_replica / test_case_fields_preserved_in_replica — SvcValidateReportUseCase no routable recipients
## Summary

Two tests in `test/demo/test_pcr_bootstrap.py::TestBootstrapSequence` fail intermittently in CI with:

```
AssertionError: validate-report trigger failed (422): SvcValidateReportUseCase failed: no routable recipients for report activity
```

## Affected Tests

- `test_announce_creates_case_replica` (line 360)
- `test_case_fields_preserved_in_replica` (line 400)

Both call `_bootstrap_case_for_participant()` (line 317) which triggers `validate-report`.

## Evidence

- **CI failure**: `gh run view 31211166984` — both tests FAILED
- **Locally**: `uv run pytest -m "" test/demo/test_pcr_bootstrap.py::TestBootstrapSequence::test_announce_creates_case_replica test/demo/test_pcr_bootstrap.py::TestBootstrapSequence::test_case_fields_preserved_in_replica` — 2 passed
- Neither test is touched by the branch that surfaced this (PR #2083)
- Not in `notes/flaky-tests.md` catalog prior to this issue

## Root Cause Hypothesis

`EmitValidateReportActivity` logs: _no routable recipients for report activity_. The `ValidateReportBT` cannot find a reporter-actor recipient — likely a timing or ordering issue in the bootstrap sequence where the participant record isn't yet visible when the validate-report BT runs. Exact race window not yet identified.

## Blocked PRs

- PR #2083
- PR #2113
- PR #2125

## Local reproduction (2026-08-08, via PR #2113 triage)

Confirmed **not flaky in the load-dependent sense — deterministically
order-dependent**, and confirmed pre-existing on clean `origin/main`:

| Invocation | Result |
|---|---|
| `pytest -m "" test/demo/test_pcr_bootstrap.py::TestBootstrapSequence` (isolated) | 3 passed |
| `pytest -m "" test/demo/` (branch `bug/1911-...`) | 2 failed, 1038 passed |
| `pytest -m "" test/demo/` (**clean `origin/main` worktree**, e38bb967) | 2 failed, 1037 passed |
| `pytest -m "" --tb=short` (full suite, branch) | 2 failed, 7495 passed |

Reproduced with `-p no:randomly`, so this is **not** random test-order
shuffling — running the `test/demo/` directory together is sufficient and
sufficient-consistently to trigger it. The baseline run on a detached
`origin/main` worktree with none of PR #2113's changes present produces the
identical two failures, which rules out any contribution from that branch.

**Narrowing hint**: the trigger is intra-`test/demo/` cross-test state leakage
(shared DataLayer / registry / emitter global), not CI scheduling latency as
the original hypothesis suggested. A bisect over `test/demo/` test files run
before `test_pcr_bootstrap.py` should identify the polluting module.

### LINKED ISSUE #2122: bug: test_integration_script_scenarios broken on main — DEMO= source moved to .github/demo-scenarios.json
## Summary

Two tests fail on clean `origin/main` (2b6851b5):

- `test/demo/test_integration_script_scenarios.py::TestIntegrationScriptScenarios::test_valid_scenarios_matches_ci`
- `test/demo/test_integration_script_scenarios.py::TestIntegrationScriptScenarios::test_at_least_one_scenario_in_ci`

```
AssertionError: No DEMO= lines found in demo-integration.yml
assert 0 >= 1
```

## Root cause — semantic merge collision between two PRs

Neither PR is wrong in isolation; they are incompatible when combined.

- **#2114** (`bug/1914-pre-pr-integration-blind-spot`) added
  `test_integration_script_scenarios.py`, which asserts that `VALID_SCENARIOS`
  in `integration_tests/demo/run_multi_actor_integration_test.sh` matches the
  `DEMO=` values scraped out of `.github/workflows/demo-integration.yml`
  (DEMOCI-02-003).
- **#2118 / #2119** (`fix/demo-integration-matrix-startup-failure`) refactored
  the workflow so the scenario matrix now lives in the new
  `.github/demo-scenarios.json`. There are no longer any `DEMO=` lines in
  `demo-integration.yml`.

Result: the CI-side parser returns an empty set, so the comparison fails and the
`>= 1` guard trips.

```
In script but not CI: ['fccv-extension', 'fccv-handoff', 'fcv', 'fcv-reject',
                       'fcvcv', 'fv', 'fvcv-extension', 'fvcv-handoff', 'fvv']
In CI but not script: []
```

## Verification that this is not branch-specific

Reproduced on a detached `origin/main` worktree at 2b6851b5 with no other
changes present:

```
2 failed, 3 passed in 2.42s
```

Surfaced during `/pr-ship` on PR #2113 after merging current `origin/main`.
Unrelated to that PR's contents.

## Suggested fix

Point the test's CI-side extraction at `.github/demo-scenarios.json` (read the
`demo` key of each entry) instead of grepping `DEMO=` from the workflow YAML.
The `run_multi_actor_integration_test.sh` side needs no change — its
`VALID_SCENARIOS` list already holds all 9 scenarios and matches the JSON.

Worth confirming whether DEMOCI-02-003's wording needs updating to name
`.github/demo-scenarios.json` as the authoritative scenario source.

## Impact

`main` is red. Every PR branched from or merged with current `main` inherits
these two failures, which will mask genuine regressions during triage.

sei-ahouseholder:
- Closes #2086
- Closes #1897

## Summary

Four demo fixtures called `reload_config()` in teardown *before* `monkeypatch`
undid their env patches, pinning a fake CaseActor host into the module-level
config cache for the remainder of the pytest session. This fixes the teardown
order at the source, adds an autouse guard so a future ordering mistake cannot
outlive its test, and adds a regression module.

## Root cause

`vultron/config/app.py` keeps a process-global `_config_cache`;
`reload_config()` clears it and re-reads the environment. Pytest's
`monkeypatch` fixture undoes env changes **after** the requesting fixture's
teardown body runs, so `reload_config()` in teardown re-caches the *still
patched* value.

Every later demo test then derived its CaseActor ID from that leaked host
(e.g. `http://coordinator-otc.test`). No `_TestClientRouter` registers it, so
`emit()` silently dropped the `Create(CaseProposal)` (DEBUG log only), the
CaseActor never created the canonical case with its `CASE_MANAGER` participant,
and `trigger/validate-report` failed three steps downstream with
`SvcValidateReportUseCase: no routable recipients`.

No production code was at fault.

## The failure is deterministic, not nondeterministic

An earlier triage note on this issue concluded the failures were "genuinely
nondeterministic" because `pytest -m "" test/demo/` gave `[2, 0, 2]` bootstrap
failures over three identical runs. **That was a measurement artifact.**

`timeout = 5` with `timeout_method = "thread"` (`pyproject.toml:123`) kills the
**whole pytest process**, not just the slow test. A run that trips it emits no
summary line and never reaches `test_pcr_bootstrap.py` — scoring a phantom `0`.
Re-running the identical command with a driver that also records the
`+++ Timeout +++` marker reproduces `[2, 0, 2]` and shows the middle run
aborted mid-session with `timeout=True` and no summary.

At any granularity that actually runs both modules to completion, clean
`origin/main` (post-#2123, `1a911d30`) fails deterministically:

| Invocation (`-p no:randomly`, 3× each) | clean `origin/main` | this branch |
|---|---|---|
| `test_fvcv_handoff_demo.py` + `test_pcr_bootstrap.py` | `[2, 2, 2]` | `[0, 0, 0]` |
| `...::TestOwnershipTransferAnnounceReachesFinderAC5c` + bootstrap | `[2, 2, 2]` | — |
| `test_pcr_bootstrap.py` alone | `[0, 0, 0]` | `[0, 0, 0]` |
| full `test/demo/` (4×) | `[2, 0, 2]` | `[0, 0, 0, 0]` |

The first row contradicts the earlier pairwise sweep that reported zero failures
across all 27 preceding files; that sweep should be treated as unreliable (it
also suffered shell word-splitting no-ops). Confirmed still failing `[2,2,2]` on
post-#2123 `main`, so #2123 fixed #2122 — not this.

## Changes

- **`test/demo/test_fvcv_handoff_demo.py`**, **`test/demo/test_pcr_late_joiner.py`**,
  **`test/demo/test_case_proposal_round_trip.py`**, **`test/demo/test_pcr_engage_case.py`**:
  call `monkeypatch.undo()` before `reload_config()` in fixture teardown.
  (`test_fv_demo.py` already avoided this by using an explicit `MonkeyPatch()`.)
- **Six non-demo fixtures** — `test/core/behaviors/case/nodes/test_case_setup.py`,
  `test/core/behaviors/case/test_bug_26040902_regression.py`,
  `test/core/behaviors/case/test_receive_report_case_tree.py`,
  `test/core/use_cases/received/test_ack_validate_report_received.py`,
  `test/core/use_cases/received/test_submit_report_received.py`,
  `test/core/use_cases/test_validate_report_ledger_routing.py`: same teardown
  ordering fix. All six patch `VULTRON_ACTOR__CASE_ACTOR_SERVICE_URL` and
  reloaded while still patched, leaking `http://case-actor:7999/api/v2` for the
  rest of the session. Nothing outside `test/demo/` is covered by the guard.
- **`test/demo/conftest.py`**: add `config_snapshot()` /
  `restore_config_if_leaked()` helpers and an autouse
  `restore_case_actor_url_after_each_test` fixture that snapshots the **full**
  config around every demo test and reloads if it drifted — so drift in a
  setting other than the two #2086 URLs is caught too. The repair verifies the
  post-reload state and raises rather than reporting a success it did not
  achieve (ARCH-15-001). Each repair is recorded on a `ConfigLeakLedger`.
- **`test/demo/test_config_leak_guard.py`** (new): 8 tests covering both
  teardown orderings, the guard helpers, non-URL drift, the
  unrepairable-pollution path, the demo session's baseline CaseActor URL, and
  `TestNoFixtureLeakedConfig` — a ratchet asserting the guard never had to
  repair anything, so the source fixes stay enforced.
- **`test/core/behaviors/case/test_create_tree.py`**,
  **`test/core/behaviors/case/nodes/test_communication.py`**: configure
  `case_actor_service_url` in their own autouse fixture (#1897). Both run
  `ResolveCaseActorUrlsNode` — which FAILs when that URL is None
  (CP-08-002/003) — without setting it, and passed only by borrowing the leak
  the change above removes.
- **`test/AGENTS.md`**: document the teardown-ordering rule, the guard's
  function-scope-only limitation, and the converse rule that a module needing a
  `VULTRON_*` setting must set it itself.
- **`notes/flaky-tests.md`**: correct the #2086 entry, drop the #2086 and #2122
  rows (the latter fixed by #2123), and document the timeout-vs-pass trap.
- **`plan/incoming/learnings/`**: three learnings — the silent
  `_TestClientRouter` drop that masked the cause, the module-level config cache
  footgun, and the thread-timeout measurement trap.

## The first revision's regression coverage was vacuous

Worth recording, because the guard was the problem. With the autouse guard in
place and **all four source fixes reverted**, `test/demo/` reported
`25 passed` — nothing failed. The guard repaired every leak so reliably that no
test could observe a regression of the fixes the PR exists to make. A
three-way experiment isolated it:

| guard | source fixes | result |
|---|---|---|
| off | present | 20 passed (control) |
| on | **reverted** | **25 passed — nothing fails** |
| off | reverted | both `TestBootstrapSequence` tests FAIL, "no routable recipients" |

`ConfigLeakLedger` + `TestNoFixtureLeakedConfig` resolve this: the guard still
repairs (so one bad teardown cannot cascade), but every repair is recorded and
asserted on. Validated by reverting two *different* fixtures in turn — each
produced a failing ratchet naming the offending host.

## Verification

Rebased onto `origin/main` @ `1a911d30` (includes #2123).

- Unit suite: `6453 passed, 362 skipped, 2 xfailed, 5552 subtests passed`.
- Full suite incl. integration and demo (`-m ""`):
  `7511 passed, 362 skipped, 3 xfailed, 1 xpassed, 5552 subtests passed` —
  **0 failures**.
- xfail ratchet: all of #1991, #1992, #1993, #1994 confirmed open.
- Black, flake8, mypy, pyright, markdownlint all clean.

### #1897 was surfaced, not caused, by this PR

Removing the leak broke 4 tests that had been depending on it
(`test_create_tree.py` ×3, `test_communication.py` ×1). Proven by matched runs
at identical fixed order (`-p no:randomly`): clean branch HEAD passes the full
suite; with the six-fixture fix applied, exactly those 4 fail. They also fail on
clean `origin/main` when run as a subset — which is #1897, filed earlier and
still open. Fixed properly here (each module configures the URL itself) rather
than by restoring the leak.

## Notes

- One of the four full-suite trials still tripped the global 5s thread timeout.
  That is pre-existing and unrelated (see also
  `20260803-pytest-full-suite-timeout.md`); the learning suggests raising the
  demo/integration timeout or switching to `timeout_method = "signal"` so a slow
  test fails alone instead of voiding the session.
- The autouse guard contains the blast radius but does not remove the footgun;
  the durable fix (a `config_override()` context manager, or config via a
  FastAPI dependency tests can override) is captured as a learning rather than
  done here, to keep this scoped to the flake.
- The guard is **function-scoped only**: a module-, class-, or session-scoped
  fixture pollutes the cache before the guard snapshots it, so higher-scoped
  fixtures must get the ordering right themselves. Documented in
  `test/AGENTS.md` and in the fixture docstring rather than fixed, since no
  current fixture is in that position.
- `TestNoFixtureLeakedConfig` only observes fixtures that ran *before* it in the
  session, so it is order-dependent by nature — a ratchet that catches
  regressions on most orderings, not a proof of absence on every one.

### DIFF
--- a/notes/flaky-tests.md
+++ b/notes/flaky-tests.md
@@ -26,26 +26,47 @@ and fall through to Level 2 (GitHub label search).
 | Test node ID | Issue | Last blocked |
 |---|---|---|
 | `test/bt/test_vultrabot.py::MyTestCase::test_main` | — | 2026-05-05 |
-| `test/demo/test_pcr_bootstrap.py::TestBootstrapSequence::test_announce_creates_case_replica` | #2086 | 2026-08-08 |
-| `test/demo/test_pcr_bootstrap.py::TestBootstrapSequence::test_case_fields_preserved_in_replica` | #2086 | 2026-08-08 |
-| `test/demo/test_integration_script_scenarios.py::TestIntegrationScriptScenarios::test_valid_scenarios_matches_ci` | #2122 | 2026-08-08 |
-| `test/demo/test_integration_script_scenarios.py::TestIntegrationScriptScenarios::test_at_least_one_scenario_in_ci` | #2122 | 2026-08-08 |
 
-> Note: the two `test_integration_script_scenarios` entries are **hard-broken on
-> `main`, not flaky** — they fail deterministically. #2114 added a test that
+> Note: the two `test_integration_script_scenarios` entries were **hard-broken
+> on `main`, not flaky** — they failed deterministically. #2114 added a test that
 > scrapes `DEMO=` from `demo-integration.yml` while #2118/#2119 moved the
 > scenario matrix to `.github/demo-scenarios.json`, leaving nothing to scrape.
-> A semantic merge collision between two individually-correct PRs. See #2122.
+> A semantic merge collision between two individually-correct PRs. **Fixed by
+> #2123**; rows removed 2026-08-08.
 >
-> Note: the two `TestBootstrapSequence` entries are **genuinely
-> nondeterministic**. They pass reliably in isolation (`[0,0,0]` over 3 runs)
-> but running the whole `test/demo/` directory gives `[2,0,2]` failures over 3
-> identical invocations — so `-p no:randomly` does NOT stabilise them. No single
-> preceding file reproduces the failure when paired with the target (all 27
-> checked), meaning the trigger is cumulative accumulated state plus a
-> nondeterministic interaction, not a single polluting module. File-level
-> bisection cannot converge because clean results are false negatives. Verified
-> pre-existing on clean `origin/main`. See #2086 for the full data.
+> Note: the two `TestBootstrapSequence` entries were **deterministic
+> cross-module config leakage, not nondeterminism** — root-caused and **fixed by
+> #2126**; rows removed 2026-08-08.
+>
+> An earlier revision of this note claimed they were "genuinely
+> nondeterministic" because `pytest -m "" test/demo/` gave `[2,0,2]` bootstrap
+> failures over three identical runs. That reading was an artifact of the
+> measurement. `timeout = 5` with `timeout_method = "thread"`
+> (`pyproject.toml:123`) kills the **whole pytest process**, not just the slow
+> test, so a run that trips it emits no summary line and never reaches
+> `test_pcr_bootstrap.py` at all — scoring a phantom `0`. Re-running the same
+> command with a driver that records the `+++ Timeout +++` marker reproduces
+> `[2,0,2]` and shows the `0` run aborted mid-session.
+>
+> At any granularity that actually runs both modules to completion, the failure
+> is deterministic on clean `origin/main`:
+>
+> | Invocation (`-p no:randomly`, 3× each) | Bootstrap failures |
+> |---|---|
+> | `test_fvcv_handoff_demo.py` + `test_pcr_bootstrap.py` | `[2, 2, 2]` |
+> | `...::TestOwnershipTransferAnnounceReachesFinderAC5c` + bootstrap | `[2, 2, 2]` |
+> | `test_pcr_bootstrap.py` alone | `[0, 0, 0]` |
+>
+> The earlier pairwise bisect that found "no single polluting file (all 27
+> checked)" disagrees with the first row above; treat that sweep as unreliable.
+> The cause was `reload_config()` running before `monkeypatch.undo()` in four
+> demo fixture teardowns, pinning a fake CaseActor host into the module-level
+> config cache for the rest of the session. See #2086 and PR #2126.
+>
+> **Lesson for future triage here**: a pytest run that aborts on the global
+> thread-method timeout looks identical to a clean pass if you only count
+> `FAILED` lines. Always check for a summary line and the `+++ Timeout +++`
+> marker before concluding a test is nondeterministic.
 >
 > Note: `test_vultrabot` shows `SUBFAILED` in the full suite due to py_trees
 > blackboard global-state ordering, but exit code stays 0 (unittest subtest
--- a/plan/history/2608/implementation/ISSUE-2086.md
+++ b/plan/history/2608/implementation/ISSUE-2086.md
@@ -1,0 +1,73 @@
+---
+source: ISSUE-2086
+timestamp: '2026-08-08T01:49:37.124299+00:00'
+title: Fix demo config cache leak causing flaky bootstrap validate-report
+type: implementation
+---
+
+**Issue**: [#2086](https://github.com/CERTCC/Vultron/issues/2086) — Flaky CI:
+`TestBootstrapSequence.test_announce_creates_case_replica` /
+`test_case_fields_preserved_in_replica` — `SvcValidateReportUseCase` no
+routable recipients.
+
+**PR**: [#2126](https://github.com/CERTCC/Vultron/pull/2126)
+
+## Symptoms
+
+Two `test/demo/test_pcr_bootstrap.py` tests failed intermittently in CI with
+`SvcValidateReportUseCase: no routable recipients`. Secondary symptom:
+`EnsureEmbargoExists` reporting no active embargo. Passed in isolation.
+
+## Root cause
+
+Not a race and not a production defect. `vultron/config/app.py` keeps a
+process-global `_config_cache`; `reload_config()` clears it and re-reads the
+environment. Pytest's `monkeypatch` fixture undoes env changes *after* the
+requesting fixture's teardown body runs, so a teardown that calls
+`reload_config()` before the undo re-caches the still-patched value for the
+remainder of the session.
+
+Four demo fixtures had that order, repointing `VULTRON_SERVER__BASE_URL` and
+`VULTRON_ACTOR__CASE_ACTOR_SERVICE_URL` at fake hosts. Every later demo test
+then derived its CaseActor ID from the leaked host (e.g.
+`http://coordinator-otc.test`), which no `_TestClientRouter` registers.
+`_TestClientRouter.emit` silently dropped the `Create(CaseProposal)` at DEBUG
+level, so the CaseActor never created the canonical case with its
+`CASE_MANAGER` participant and embargo, and `trigger/validate-report` failed
+three steps downstream. Because it only affects tests running after the
+offender and `pytest-randomly` reseeds order per run, it read as flakiness.
+
+## Fix
+
+- Reordered teardown to `monkeypatch.undo()` before `reload_config()` in
+  `test_fvcv_handoff_demo.py`, `test_pcr_late_joiner.py`,
+  `test_case_proposal_round_trip.py`, `test_pcr_engage_case.py`.
+  `test_fv_demo.py` already avoided the trap via an explicit `MonkeyPatch()`.
+- Added `config_url_snapshot()` / `restore_config_if_leaked()` plus an autouse
+  `restore_case_actor_url_after_each_test` fixture in `test/demo/conftest.py`
+  that snapshots the two URLs around every demo test and reloads on drift.
+- Added `test/demo/test_config_leak_guard.py` (5 tests) covering both teardown
+  orderings, the guard helpers, and the demo session's baseline CaseActor URL.
+
+No production code changed.
+
+## Discarded approach
+
+An initial fix made `_compute_report_addressees` fall through to the offer
+actor when a case had no `CASE_MANAGER`. It was reverted: it broke the
+fail-closed routing invariant from issue #1854 AC-2, asserted by
+`test_close_report_raises_when_case_exists_without_case_manager`.
+
+## Verification
+
+Deterministic repro with `-p no:randomly`; 16/16 pass after the fix. Unit
+suite 6453 passed. Integration suite 1050 passed, with 2 pre-existing
+`test_integration_script_scenarios.py` failures reproduced on a clean
+`origin/main` worktree and tracked by
+[#2122](https://github.com/CERTCC/Vultron/issues/2122). Black, flake8, mypy,
+pyright clean.
+
+## Learnings
+
+- `20260808-silent-delivery-drop-masks-config-leak.md` (`signal: concern`)
+- `20260808-module-level-config-cache-fragile-in-tests.md` (`signal: concern`)
--- a/plan/incoming/learnings/20260808-module-level-config-cache-fragile-in-tests.md
+++ b/plan/incoming/learnings/20260808-module-level-config-cache-fragile-in-tests.md
@@ -1,0 +1,47 @@
+---
+title: "Module-level config cache makes reload_config() order-sensitive in test teardown"
+type: learning
+timestamp: "2026-08-08"
+source: ISSUE-2086
+signal: concern
+---
+
+`vultron/config/app.py` holds a module-level `_config_cache`. `get_config()`
+lazily populates it; `reload_config()` clears it and re-reads the environment.
+That makes the *cache* process-global while the *environment* it derives from is
+per-test — and pytest's `monkeypatch` fixture undoes env changes **after** the
+requesting fixture's teardown body runs.
+
+So this order is a latent session-wide leak:
+
+```python
+# teardown
+reload_config()      # re-caches the still-patched fake host
+# ... monkeypatch's own undo runs here, too late
+```
+
+and this order is correct:
+
+```python
+monkeypatch.undo()
+reload_config()
+```
+
+Four demo fixtures had the buggy order (`test_fvcv_handoff_demo.py`,
+`test_pcr_late_joiner.py`, `test_case_proposal_round_trip.py`,
+`test_pcr_engage_case.py`). One (`test_fv_demo.py`) had already hit this and
+worked around it by using an explicit `MonkeyPatch()` instead of the fixture,
+with a comment explaining why — the knowledge existed but was local to one file
+and never generalized.
+
+Because the leak only bites tests that run *after* the offender, and
+`pytest-randomly` reseeds order each run, it presented as CI flakiness rather
+than a deterministic failure.
+
+**Suggested follow-up**: the durable fix is to stop reaching for a module-level
+singleton in tests — e.g. expose config via a FastAPI dependency that tests
+override, or provide a `config_override()` context manager in
+`vultron/config/app.py` that owns both the env patch and the cache invalidation
+so callers cannot get the order wrong. The current fix (source fixes plus an
+autouse snapshot/restore guard in `test/demo/conftest.py`) contains the blast
+radius but does not remove the footgun.
--- a/plan/incoming/learnings/20260808-pytest-thread-timeout-fakes-nondeterminism.md
+++ b/plan/incoming/learnings/20260808-pytest-thread-timeout-fakes-nondeterminism.md
@@ -1,0 +1,45 @@
+---
+title: "Global pytest thread-timeout aborts look like passing runs and fake nondeterminism"
+type: learning
+timestamp: "2026-08-08"
+source: ISSUE-2086
+signal: tooling-issue
+---
+
+`pyproject.toml` sets `timeout = 5` with `timeout_method = "thread"`. The thread
+method cannot cancel a single test — it kills the **entire pytest process**. The
+run then emits no `N passed / N failed` summary line and simply stops wherever it
+was.
+
+If a triage script counts `FAILED` lines to score a run, an aborted run scores
+**zero failures** — indistinguishable from a clean pass. When the abort happens
+before the suspect test is even reached, the test appears to have passed.
+
+This produced a false conclusion on #2086. Three identical
+`pytest -m "" test/demo/` runs gave `[2, 0, 2]` bootstrap failures, which was
+read as proof the failures were "genuinely nondeterministic" and that
+`-p no:randomly` did not stabilize them. Re-running with a driver that also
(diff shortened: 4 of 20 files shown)


sei-ahouseholder (commented):
## PR Triage — 12 phases complete

**CI**: all 10 checks green. **Verdict**: the fix is correctly root-caused and the mechanism analysis is excellent — but its regression coverage cannot fail, and two of the six sibling call sites classes are unaddressed.

**Findings: 2 FAIL, 4 IMPROVE, 0 NEW-ISSUE.** All findings were empirically verified, not inferred.

---

### FAIL 1 — Regression coverage for the source fix is vacuous (`test_config_leak_guard.py:31`)

The module docstring claims the tests pin *both* halves: the teardown reorder **and** the guard. Only the guard half is pinned.

Verified by reverting all four `monkeypatch.undo()` source fixes while leaving the autouse guard enabled:

| guard | source fix | result |
|---|---|---|
| ON | reverted | **25 passed — nothing fails** |
| OFF | reverted | both `TestBootstrapSequence` tests FAIL — exact #2086 symptom |
| ON | as shipped | 25 passed |

The guard repairs the leak so reliably that no test can observe a regression of the four teardown fixes. `TestDemoSessionBaseline` (line 118) is unconditionally green for the same reason, contradicting its docstring claim that "a surviving leak fails it here".

**Failure scenario**: someone tidies `test_pcr_engage_case.py:129-133` back to `reload_config(); monkeypatch.undo()`. CI stays fully green; the four inline `#2086` comments are unenforced.

**Suggested fix**: record leak detections on a session-scoped collector and assert zero detections across the demo session. Keeps the repair (no CI flakiness) while still failing when a fixture regresses.

### FAIL 2 — `restore_config_if_leaked()` claims a repair it did not perform (`conftest.py:262`)

The repair is `reload_config()`, which re-reads `os.environ`. If the env is still dirty, the reload re-caches the *same* pollution — but the function returns `True` and logs `"...reloading (#2086)"` as though it succeeded. It never re-snapshots.

Verified: guard returned `True`, post-"repair" snapshot was still `('http://localhost:7999', 'http://dirty-env.invalid/api/v2')` against a clean `before` of `('http://localhost:7999', 'http://localhost:7999/api/v2')`.

This is the project's own **"Silent `None` Returns and Fake `SUCCESS` Are the Same Bug"** pitfall (ARCH-15-001–004) in test-infrastructure form. A leak that reaches this path is now *harder* to diagnose than before the PR, because the log says it was handled.

**Suggested fix**: re-snapshot after reload; raise if drift persists. Or restore the prior `AppConfig` into `_config_cache` directly rather than reloading from a possibly-dirty env.

---

### IMPROVE 3 — Six non-demo fixtures have the identical bug, outside the guard's reach

The guard is scoped to `test/demo/conftest.py`, but the same reload-before-undo pattern exists in:

- `test/core/use_cases/received/test_submit_report_received.py:44`
- `test/core/use_cases/received/test_ack_validate_report_received.py:56`
- `test/core/use_cases/test_validate_report_ledger_routing.py:161`
- `test/core/behaviors/case/test_receive_report_case_tree.py:64`
- `test/core/behaviors/case/test_bug_26040902_regression.py:54`
- `test/core/behaviors/case/nodes/test_case_setup.py:78`

Verified the leak escapes: running `test_submit_report_received.py` then a probe in the same session shows the probe seeing `case_actor_service_url='http://case-actor:7999/api/v2'`.

Note `test/adapters/driven/test_get_datalayer.py:35-38` already documents this exact hazard and handles it correctly. The right pattern exists in-tree; these six don't follow it. Same family — one sentence covers fixing both, no "also" needed.

### IMPROVE 4 — Guard docstring overclaims (`conftest.py:305`)

"A leak cannot outlive the test that caused it" holds only for **function-scoped** offenders. For a module-scoped fixture the pollution predates the guard's `before` snapshot, so `after == before` and the guard reports clean. `test/demo/` already has module-scoped config-mutating fixtures (`test_fv_demo.py:56`, `test_fvcv_handoff_demo.py:39`) — correct order today, but zero guard protection.

### IMPROVE 5 — Inert `reload_config()` (`test_config_leak_guard.py:109`)

`_repoint_config_at_fake_host()` already reloads at line 54, so line 109 is a no-op — verified by deleting it: all 5 tests still pass. The test exercises "pollute and don't clean up", not the reload/undo *ordering* its name advertises.

### IMPROVE 6 — Snapshot covers 2 of several cacheable fields (`conftest.py:247`)

`VULTRON_ACTOR__DEFAULT_CASE_ROLES` (patched module-scoped at `test_fv_demo.py:68`), `mode`, and `pre_bootstrap_queue_timeout_seconds` are cached and invisible to the snapshot. `get_config().model_dump()` would be strictly broader at no extra cost.

---

### Phases with no findings

Issue linkage (`Closes #2086` at top, scope matches, no creep) · PR body format (all sections, real test counts, Co-Authored-By in all 3 commits) · specs (test-only) · ADR (no architectural signals) · notes/docs (`flaky-tests.md` correctly corrected, 3 learnings filed, valid frontmatter) · CI (10/10 green).

The root-cause analysis and the timeout-measurement-artifact correction are genuinely strong work — the gap is that the fix isn't pinned by a test that can fail.

sei-ahouseholder (commented):
## PR Execute: #2126 — fix(test): stop demo config cache leaking across tests

**Fixes applied**: 1 commit (`0b7a90d6`)
**Issues filed**: 0
**Tests run**: unit + integration
**CI status after push**: ✅ passing (all 10 checks green against `0b7a90d6`)

---

### Fixed

| Finding | Commit |
|---|---|
| `phase10-guard-masks-source-fix-0` (**FAIL**): autouse guard repaired leaks so reliably that no test could observe a regression of the four teardown fixes | `0b7a90d6` |
| `phase10-guard-false-repair-claim-0` (**FAIL**): guard returned `True` while pollution survived | `0b7a90d6` |
| `phase10-source-fix-incomplete-outside-demo-0` (IMPROVE): six non-demo fixtures had the same leak, unguarded | `0b7a90d6` |
| `phase10-guard-blind-to-higher-scopes-0` (IMPROVE): guard is function-scoped only | `0b7a90d6` |
| `phase10-dead-reload-line-0` (IMPROVE): unreachable reload in test helper | `0b7a90d6` |
| `phase10-snapshot-two-fields-0` (IMPROVE): snapshot covered only two URL fields | `0b7a90d6` |

### The two FAILs were the same underlying problem

Triage proved the regression coverage was vacuous rather than asserting it. With
the guard on and **all four source fixes reverted**, `test/demo/` reported
`25 passed` — nothing failed:

| guard | source fixes | result |
|---|---|---|
| off | present | 20 passed (control) |
| on | **reverted** | **25 passed — nothing fails** |
| off | reverted | both `TestBootstrapSequence` tests FAIL, "no routable recipients" |

Fixed with `ConfigLeakLedger` + `TestNoFixtureLeakedConfig`: the guard still
repairs (one bad teardown cannot cascade), but every repair is recorded and
asserted on. Validated by reverting two *different* fixtures in turn — each
produced a failing ratchet naming the offending host.

The false-repair bug was likewise proven with a probe, not inferred: the guard
returned `True` while `case_actor_service_url` was still
`http://dirty-env.invalid/api/v2`. `restore_config_if_leaked()` now verifies the
post-reload state and raises (ARCH-15-001 — a fake `SUCCESS` is the same bug as
a silent `None`).

### Also closes #1897 — surfaced, not caused, by this PR

Removing the out-of-demo leak broke 4 tests that had been silently depending on
it (`test_create_tree.py` ×3, `test_communication.py` ×1). Both modules run
`ResolveCaseActorUrlsNode`, which FAILs when `case_actor_service_url` is None
(CP-08-002/003), without ever setting it.

Causality established by matched runs at identical fixed order (`-p no:randomly`):

| base | result |
|---|---|
| clean branch HEAD, full suite | 6453 passed, 0 failed |
| + six-fixture leak fix, full suite | exactly those 4 fail |
| clean `origin/main`, those 2 modules as a subset | same 4 fail |

The third row is #1897, filed earlier and still open. Fixed properly — each
module configures the URL in its own autouse fixture — rather than by restoring
the leak.

### Verification

- Unit suite: `6453 passed, 362 skipped, 2 xfailed, 5552 subtests passed`.
- Full suite incl. integration and demo (`-m ""`):
  `7511 passed, 362 skipped, 3 xfailed, 1 xpassed, 5552 subtests passed` — **0 failures**.
- xfail ratchet: #1991, #1992, #1993, #1994 all confirmed open.
- Black, flake8, mypy, pyright, markdownlint clean; all pre-commit hooks passed.

### Filed as Issues

None — all 6 findings were `fix-now`.

### Deferred — Awaiting Your Input

None.

### Skipped

None.

---

*Execute artifact: `.claude/pr-2126-execute.json`*

sei-ahouseholder (commented):
## PR Verify: #2126 — fix(test): stop demo config cache leaking across tests

**Findings verified**: 6 / 6 (execute results == triage findings, IDs match)
**CI status**: ✅ all 10 checks pass against `0b7a90d6`
**Integration tests**: ✅ covered — CI `Tests (pytest)` runs `uv run pytest -m ""`

---

### Per-Finding Verdicts

|---|---|---|---|
| `phase10-guard-masks-source-fix-0` | FAIL | fixed `0b7a90d6` | **CONFIRMED** |
| `phase10-guard-false-repair-claim-0` | FAIL | fixed `0b7a90d6` | **CONFIRMED** |
| `phase10-source-fix-incomplete-outside-demo-0` | IMPROVE | fixed `0b7a90d6` | **CONFIRMED** |
| `phase10-guard-blind-to-higher-scopes-0` | IMPROVE | fixed `0b7a90d6` | **CONFIRMED** |
| `phase10-dead-reload-line-0` | IMPROVE | fixed `0b7a90d6` | **CONFIRMED** |
| `phase10-snapshot-two-fields-0` | IMPROVE | fixed `0b7a90d6` | **CONFIRMED** |

All six checked at **HEAD**, not merely in the commit diff. Working tree clean;
local HEAD == `origin/bug/2086-flaky-bootstrap-validate-report` == `0b7a90d6`.

### The FAIL findings were verified behaviorally, not structurally

Since the finding was that this PR's regression coverage was *vacuous*, grepping
for the ratchet's presence would have repeated the original mistake. So the
ratchet was re-tested by reverting a fixture and confirming it fires:

- Reverted `test_pcr_late_joiner.py` to the buggy order (`reload_config()` before
  `monkeypatch.undo()`) → `test_guard_recorded_no_leaks` **FAILED**, naming the
  offending host in its drift message
  (`case_actor_service_url: 'http://localhost:7999/api/v2' -> 'http://owner-late-joiner.test/api/v2'`).
- Fixture restored; tree clean.

`restore_config_if_leaked()`'s post-reload verification and `RuntimeError` path
confirmed present at `test/demo/conftest.py:347`; all 8 guard tests pass.

### Notes

- **Demo Integration did not run on this branch — correctly.** Its `pull_request`
  path filter covers `vultron/**`, `docker/**`, `integration_tests/**`,
  `test/ci/**`, `pyproject.toml`, `uv.lock`, and workflow files. This PR is
  test-only under `test/core/`, `test/demo/`, `test/AGENTS.md`, `notes/`, and
  `plan/`, so it touches no filter path. Integration coverage comes from
  `Tests (pytest)` running the full `-m ""` suite, which is green.
- `#1897` is also closed by this PR. It was surfaced (not caused) by the
  out-of-demo leak fix, and fixed properly — each affected module now configures
  `case_actor_service_url` itself rather than borrowing another module's leak.

### Deferred — Awaiting Your Input

None.

---

*Artifacts cleaned up. Verified against `0b7a90d6`.*

PR #2951, opened 2026-09-01
Treat an unset PXA dimension as no threat present

### LINKED ISSUE #2877: bug: _threat_present() swallows AttributeError when pxa is None — silently skips embargo termination
## Summary

`_threat_present()` in `threat_termination.py` accesses `getattr(case_status, "pxa").state` without guarding for `pxa is None`. When a participant has no PXA dimension set (a valid state), this raises `AttributeError`, which is caught by a broad `except Exception: return False`. The function silently reports "no threat" and embargo termination is skipped with no log entry.

## Location

`vultron/core/behaviors/status/nodes/threat_termination.py` ~line 101

## Failure scenario

1. A case has an active embargo that should be terminated.
2. At least one participant's `CaseStatus` has `pxa=None` (PXA dimension not yet set).
3. `_threat_present()` raises `AttributeError` on `.state`.
4. The broad `except Exception` catches it and returns `False`.
5. Embargo termination is skipped. No warning or error is logged.

## Fix

Add an explicit guard before the `.state` access:

```python
if case_status.pxa is None:
    return False
```

`None` is the correct "no threat" answer — a participant with no PXA dimension has not asserted any public-exploit or attack status. Making this explicit also allows the broad `except Exception` to be narrowed or removed, so future `AttributeError`s from other causes do not disappear silently.

## Issue type

Bug

arpitjain099:
Fixes #2877.

`_threat_present()` reads `getattr(case_status, "pxa").state` without checking whether `pxa` is set, so a participant whose PXA dimension is still `None` blows up on the attribute access. The guard is the one the issue proposes: an unset PXA dimension is an assertion of nothing, which is not a threat, so it returns `False`.

One correction to the report, since it changes what an operator sees. There is no broad `except Exception` around that access; the only `try` in the function wraps the `pxa_state != CS_pxa.pxa` comparison further down. So this does not silently skip embargo termination, it raises:

```
>>> node = _ThreatTerminationSkipConditionNode(status_obj=SimpleNamespace(case_status=SimpleNamespace(pxa=None)), case_id="case-1")
>>> node._threat_present()
AttributeError: 'NoneType' object has no attribute 'state'
>>> node.update()
AttributeError: 'NoneType' object has no attribute 'state'
```

The fix is the same either way, and the outcome afterwards is the skip the issue describes, this time deliberately.

`test/core/behaviors/status/nodes/test_threat_termination.py` covers the unset dimension, the all-lowercase `pxa` state, and each of P, X and A being set. The first case fails on main with the `AttributeError` above.

`pytest test/core/behaviors/status/nodes/test_threat_termination.py` passes. Three tests in `test/core/behaviors/status/test_partial_accept_participant_status.py` fail in my checkout both with and without this change, so they look unrelated.

### DIFF
--- a/test/core/behaviors/status/nodes/test_threat_termination.py
+++ b/test/core/behaviors/status/nodes/test_threat_termination.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+
+#  Copyright (c) 2026 Carnegie Mellon University and Contributors.
+#  - see Contributors.md for a full list of Contributors
+#  - see ContributionInstructions.md for information on how you can Contribute to this project
+#  Vultron Multiparty Coordinated Vulnerability Disclosure Protocol Prototype is
+#  licensed under a MIT (SEI)-style license, please see LICENSE.md distributed
+#  with this Software or contact permission@sei.cmu.edu for full terms.
+#  Created, in part, with funding and support from the United States Government
+#  (see Acknowledgments file). This program may include and/or can make use of
+#  certain third party source code, object code, documentation and other files
+#  ("Third Party Software"). See LICENSE.md for more details.
+from types import SimpleNamespace
+
+import pytest
+
+from vultron.core.states.cs import CS_pxa
+from vultron.core.behaviors.status.nodes.threat_termination import (
+    _ThreatTerminationSkipConditionNode,
+)
+
+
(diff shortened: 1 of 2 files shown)


ahouseholder:
Hi @arpitjain099, thanks for the PR. We're not quite ready to take PRs from outside the project team yet though so I'm going to close this unmerged.

PR #2909, opened 2026-08-31
refactor(core): eliminate all isinstance(VulnerabilityCase) guards from vultron/core/

sei-ahouseholder:
- Closes #2490

## Summary

Removes all 120 `isinstance(x, VulnerabilityCase)` guards from `vultron/core/` by adding a typed `read_case() -> VulnerabilityCase | None` method to `CasePersistence` and narrowing `find_case_by_report_id` / `find_case_by_short_id` return types. All call sites switch from `dl.read()` + isinstance guard to `dl.read_case()` + `is None` check; the isinstance logic is confined to the SQLite adapter where it belongs.

## Changes

- **`vultron/core/ports/case_persistence.py`**: Added `read_case()` to protocol; narrowed `find_case_by_*` return types to `VulnerabilityCase | None`
- **`vultron/adapters/driven/datalayer_sqlite/datalayer.py`**: Implemented `read_case()` and updated `find_case_by_*` with isinstance in adapter (allowed per AC-1)
- **`vultron/core/behaviors/` (54 files)**: Replaced `dl.read()` + `isinstance` with `dl.read_case()` + `is None`; removed now-unused `VulnerabilityCase` imports from 36 files
- **`vultron/core/services/embargo_lifecycle.py`**: 7 guards replaced
- **`vultron/core/use_cases/`**: 14 guards replaced across received/ and triggers/
- **`vultron/core/AGENTS.md`**: Updated pitfall entry — isinstance checks no longer needed after `read_case()`; use `type_` duck-typing for wire-side validation
- **`vultron/core/models/_helpers.py`**: `cast(str, ...)` restores concrete return type after isinstance removal
- **`vultron/core/behaviors/case/nodes/participant/participant_add.py`**, **`owner.py`**: `cast(VulnerabilityCase, ...)` for blackboard-sourced slots (data_type=object ports)
- **`test/core/behaviors/test_performance.py`**: Added `_mock_read_case_helper` and `dl.read_case.side_effect` to `mock_datalayer` fixture (was missing after `read_case()` migration)
- **`vultron/core/behaviors/report/nodes/conditions.py`**: `EnsureEmbargoExists` now separately logs "case not found" vs "no active embargo" for clearer operator diagnostics
- **`specs/datalayer.yaml`**: DL-04-002 updated to include `read_case` in the required minimum `CasePersistence` surface
- **`notes/datalayer-design.md`**: Added `read_case()` as preferred accessor for `VulnerabilityCase` retrieval

## Verification

[...]

PR #2882, opened 2026-08-31
refactor(ownership-transfer): move ForwardOffer outbox write into BT effect node

sei-ahouseholder:
- Closes #2067

## Summary

Eliminates the procedural `add_activity_to_outbox` call in `OfferCaseOwnershipTransferReceivedUseCase.execute()` that violated ADR-0022/CLP-10-005, moving the CaseActor's forwarding responsibility into a proper BT effect node (`ForwardOfferToTransfereeNode`) wrapped by `create_case_manager_gated_tree`.

## Changes

- **`vultron/core/behaviors/case/nodes/ownership_transfer.py`**: Adds `ForwardOfferToTransfereeNode` extending `_EmitSingleActivityBase`. Overrides `update()` to log WARNING with "no trigger_activity" and return FAILURE when `trigger_activity_factory` is absent from the blackboard. `_call_factory()` calls `offer_case_ownership_transfer()` with `actor=self.actor_id`, `to=[transferee_id]`, `attributed_to=original_actor_id` (CM-21-005, ADR-0053).
- **`vultron/core/behaviors/case/ownership_transfer_tree.py`**: Adds `create_offer_ownership_transfer_tree(case_id, transferee_id, original_actor_id)` factory. Wraps `ForwardOfferToTransfereeNode` in `create_case_manager_gated_tree` and passes it as `effect_nodes` to `create_receive_activity_tree`. Omits the forwarding node when `transferee_id` or `original_actor_id` is `None` so the ledger-commit gate still fires.
- **`vultron/core/use_cases/received/actor/ownership.py`**: Refactors `execute()` to use `create_offer_ownership_transfer_tree` and pass `trigger_activity=self._trigger_activity` to `BTBridge`. Removes all procedural forwarding code, `_resolve_case_manager_id`, `add_activity_to_outbox`, and `VulnerabilityCase` imports from this file.
- **`vultron/core/behaviors/case/nodes/__init__.py`**: Exports `ForwardOfferToTransfereeNode`.
- **`test/core/behaviors/case/nodes/test_offer_ownership_transfer_tree.py`**: 5 new tests — CaseActor forward queues in outbox, non-CM role gate skips node, absent factory warning, and two tree-factory None-argument guards.

## Verification

- 15 tests pass for ownership-transfer paths (10 pre-existing + 5 new)
- 7882 unit tests pass, integration suite passes
- Black, flake8, mypy, pyright clean

PR #3013, opened 2026-09-01
fix(demo): route ownership transfer through the CaseActor per ADR-0053

### LINKED ISSUE #2067: refactor(ownership-transfer): move ForwardOffer outbox write into BT effect node
## Problem

`OfferCaseOwnershipTransferReceivedUseCase.execute()` contains a procedural
outbox write (`add_activity_to_outbox`) **after** `bridge.execute_with_setup()`.

Per AGENTS.md:
> Protocol-Significant Behavior: Any action that affects protocol-observable
> state (emitting an Activity…) MUST be implemented as BT nodes/subtrees,
> never as procedural code outside the tree.

The current code is therefore non-conformant: the forward-to-transferee outbox
write is outside the BT.

## Required change

Introduce a `ForwardOfferToTransfereeNode` BT node (or similar) and pass it as
an `effect_node` to `create_receive_activity_tree(…, effect_nodes=[ForwardOfferToTransfereeNode(…)])`.

This requires:
1. A new `ForwardOfferToTransfereeNode` class in
   `vultron/core/behaviors/case/nodes/ownership_transfer.py`
2. Refactoring `OfferCaseOwnershipTransferReceivedUseCase.execute()` to remove
   the procedural `add_activity_to_outbox` call
3. Updating the relevant BT tree factory (`create_receive_activity_tree` call)
   in `ownership.py`

## Context

Surfaced during PR #2044 triage (finding phase8-outbox-write-outside-bt-0).
The fix was intentionally deferred because it requires a non-trivial BT node
— designing and testing `ForwardOfferToTransfereeNode` is a meaningful effort
distinct from the routing fix that PR #2044 delivers.

## References

- PR #2044 (ownership-transfer routing via CaseActor)
- ADR-0053
- CM-21-005
- AGENTS.md § Protocol-Significant Behavior

### LINKED ISSUE #2789: fix(demo): implement ADR-0053 CaseActor routing in transfer_ownership_demo.py; create fvcv_handoff_demo.py
## Summary

The BT nodes (`EmitOfferCaseOwnershipTransferNode`, `EmitAcceptCaseOwnershipTransferNode`) were fixed in PR #2044 to address the CaseActor per ADR-0053. However, `vultron/demo/exchange/transfer_ownership_demo.py` still uses direct P2P routing via `post_to_inbox_and_wait`. The `fvcv_handoff_demo.py` referenced in ADR-0053's validation section does not exist yet. This issue also encompasses the remaining BT cleanup in #2067.

## Acceptance Criteria

- [ ] AC-1: `transfer_ownership_demo.py` updated to route Offer and Accept through the CaseActor (not directly to coordinator/vendor inbox) per ADR-0053 concrete routing model
- [ ] AC-2: `vultron/demo/exchange/fvcv_handoff_demo.py` created — full FVCV-handoff scenario including ownership transfer; Finder receives `Announce(CaseLedgerEntry)` for the transfer (the validation criterion from ADR-0053)
- [ ] AC-3: `ForwardOfferToTransfereeNode` BT effect node introduced per #2067, removing the procedural `add_activity_to_outbox` call from `OfferCaseOwnershipTransferReceivedUseCase`
- [ ] AC-4: Existing ownership-transfer tests pass; new integration test verifies Finder notification via ledger broadcast

## Reference

Source: #602
Docs PR: https://github.com/CERTCC/Vultron/pull/2784
ADR-0053: `docs/adr/0053-ownership-transfer-routed-via-caseactor.md`
Notes: `notes/ownership-transfer.md`
Companion issue: #2067 (ForwardOffer BT effect node — close as part of this work)

### LINKED ISSUE #3012: bug(wire): _build_activity_snapshot drops attributed_to, making delegated authorship unrecoverable in core
## Summary

`_build_activity_snapshot` in `vultron/wire/as2/extractor/_builders.py` did not
copy `attributed_to` onto the `VultronActivity` snapshot it builds for every
received event. Because that snapshot is what a received-side use case sees as
`request.activity`, the delegated author of a message was unrecoverable in core:
`request.actor_id` names the *sender*, which for a delegated message is the
CaseActor (CM-24-001), and `attributed_to` — the field CM-24-002 exists to carry
"so that receivers can recover the originating identity" — was always `None`.

The visible consequence was in the ownership-transfer forward hop:
`OfferCaseOwnershipTransferReceivedUseCase` passed
`original_actor_id = request.actor_id` to `ForwardOfferToTransfereeNode`, so the
Offer the CaseActor forwarded to the transferee named **the CaseActor** as
`attributed_to` rather than the vendor who asked for the transfer. Every replica
that materialised `VultronOwnershipTransferOfferRecord.actor_id` from the ledger
snapshot recorded the same wrong offerer.

Found by the pre-PR code review on #2789.

## Fixed in #2789

Both halves landed with the #2789 build, because the demo added there is the
first thing that exercises this path end-to-end (the exchange demo previously
posted the Offer straight to the transferee):

- `_build_activity_snapshot` now carries
  `attributed_to=_get_id(getattr(activity, "attributed_to", None))`.
- `OfferCaseOwnershipTransferReceivedUseCase` now reads
  `_as_id(request.activity.attributed_to) or request.actor_id`, so the
  no-CaseActor direct-send case (CM-24-003) still resolves.
- Regression test:
  `test_forwarded_offer_attributes_the_requesting_participant` in
  `test/core/use_cases/received/actor/test_ownership.py`.

## Why this is filed rather than folded in silently

The same snapshot feeds `CaseLedgerEntry.payloadSnapshot`, so adding a field to
it changes snapshot content — and therefore `entry_hash` — for every delegated
message (invite, ownership transfer, and any future delegated flow). The full
unit tier (8015 passed) and integration tier (1249 passed) are green with the
change, and no test asserts an exact snapshot dict or hash, but the
canonical-snapshot content question belongs to CLP-07 and deserves to be
recorded rather than buried in a demo-routing PR.

## Follow-up worth checking

- Are there other received-side use cases that read `request.actor_id` where
  they mean "who asked for this", now that `attributed_to` is available? The
  invite/accept handshake (PCR-08-007/008) is the obvious peer to audit.
- `ForwardOfferToTransfereeNode`'s docstring already claimed
  `attributed_to=original_offerer`; it was accurate about intent and wrong about
  the value it received. Other node docstrings may carry the same latent claim.

sei-ahouseholder:
- Closes #2789
- Closes #2067
- Closes #3012

## Summary

Routes the ownership-transfer `Offer` and its `Accept`/`Reject` through the
CaseActor in `transfer_ownership_demo.py` per ADR-0053, restores the
`ForwardOfferToTransfereeNode` BT wiring that a catch-up merge silently reverted,
and fixes the extractor gap that made a delegated message's author unrecoverable
in core.

## Changes

- **`vultron/core/use_cases/received/actor/ownership.py`**: `execute()` builds
  `create_offer_ownership_transfer_tree` and hands the `TriggerActivityPort` to
  `BTBridge`, so the CM-21-005 forward happens inside the CM-gated effect node
  (CLP-10-005). This is #2882's change, which `main` had lost — see below. Also
  reads the delegated author from `attributed_to`, falling back to `actor_id` for
  the no-CaseActor case (CM-24-003).
- **`vultron/wire/as2/extractor/_builders.py`**: `_build_activity_snapshot` now
  carries `attributed_to`. It previously dropped it, so the field CM-24-002 exists
  to carry — "so that receivers can recover the originating identity" — was always
  `None` in core, and the CaseActor forwarded an Offer attributing the vendor's
  intent to itself (#3012).
- **`vultron/demo/exchange/transfer_ownership_demo.py`**: both paths now POST the
  `Offer` to the CaseActor's inbox with `actor=case_actor_id,
  attributed_to=vendor` (CM-24-001/002), gate on the CaseActor's *forwarded* Offer
  reaching the transferee, and address the response back to the CaseActor
  (CM-21-006).
- **`vultron/demo/helpers/workflow.py`**: three new shared helpers —
  `setup_canonical_case` (report → validate → `Create(CaseProposal)` → CaseActor,
  extracted from `initialize_participant_demo`),
  `case_actor_invites_actor_to_case` (the CaseActor-routed Invite/Accept
  handshake), and `await_forwarded_ownership_transfer_offer`.
- **`vultron/demo/helpers/polling.py`**: extracts
  `case_actor_participant_id_in(case)` as the read-free half of
  `find_case_actor_participant_id`, for callers that already hold the case.
- **`vultron/demo/scenario/fvcv_handoff_demo.py`**: `_phase_ownership_handoff`
  now asserts ADR-0053's own validation criterion — the Finder's replica holds the
  `accept_case_ownership_transfer` ledger entry, which can only reach it via the
  CaseActor's `Announce(CaseLedgerEntry)` fan-out (CM-21-007).
- **`notes/ownership-transfer.md`**, **`docs/howto/.../transfer_ownership.md`**:
  the note claimed #2067 was wired (it was not, see below); the how-to still
  carried a "Demo shortcut" admonition saying the demo bypassed the CaseActor.

## Two things worth a reviewer's attention

**#2882 was reverted by a merge, and no test noticed.** `task/2490` branched from
a pre-#2882 `main`; its catch-up merge (landing as #2909) resolved the conflict in
`ownership.py` by keeping its own side, restoring the procedural
`add_activity_to_outbox` call. Everything else from #2882 survived — the node, the
tree factory, the export, the tests — so `create_offer_ownership_transfer_tree`
sat unreachable with only test callers. Both shapes queue the same activity id in
the same outbox with the same factory kwargs, which is exactly why every
behavioural assertion passed either way. The new
`test_offer_cascade_forward_lives_in_the_bt_not_in_execute` asserts the tree
identity instead, and fails against the reverted code (verified).

**AC-2 and part of AC-4 were already delivered, under different paths.** The issue
asked for `vultron/demo/exchange/fvcv_handoff_demo.py`; a full FVCV-handoff
scenario has existed at `vultron/demo/scenario/fvcv_handoff_demo.py` since
ADR-0053 was written, and `exchange/` is for single-exchange demos against one
container. Creating the named file would have produced a second FVCV demo in the
wrong package. Likewise AC-4's "new integration test verifies Finder notification
via ledger broadcast" already exists as
`TestOwnershipTransferAnnounceReachesFinderAC5c`. The one genuine gap — the
scenario itself never asserted the Finder-notification criterion — is closed here.

## Verification

- 8015 unit tests pass (3 new), 1249 integration tests pass
- Verified the AC-3 regression test fails against the reverted implementation
- `mkdocs build --strict` still aborts on the 101 pre-existing `markdown_exec`
  warnings documented in `plan/incoming/learnings/20260831-mkdocs-strict-101-pre-existing-warnings.md`;
  none reference the page changed here


sei-ahouseholder (commented):
## Triage — PR #3013

**Merge state:** MERGEABLE (CLEAN) against `main` · **CI:** 20 success / 2 skipped on `1dd0dc81` (incl. 4 Demo Integration + 4 Invariant Harness jobs)

**8 findings: 3 FAIL, 5 IMPROVE, 0 NEW-ISSUE.** All are `fix-now`. Two sub-agent findings were investigated and **refuted** — see the bottom section, because the reasoning matters for future reviews of this area.

---

### FAIL

**1. `vultron/wire/as2/extractor/_builders.py:121` — a non-string `attributedTo` becomes a Python repr in the ledger**

AS2 permits `attributedTo` to be an object or an array, and `as_Object.attributed_to` is typed `Any | None`. `_get_id` falls through to `getattr(field, "id_", str(field))`, so a parsed inline dict yields `"{'id': 'https://b/y', 'type': 'Person'}"` and a list yields `"['https://b/y', 'https://c/z']"`. Before this PR the value was dropped, so the flaw was unreachable for this field — **this PR makes it reachable**. It passes `NonEmptyString`, so nothing raises: the garbage is hashed into `entry_hash`, broadcast to every participant, and becomes the forwarded Offer's `attributed_to` — the exact field CM-24-002 exists to make recoverable.

> Fix: resolve dict/list forms and return `None` when unresolvable, rather than a repr. A repr is never a valid URI, so no caller can depend on the current fallback.

**2. `vultron/core/use_cases/received/actor/ownership.py:77` — peer-supplied `attributed_to` is relayed unvalidated**

`_as_id(request.activity.attributed_to) or request.actor_id` goes straight to `ForwardOfferToTransfereeNode`. The receive tree has `precondition_guards=[]`, and nothing in `vultron/core` validates `attributed_to` (CLP-07-003 only inspects `payloadSnapshot.actor`). Before this PR the value was the sender of record and could not be spoofed.

> Scenario: vendor2 — a participant, not the owner — POSTs `Offer(case, target=X)` with `actor=vendor2, attributedTo=vendor1`. The CaseActor ledgers `attributed_to=vendor1` and forwards to X an Offer that appears to carry the case owner's intent.
>
> Fix: honour `attributed_to` only in the delegated shape CM-24-001 defines — when the inbound `actor` is the case's CaseActor — else fall back to `request.actor_id`. Verified that production always sends `actor=case_actor_id` (`_prepare_delegated_context`), so this keeps production behaviour identical while restoring the pre-change value for the spoofing case. Spec-grounded, not a new design decision.

**3. `vultron/demo/AGENTS.md:100` — the self-delivery pitfall contradicts ADR-0053 and this PR**

The section claims `AcceptCaseOwnershipTransferTriggerBT` addresses the Accept "only to the **offering** actor" and that a demo therefore **MUST** self-deliver it. `EmitAcceptCaseOwnershipTransferNode._call_factory` sets `to=[case_actor_id]` (CM-21-006), so the premise is false. Verified: **zero** `post_to_inbox_and_wait` calls remain in either `fvcv_handoff_demo.py` or `fccv_handoff_demo.py` — yet the section cites both "for the correct pattern", and its ✅/❌ example teaches the opposite of ADR-0053. The same stale claim is repeated at line 193 in the "Never Carry One Actor's Mail" exception paragraph.

> An agent writing the next CaseActor-routed demo would follow this MUST and reintroduce the cross-actor mail-carry that CONCERN-1635 and ADR-0053 eliminated, with AGENTS.md as justification.

---

### IMPROVE

**4. `vultron/demo/scenario/fccv_handoff_demo.py:257`** — the sibling handoff scenario did not get the Finder-notification check this PR added to fvcv. AGENTS.md "Fix One, Miss the Siblings" mandates the scan. Without it, a CM-21-007 fan-out regression in the FCCV topology passes green while the coverage matrix reports the cascade as verified.

**5. `vultron/demo/exchange/transfer_ownership_demo.py:252`** — Step 5 verifies `attributed_to` with a one-shot read gated only on `post_to_inbox_and_wait`'s fixed 1 s sleep, where the new routing does strictly more async work than the old (receipt commit + Announce fan-out + role change). `wait_for_case_attributed_to` exists and is what fvcv uses for this assertion.

**6. `notes/ownership-transfer.md:9`** — body now cites CLP-07, EDF-06, ADR-0073 and ADR-0081, but `related_specs` still lists only `specs/case-management.yaml` (NF-06-001). The frontmatter validator passes because required fields are present, so a `deepen-context` run for a CLP-07 task will not surface this note's new payload-snapshot guidance.

**7. `vultron/demo/helpers/polling.py:793`** — `case_actor_participant_id_in` was extracted with no direct test; `find_case_actor_participant_id` is only ever monkeypatched, so the predicate inherits zero coverage, and `setup_canonical_case` now asserts on its result.

**8. `test/wire/as2/test_extractor.py`** — the `attributed_to` change is asserted only incidentally, inside a core ownership use-case test. The changed code is wire-layer and applies to every received event; a future extractor refactor would look for the assertion here. That layering mismatch is how the field came to be missing in the first place.

---

### Refuted sub-agent findings

Two findings claimed the demo fails outright at Step 1 and Step 5 because `Invite.target` and the rebuilt `Offer.target` are bare URIs, producing `VultronCanonicalEntryError` on the CaseActor's commit. Both were **refuted** by probing the real flow rather than the use case in isolation:

```
CaseActor ledger after setup:  ... (8, accept_invite_actor_to_case, recorded), (9, add_case_participant, recorded)
CaseActor ledger after accept: ... (11, accept_case_ownership_transfer, recorded)
case.attributedTo:             …/actors/coordinator
coordinator participant roles: ['coordinator', 'case_owner']   ← CM-21-002 grant fired
accumulated demo failures:     []
```

The mechanism both repros missed: the inbox pipeline's `RehydrateActivityNode` expands bare reference URIs from the **receiving actor's own store** before dispatch. The Invite's `target` (a case URI) resolves from the CaseActor's store; the rebuilt Offer's `target` (a coordinator URI) resolves because `seed_peer` registers the coordinator there — which is precisely the precondition this PR adds and documents. Constructing the received-side use case directly with a hand-built event skips that node, so the bare URI survives to the commit boundary and the failure looks real.

Worth remembering for this area: a canonical-entry `payloadSnapshot` assertion tested outside the inbox pipeline will over-report bare-reference failures.

---

### Also cleared

- **Extractor blast radius — clean.** `attributed_to` is not in `_INLINE_OBJECT_KEYS` (cannot trip `_bare_inline_object_path`); not read by `_snapshot_type`/`_snapshot_object_type` (signature resolution unchanged); not in `_AS_OBJECT_REF_FIELDS` (`_dehydrate_data` passes it through, `create`→`read` round-trips); `build_activity_payload_snapshot` is the single snapshot builder, so `_find_equivalent_recorded_entry` cannot see two shapes for one logical entry; `parser.py` treats `payloadSnapshot` as opaque so `verify_hash()` still matches on replicas.
- **CLP-07-011 conformance improves** — the snapshot MUST be "the verbatim AS2 activity that was asserted"; carrying `attributed_to` makes it strictly more verbatim.
- Empty-string `attributed_to` is unreachable (`as_Offer` rejects it at construction), so `_get_id` cannot feed `""` into a `NonEmptyString` field.
- `demo_gate` nesting is correct — all three context managers funnel through `_demo_accumulate`, so a Step-5 failure is recorded and surfaced by `assert_demo_success()`.
- Rebuilt-Offer id reuse is safe — `_store_nested_inbox_object` uses `dl.create`, which raises on collision and is caught.
- PR body format, issue linkage, ADR treatment (ADR-0053 revised in place per `docs/adr/index.md`), and the `assert` convention in `setup_canonical_case` all pass.
- `Closes #2067` references an already-closed issue (closed by #2882, whose change was then reverted). Accurate as documentation; no action.

sei-ahouseholder (commented):
## Execute — PR #3013

**All 8 triage findings fixed. CI green: 22/22.** Merge state MERGEABLE (CLEAN) against `main`. 3 CI-loop iterations.

`8087` unit / `1254` integration tests pass; black, flake8, mypy, pyright clean.

### FAIL findings

| Finding | Fix |
|---|---|
| `_get_id` reprs a non-URI `attributedTo` | Resolves dict and array forms; returns `None` rather than `str(field)` when no id is recoverable. Covers the `origin`/`context`/`in_reply_to` call sites uniformly. `455f3a31a` |
| Unvalidated relay of peer `attributed_to` | New `_resolve_offering_actor_id()` honours it only when the sender is this case's CaseActor — the delegated shape CM-24-001 defines — and logs the refusal. Production sends `actor=case_actor_id` (`_prepare_delegated_context`), so production is unchanged; the spoofing case falls back to the sender. `455f3a31a` |
| Stale self-delivery pitfall in `vultron/demo/AGENTS.md` | **Resolved by taking `main`'s version.** See below. `9e8f193fc` |

### IMPROVE findings

- **fccv sibling gap** — added the CM-21-007 broadcast check to `fccv_handoff_demo`, and a new AST ratchet (`test_handoff_scenarios_verify_broadcast.py`) parametrized over every scenario defining `_phase_ownership_handoff`, so the next one cannot omit it. Verified non-vacuous by removing the check and watching it fail.
- **One-shot `attributed_to` read** — accept path now polls `wait_for_case_attributed_to`. The reject path keeps its one-shot read on purpose (it asserts a *non*-change, and a rejection commits no observable) and is now labelled temporal per EDF-06-006.
- **Notes frontmatter** — `related_specs`/`related_notes` updated with reverse links (NF-06-001). `notes/datalayer-design.md` was considered and left out: the ADR-0073 reference there is in passing, not substantive.
- **`case_actor_participant_id_in`** — 5 tests, including that it follows `CASE_ACTOR_SLUG` and that `find_case_actor_participant_id` delegates to it.
- **Extractor test placement** — direct wire-layer tests for `attributed_to` survival plus the malformed-input cases.

### Three things worth flagging

**1. `main` had already fixed the stale pitfall, better than I did.** A concurrent docs restructure (IDEA-605) condensed `vultron/demo/AGENTS.md` and deleted the self-delivery rule outright, replacing it with "There is **no self-delivery exception**" plus a normative `notes/ownership-transfer.md` § "The Accepting Actor's Replica Updates via the CaseActor's Announce". I dropped my inverted section in favour of `main`'s and verified `main`'s version is self-consistent — both `AGENTS.md` and `notes/demo-scenario-authoring.md` link to that section, and it exists.

**2. A merge conflict, not a test failure, was blocking CI.** After iteration 2 the PR went `CONFLICTING`/`DIRTY`, so GitHub could not compute a merge commit and scheduled **none** of the `pull_request`-triggered workflows. Only the push-triggered CodeQL/Analyze jobs ran — and `wait-for-ci` read that as **3/3 passing**. A partial check set reporting green is the inverse of the "a red job is not evidence its assertions ran" trap, and it is worth knowing that `total` in the CI summary can silently shrink from 22 to 3. Credit to @sei-ahouseholder for asking the question that pointed straight at it.

**3. One of my own verification steps was a false pass.** I checked the new ratchet was non-vacuous with `git stash push -- <file>`, but the change was committed by then, so the stash was a no-op and the ratchet "passed" against a file that still had the check. Re-verified by editing the file in place; it does fail correctly. Committed as part of `b84981021`.

### Refuted triage findings — unchanged

The two bare-URI `target` findings remain refuted; nothing in this phase changed that. `RehydrateActivityNode` expands those references from the receiving actor's own store, and the CaseActor's ledger contains all three expected entries with the transferee ending as `['coordinator', 'case_owner']`.

sei-ahouseholder (commented):
## Verify — PR #3013

**Overall verdict: READY-TO-MERGE**

```text
Merge state:  MERGEABLE (CLEAN) — base main
CI status:    passing — 22 checks (20 pass, 2 expected skips)
Findings:     8/8 CONFIRMED at HEAD (3 FAIL, 5 IMPROVE)
Artifacts:    cleaned up
```

### Merge-state gate

Live re-check (not the artifact value): `MERGEABLE` / `CLEAN`. `main` advanced 10 commits since execute's last sync and the branch is *not* flagged `BEHIND`, so the repo does not require up-to-date branches here. Execute resolved three files during its sync — `vultron/demo/AGENTS.md`, `notes/ownership-transfer.md`, `notes/event-driven-control-flow.md` — and a marker scan at HEAD is clean, so nothing survived the resolution.

### Per-finding verdicts

| # | Finding | Verdict | Evidence at HEAD |
|---|---|---|---|
| 1 | `_get_id` reprs a non-URI `attributedTo` (FAIL) | ✅ CONFIRMED | dict and list branches present, no `str(field)` fallback remains |
| 2 | Unvalidated relay of peer `attributed_to` (FAIL) | ✅ CONFIRMED | `_resolve_offering_actor_id` present; guard `request.actor_id == case_actor_id` and the refusal WARNING both live; call site wired |
| 3 | Stale self-delivery pitfall (FAIL) | ✅ CONFIRMED | `no self-delivery exception` present once; both the original stale heading and my inverted replacement absent; the linked `notes/ownership-transfer.md` section exists |
| 4 | fccv sibling missing broadcast check (IMPROVE) | ✅ CONFIRMED | check present in fccv, `finder_client` threaded, ratchet file present and passing over both scenarios |
| 5 | One-shot `attributed_to` read (IMPROVE) | ✅ CONFIRMED | `wait_for_case_attributed_to` used in the accept path; reject path labelled `EDF-06-006` |
| 6 | Notes frontmatter (IMPROVE) | ✅ CONFIRMED | both specs added; reverse links present in `case-ledger-authority.md` and `event-driven-control-flow.md`; unioned cleanly with `main`'s `demo-scenario-authoring.md` |
| 7 | `case_actor_participant_id_in` untested (IMPROVE) | ✅ CONFIRMED | `TestCaseActorParticipantIdIn` present; 30 tests in the file pass |
| 8 | Extractor test placement (IMPROVE) | ✅ CONFIRMED | both new wire-layer tests present and passing |

Every fix was checked against the file at HEAD, not just its commit diff, and the new tests were re-run at HEAD rather than trusted from execute's log.

### Notes for the reviewer

- **Finding 3 was resolved by `main`, not by this PR.** A concurrent docs restructure deleted the stale rule with a better treatment than the inverted section written here, which was dropped in favour of it. Recording that explicitly because "CONFIRMED" would otherwise imply this PR authored the fix.
- **The 3 FAIL findings were all introduced or exposed by this PR's own changes**, and two of them by the `attributed_to` fix specifically — worth a reviewer's eye on `_resolve_offering_actor_id`'s guard condition, since it is the piece that decides whether a peer can name someone else as the offerer.
- **The two refuted triage findings remain refuted.** Nothing in execute changed the basis for that: `RehydrateActivityNode` expands bare reference URIs from the receiving actor's own store, which is why the bare-`target` failure reproduces only outside the inbox pipeline.

Issue #3019, opened 2026-09-02
concern: BT execution cannot distinguish a programming error from a protocol failure


sei-ahouseholder:
## Concern

`BTBridge.execute_tree` (`vultron/core/behaviors/bridge.py:469`) wraps the whole
tick loop in `except Exception` and returns
`BTExecutionResult(status=Status.FAILURE, ...)`. `BTExecutionResult` carries only
`status`, `feedback_message`, and `errors`, so a `TypeError`/`KeyError`/
`AttributeError` from broken code is **indistinguishable** from a deliberate
protocol failure. Callers then make protocol decisions from a bare `FAILURE`.

The traceback *is* logged — the bridge calls `logger.exception()` at ERROR before
returning — so this is a lost-classification problem, not a silent-swallow
problem.

## Why the obvious fixes do not work

**Deleting the catch breaks working behavior.** 28 deliberate
`raise VultronError(...)` / `raise VultronCanonicalEntryError(...)` calls inside
node code (`sync/nodes/chain.py`, `sync/nodes/canonical_entry.py`,
`sync/nodes/replay.py`) rely on this catch to become a tree `FAILURE`. It is the
intended domain-raise-to-FAILURE mechanism.

**Letting it propagate only moves the problem.**
`vultron/core/use_cases/received/sync.py:133` already wraps `_run_announce_bt` in
its own `except Exception` and re-buffers the ledger entry. Propagating from the
bridge lands the exception one frame up in a handler with the same blindness.

## Concrete harm

Two sites make a decision from an unclassified `FAILURE`:

- **`vultron/core/use_cases/received/sync.py:288`** — logs `FAILURE` at **DEBUG**.
  At the default INFO level a genuine bug leaves only the bridge's own ERROR
  line, without the caller's context.
- **`vultron/core/use_cases/received/sync.py:147`** — re-buffers the ledger entry
  for retry when the BT returns `FAILURE`. If the cause is a code bug, this is a
  retry of something that can never succeed: a programming error becomes a
  permanent redelivery loop instead of a loud stop.

No caller branches on `errors`; the two that read it
(`received/case/validate.py:130`, `received/report.py:181`) only log it.

## Structural gap underneath

The documented convention is "`update()` is the sole `try/except` in BT nodes"
(root `AGENTS.md`). That covers `update()` only — `setup()` and `initialise()`
have **no** node-level protection, which is how a port `data_type` `TypeError`
[...]

Issue #2907, opened 2026-08-31
improve: harden blackboard participant_case port declarations from data_type=object to VulnerabilityCase


sei-ahouseholder:
## Summary

Several BT nodes in `vultron/core/behaviors/case/nodes/participant/` read a `VulnerabilityCase` from the py_trees blackboard via `_try_get_input("participant_case")`, where the port is declared `data_type=object`. After #2490 replaced `isinstance` guards with `cast()` for type narrowing, the runtime contract relies on the blackboard writer always storing a `VulnerabilityCase` — which is guaranteed by BT composition but not enforced by py_trees at runtime.

## Affected nodes

- `CaseHasActiveEmbargoNode` (participant_add.py:337)
- `CaseHasNoActiveEmbargoNode` (participant_add.py:379)
- `PersistOwnerCaseNode` (owner.py:343)
- `RecordParticipantAddedEventNode` (participant_add.py:296)

## Suggested Fix

Change `data_type=object` to `data_type=VulnerabilityCase` in the port declarations for these nodes. This documents the contract and may enable future framework-level enforcement.

## Source

Found by pre-PR code review on #2490.

sei-ahouseholder:
Claimed on branch `task/2907-harden-participant-case-port-types`.

sei-ahouseholder:
**History: implementation — Harden blackboard port declarations to concrete domain types**


## Issue #2907 — improve: harden blackboard participant_case port declarations from data_type=object to VulnerabilityCase

PR: https://github.com/CERTCC/Vultron/pull/3018 (also closes #3011)

Tightened every `participant_case`, `log_entry`, and `replay_entry` py_trees
typed-Port declaration from `data_type=object` to the concrete domain class its
writer guarantees.

**Why it mattered**: py_trees checks `data_type` on both sides —
`_set_output()` and `get_input()` each raise `TypeError` — so the declaration
*is* the enforcement, and `data_type=object` switches it off. That became
load-bearing when #2490 replaced the `isinstance(x, VulnerabilityCase)` guards
with `cast()`: the port check is now the only thing between a wrong-typed
blackboard value and a dereference deep inside `update()`.

**Scope delivered**:

- `owner.py` — the two `participant_case` readers #2909 missed
  (`AdvanceOwnerRmToAcceptedNode`, `RecordOwnerJoinedEventNode`), completing the
  nine-declaration roster.
[...]

Issue #2944, opened 2026-08-31
Retarget the ARCH-22 ratchet at the declared exemption set; retire the xfail


sei-ahouseholder:
## Summary

Retarget the ARCH-22 ratchet at the declared exemption set and retire the xfail,
completing #2670 and closing #2673.

The goal test currently asserts that `vultron/wire/` reaches **zero**
`vultron.core.models` imports. That was unreachable: ARCH-12-001 mandates the
shared-base inheritance, ARCH-12-010 the core type-map fallback, and ARCH-12-005
(before ADR-0081) the projection methods. ADR-0081 removed two of the three
structural causes; one justified exemption remains.

## Acceptance Criteria

- [ ] AC-1: `KNOWN_VIOLATIONS` contains exactly the declared exemption set — after
      the preceding tasks, the single `find_in_core_type_map` import in
      `vultron/wire/as2/vocab/base/registry.py`, mandated by ARCH-12-010.
- [ ] AC-2: Each exemption is annotated inline with the requirement ID that mandates
      it, so a future reader does not mistake it for unfinished work.
- [ ] AC-3: `test_wire_core_model_import_boundary_goal` asserts **equality with the
      exemption set** rather than emptiness, and its `xfail(strict=True)` marker is
      deleted (ARCH-22-003 as amended).
- [ ] AC-4: The xfail reason-string requirement from #2673 is satisfied — either the
      marker is gone or it carries a `Tracked by #N` reference.
- [ ] AC-5: `#2670` and `#2673` are closed by the PR that lands this.
- [ ] AC-6: Full unit suite passes.

## Prior Art

- `test/architecture/test_wire_no_core_model_imports.py` — the ratchet, its
  two-sided tests, and the goal test to retarget.
- `test/architecture/test_normalize_wire_to_core_ratchet.py` — a worked example of
  an exemption set annotated with the issue that owns each entry (deleted by the
  `extra="forbid"` task, but useful as a model first).

## Reference

Source: #2830 (G02). ADR: `docs/adr/0081-wire-core-boundary-pairing-registry.md`.
Docs PR: https://github.com/CERTCC/Vultron/pull/2931
Notes: `notes/wire-core-boundary.md`.
Specs: ARCH-22-001, ARCH-22-002, ARCH-22-003, ARCH-12-010. Closes #2670 and #2673.

sei-ahouseholder:
[...]

Issue #2885, opened 2026-08-31
Replace RequireCaseOwnerApprovalNode with a conversation-state routing gate


sei-ahouseholder:
## Summary

`RequireCaseOwnerApprovalNode` (`vultron/core/behaviors/call_out/nodes.py:98`) returns
`Status.FAILURE` unconditionally with `success_rate = 0.0`. It is **deleted, not
completed**: an Evaluator asked "is this approved?" at the moment of asking can only
ever answer *no*, because no answer exists yet. That is the wrong shape, not an
unfinished implementation — see the ADR-0076 amendment.

Each gate becomes a routing subtree whose branches are disjoint (ASK-02-001,
ASK-02-002, RSH-07-004):

```text
GatedAction  (first fitting branch wins)
├── reply in hand      → Accept: perform the action.  Reject: refuse, record, stop.
├── asked, waiting     → stop.  Do not ask again.
├── asked, expired     → re-ask with a fresh deadline, stop.
└── never asked        → record the not-yet state, emit Offer(Proposal), stop.
```

Disjoint branches are what make re-entry safe. Note why the tempting shortcut fails:
replaying the original message so the same tree runs again dies at the first
idempotency guard, because CLP-13-001 requires a guard that detects a duplicate to
return `FAILURE` and write nothing — aborting before the gate is reached.

`add_participant_status_tree` already leaves a legitimate not-yet state: the
participant's claim is appended by `AppendParticipantStatusNode` and only *adoption as
canonical* is gated (RSH-01), which satisfies ASK-02-005 without new state.

**Do not** relax the gate to a permissive backend to make a blocked path proceed
(RSH-07-005). `STATUS_AUTHORIZATION_PERMISSIVE` is for trusted-participant and demo
deployments; using it to route around an unreachable gate converts a protocol
guarantee back into a configuration posture, which is the defect CONCERN-2092 filed.

Reconciles #2812 (design unresolved) and #2809 (implement the round-trip) — the same
issue filed twice.

## Acceptance Criteria

- [ ] AC-1: `RequireCaseOwnerApprovalNode` is removed; no `CallOutBackendFactory`
      returns an unconditional-`FAILURE` node as a conservative default.
- [ ] AC-2: A reusable gated-action tree factory produces the four-branch routing
      structure above, parameterised by ask kind, subject, and the action subtree.
- [ ] AC-3: `StatusAdoptionGate` and `EmbargoTeardownAuthorizationGate` both use it
[...]

PR #3026, opened 2026-09-02
fix(bt): separate a programming error from a protocol failure in BT execution

### LINKED ISSUE #3048: Concern: test_create_case_activity_missing_case_id is duplicated byte-for-byte in two files
## Concern

`test_create_case_activity_missing_case_id` exists **byte-identically** in two
files:

- `test/core/behaviors/report/test_nodes.py:567`
- `test/core/behaviors/report/nodes/test_case_creation.py:116`

Confirmed by diffing the extracted function bodies — they are the same test, not
two tests that happen to share a name.

## Why it matters

Surfaced by PR #3026, which had to apply the same four-line edit twice to keep the
two copies in step. The next such fix will touch one copy and not the other, and
nothing detects the divergence: both files pass, and the stale copy keeps asserting
the old behavior while reading as coverage.

This is the DRY standard (CS-22-001) applied to tests, where the cost is unusual —
a duplicated *test* does not just repeat code, it manufactures false confidence
once the copies drift.

## Why it is not a simple delete

Which copy is canonical is bound up with a layout question. `report/nodes/` already
exists as a subpackage test directory, so per CS-18-004 / NODES-SPLIT-883 the
mirroring rule says node tests belong under `report/nodes/test_<submodule>.py` and
`report/test_nodes.py` should be the residual. But `report/test_nodes.py` is large
and holds much more than this one test, so resolving this properly means deciding
whether that file gets split rather than just picking a copy to delete.

## Suggested resolution

1. Audit `test/core/behaviors/report/test_nodes.py` against
   `test/core/behaviors/report/nodes/` for the full set of duplicated tests — this
   one was found incidentally, so it is unlikely to be alone.
2. Decide whether `report/test_nodes.py` is split to mirror `report/nodes/` or
   retained with an explicit ownership rule.
3. Delete the redundant copies in whichever direction that decision implies.
4. Consider an architecture ratchet for identical test function names across a
   package and its subpackage test dir — `test/architecture/_corpus.py` already
   provides the shared AST cache this would need.

## Source

Found by `/pr-ship` code review on PR #3026 (`Refs #3019`). Pre-existing
duplication, not introduced by that PR.

sei-ahouseholder:
- Refs #3019
- Closes #3048

## Summary

`BTBridge.execute_tree` caught every exception and returned a bare
`Status.FAILURE`, so a `TypeError` from broken code was indistinguishable from a
deliberate protocol failure. This splits the handler and records the distinction
on `BTExecutionResult`, then makes the test harness refuse to accept a crash as
a protocol failure.

## Motivation

Callers decide **retry, re-buffer, and log level** from that status, and
retrying a code bug never converges. Two shapes of harm exist today:
`received/sync.py:147` re-buffers a ledger entry for retry on `FAILURE`, and
`received/case_participant.py:79` raises a retryable error that lands the
activity back on the inbox queue — forever, if the cause is a bug.

Two obvious fixes do not work, which is why this classifies instead:

- **Deleting the catch breaks working behavior.** 28 nodes deliberately
  `raise VultronError(...)`; that catch is the intended
  domain-raise-to-FAILURE mechanism.
- **Letting it propagate only moves the problem.** `received/sync.py:133`
  already wraps the call in its own `except Exception`, so the exception lands
  one frame up in a handler with the same blindness.

**No production control flow changes here.** Nothing in `vultron/` reads the new
flag yet — this commit adds the signal and the test-side enforcement. Wiring the
callers is step 3, tracked on #3019.

## Changes

- **`vultron/core/behaviors/bridge.py`**: adds
  `BTExecutionResult.internal_error`. `except VultronError` keeps today's
  meaning; `except Exception` still catches — a half-ticked tree must not escape
  into a FastAPI background task — but sets the flag. The two non-convergent
  exits (`INVALID` root, max iterations) are flagged too: a tree that will not
  settle in N ticks will not settle on a retry. Also moves `bt.setup()` inside
  the `try` (a `setup()` crash previously escaped uncaught *and* skipped
  `shutdown()`) and guards `bt.shutdown()` so a raise there cannot replace the
  classified result.
- **`test/core/behaviors/bt_harness.py`**: `assert_failure` rejects an internal
  error unless a test passes `allow_internal=True`. This is the piece that stops
  the class of bug hiding behind a green suite. The docstring states where the
  guard is blind rather than implying it is exhaustive.
- **`test/core/behaviors/test_bridge.py`**, **`test_bt_harness.py`**: 16 new
  tests covering every exit's classification, the setup/shutdown edges, and the
  guard's own behavior.

### The guard found eight tests asserting FAILURE on a crash

Four deliberately test the missing-required-port path and now say so with
`allow_internal=True`. Four named a reason they never reached:

| Test | Named reason | Actual cause | Resolution |
|---|---|---|---|
| `BroadcastCaseUpdateNode::test_failure_when_case_absent` | case absent | missing `excluded_actor_ids` port | supply the port; now fails on the absent case |
| `SendTerminateEmbargoActivityNode` factory test | factory unavailable | missing `embargo_id`, then absent embargo | renamed `test_failure_when_embargo_absent` |
| `QueueToOutboxNode::test_failure_when_datalayer_absent` | datalayer absent | missing `activity_ids` port | renamed `test_failure_when_required_ports_absent` |
| `LogDeliveryConfirmationNode::test_failure_when_datalayer_absent` | datalayer absent | missing `activity` port | renamed `test_failure_when_required_ports_absent` |

The two `datalayer_absent` names described an **unreachable** condition:
`BTBridge.setup_tree` unconditionally assigns `blackboard.datalayer`, so
`_require_datalayer()` cannot be reached through the harness. The embargo test
hit the same trap with `trigger_activity`, which `BTTestScenario` always wires —
so **BT-14-001's factory-unavailable branch still has no coverage**, noted in the
test docstring and on #3019.

## Deliberately not in scope

Recorded on #3019 with file:line evidence: the caller sites that should consult
the flag (`received/case_participant.py:79`, `triggers/_base.py:104`,
`triggers/_helpers.py:162`, `received/sync.py:147,288,364`); propagating the flag
through the nine nested-`BTBridge` nodes that currently discard it; and the
classification edge cases (`VultronActivityConstructionError` is a
`VultronError` but wraps a factory misuse; `VultronProtocolViolationError` is
classified differently depending on whether pydantic wrapped it).

## Verification

- 8184 unit tests pass (22 new, 20 duplicates deleted), 1254 integration tests pass
- Black, flake8, mypy, pyright, markdownlint clean
- Re-validated after merging current `main`
- Every `xfail` in the suite resolves to an open issue
- New tests confirmed load-bearing. Reverting the flag assignment alone breaks
  `test_programming_error_is_flagged`, `test_exception_type_is_named_in_the_message`,
  `test_max_iterations_is_an_internal_error`, `test_invalid_status_is_an_internal_error`,
  `test_setup_time_crash_is_caught_and_flagged`,
  `test_shutdown_crash_does_not_mask_the_result`,
  `test_setup_tree_crash_is_classified_not_raised`,
  `test_rejects_an_internal_error_by_default` and
  `test_the_guard_message_names_the_cause`.
  `TestAssertFailureRejectsCrashes` is **not** uniformly load-bearing on that
  mutation — its opt-in and non-FAILURE cases pass either way by design, which an
  earlier revision of this section overstated.

## Review round (`/pr-ship`)

Triage found 13 items; 11 fixed here, 2 filed. The pattern was this PR's own
thesis turned back on it — a FAILURE assertion that cannot name its reason is not
an assertion, and six opt-out sites had no reason assertion at all.

- **`assert_failure` gains `reason=`, and `allow_internal=True` is accepted only
  together with it.** Switching the crash guard off left those tests asserting
  nothing but "the tick did not hang", with the real expectation in a comment.
  The rule immediately caught the last assertion-free opt-in in the suite.
- **All six `allow_internal=True` sites now name their port**, verified against
  the actual `NoDataAvailable` messages; the case-absent and embargo tests pin
  their reasons; `test_max_iterations_*` and `test_invalid_status_*` no longer
  have byte-identical bodies.
- **`execute_with_setup` called `setup_tree()` outside every `try`**, so a wiring
  error escaped `BTBridge` entirely — no `BTExecutionResult`, and the exception
  reached the FastAPI background task. Now classified like any other internal
  error, so `BTBridge` never raises.
- **Four BT subclasses hoisted to module level** (py_trees keys its class registry
  by name — CONCERN-2321).
- **Production-side write-up added to `notes/bt-pitfalls.md`**, next to the
  nested-bridge idiom that defeats the flag: the prescribed `raise RuntimeError`
  is caught by the calling node's own `except Exception`, so `internal_error`
  stays `False`. That also corrects the #3019 record — `case/nodes/actor.py` is
  **not** the exception; the gap is uniform across all nine sites.
- **Filed**: #3048 (a test duplicated byte-for-byte in two files). Two further
  `VultronError`-laundering sites in `adapters/driven/db_record.py` were appended
  to #3019's classification-edge-case record.
- **Also repaired**: `test_incoming_learnings.py::test_every_committed_learning_file_is_archivable`
  was already failing on `origin/main` (four learning files with naive
  timestamps, BW-02-004). Unrelated to this branch, mechanical, so fixed rather
  than deferred.

### Follow-up round: #3048 fixed here rather than deferred

Triage filed the duplicated-test finding as a follow-up; on review it was fixed
in place instead, and the investigation turned out to matter.

`test/core/behaviors/report/test_nodes.py` is the flat file left behind by the
`report/nodes/` split. **20** of its 37 tests also existed under
`report/nodes/`, and **16 of those 20 had already drifted** — the risk the issue
was filed to track had happened. The drift is one-directional: the `nodes/`
copies gained `@pytest.mark.spec` markers the flat copies never had (BT-03-001,
BT-03-004, BT-10-001, RMB-09-001, RMB-15-001) plus tightened docstrings, so the
flat copies are stale ancestors.

Verified per test by AST-comparing every assertion and call before deleting: no
flat copy asserted anything its counterpart does not. The flat copies did carry
better assertion *messages*, which the split had dropped — those are restored in
`nodes/test_case_creation.py` first. `8184 = 8204 - 20` with no other failure is
the proof they were redundant rather than load-bearing.

Also caught by the new guard: `main` added
`test_junk_on_key_fails_tree_with_type_error` in two typed-port files while this
branch was open. Both assert a wrong-typed port value fails the tree — a
programming error, and the #2907 case that raised #3019 — so both now declare
`allow_internal=True, reason="not of type"` rather than asserting bare FAILURE.

Left deliberately unchanged: both exception branches still log via
`logger.exception`. #3019 records why — log levels touch the SL-04 conventions and
deserve their own decision.


sei-ahouseholder (commented):
## PR Triage: #3026 — fix(bt): separate a programming error from a protocol failure in BT execution

**Linked issues**: #3019 (concern: BT execution cannot distinguish a programming error from a protocol failure) — `Refs`, not `Closes`, which is correct: step 3 stays open by design
**Changed files**: 13 files — `core/behaviors`, `test/core/behaviors`, `notes`
**CI status**: ✅ passing (21/21 checks)
**Base branch**: `main`
**Needs integration tests**: yes (`vultron/core/behaviors/` changed) — PR body reports 1248 integration tests passing

---

### Findings

|---|---|---|---|---|
| phase7-bt-subclasses-inside-test-methods-0 | agents-md-compliance | ❌ FAIL | `_Broken`/`_SilentBug`/`_BadSetup`/`_BadShutdown` defined **inside** test methods — violates the MUST in `notes/testing-pitfalls.md` (CONCERN-2321) | fix-now |
| phase8-setup-tree-escapes-unclassified-0 | code-review | ⚠️ IMPROVE | `execute_with_setup` still calls `setup_tree()` outside both `try` blocks (`bridge.py:597`), so a wiring `TypeError` escapes `BTBridge` unclassified — the same hole this PR just closed for `bt.setup()` | fix-now |
| phase10-allow-internal-sites-assert-no-reason-0 | test-coverage | ⚠️ IMPROVE | All five `allow_internal=True` sites assert no reason; the guard they disable was their last automatic check | fix-now |
| phase10-case-absent-test-reason-unpinned-0 | test-coverage | ⚠️ IMPROVE | `test_failure_when_case_absent` now reaches the lookup but still asserts only `assert_failure(result)`; the node has a second FAILURE path | fix-now |
| phase10-embargo-test-substring-too-weak-0 | test-coverage | ⚠️ IMPROVE | `assert \"not found\"` also matches a case-absent failure the node currently tolerates; assert `EmbargoEvent` or the embargo id | fix-now |
| phase10-identical-assertions-two-branches-0 | test-coverage | ⚠️ IMPROVE | `test_max_iterations_*` and `test_invalid_status_*` have byte-identical bodies — neither pins its branch | fix-now |
| phase8-assert-failure-reason-unused-0 | code-review | ⚠️ IMPROVE | `assert_failure_reason` is the remedy the new note prescribes, yet it takes the *tree* (cannot see `internal_error`) and has **zero** call sites | fix-now |
| phase3-verification-claim-overstated-0 | pr-body-format | ⚠️ IMPROVE | Verification claims reverting the flag breaks 4 of 5 `TestAssertFailureRejectsCrashes`; only 2 of 5 actually break | fix-now |
| phase9-bt-pitfalls-note-not-updated-0 | notes-currency | ⚠️ IMPROVE | `notes/bt-pitfalls.md` § "Always Check `execute_with_setup` Return Value" not updated — and its own recommended `raise RuntimeError` idiom is what suppresses the new flag | fix-now |
| phase5-execute-tree-docstring-omits-classification-0 | spec-conformance | ⚠️ IMPROVE | `execute_tree`'s `Returns:` still lists only status/feedback/errors | fix-now |
| phase2-issue-3019-actor-py-claim-wrong-0 | issue-linkage | ⚠️ IMPROVE | #3019's deferral record says `actor.py`'s `RuntimeError` *is* flagged by the outer bridge; `update()`'s own handler swallows it first | fix-now |
| phase8-db-record-validation-error-misclassified-0 | code-review | 🎫 NEW-ISSUE | `db_record.py:302,316` raise `VultronValidationError` for genuine code bugs → classified `internal_error=False`; a site not yet named on #3019 | new-issue-no-ask |
| phase8-duplicate-test-two-files-0 | code-review | 🎫 NEW-ISSUE | `test_create_case_activity_missing_case_id` is byte-identical in two files, forcing the same edit twice | new-issue-no-ask |

**Total**: 1 FAIL · 10 IMPROVE · 2 NEW-ISSUE

---

### The through-line

Nine of these findings are the PR's own thesis turned back on the PR. It argues that a
FAILURE assertion which cannot name its reason is not an assertion — then ships five
`allow_internal=True` sites with no reason assertion at all, two new tests with identical
bodies for different branches, a prescribed remedy (`assert_failure_reason`) with zero
call sites, and a Verification bullet claiming more load-bearing coverage than it has.
The classification change itself is sound; the finding list is almost entirely about
holding the new tests to the standard the new note sets.

The one FAIL is unrelated and mechanical: four BT subclasses in the wrong scope, in the
same file where two others are correctly at module level.

### Verified clean

Spec conformance (BT-01, BT-05, EH-01–06 — no requirement describes `BTExecutionResult`,
and all BT entries are `kind: project`, so no `@pytest.mark.spec` obligation); ADR
necessity (additive field, no layer/API/schema change — ADR-0008 and ADR-0044 not
contradicted); exception-classification ordering (`except VultronError` before
`except Exception` is correct for the dual-inheritance classes; `NoDataAvailable` is
correctly internal; `vultron/bt/base/errors.py` subclasses would be misclassified but are
unreachable from `vultron/core/`); all six `BTExecutionResult(...)` sites in `vultron/`
flagged correctly, including the leadership guard left unflagged; notes frontmatter;
markdownlint; commit trailers; closing-reference placement; DEFER gating — #3019 carries a
detailed file:line record of every deferred item.

### Candidates deliberately excluded

- **`logger.exception` on both branches.** A routine `VultronNotFoundError` still emits an
  ERROR traceback, and the bridge does not act on its own new classification. Excluded as a
  properly gated deferral: #3019 records it with a stated reason ("changing log levels
  touches the SL-04 conventions and deserves its own decision"). Not re-litigated here.
- **`bt.shutdown()` failure swallowed on the SUCCESS path.** Unreachable — nothing in
  `vultron/` overrides `shutdown()` and the py_trees base is a no-op.
- **`bridge.py` at 673 lines** (CS-18-001 is SHOULD, one responsibility, no line-cap ratchet);
  **`testing-pitfalls.md:287` at 118 chars** (table row, explicitly excepted);
  **`Refs` instead of `Closes`** (correct — the Concern stays open).

---

*Triage artifact: `.claude/pr-3026-triage.json`*

sei-ahouseholder (commented):
## PR Execute: #3026 — fix(bt): separate a programming error from a protocol failure in BT execution

**Fixes applied**: 2 commits (11 of 13 findings)
**Issues filed**: 1 new (#3048) · 1 appended to #3019
**Tests run**: unit + integration — 8134 unit (22 new), 1254 integration
**CI status**: ✅ passing (21/21, green on iteration 1)
**Base sync**: ✅ merged `main` @ `ea384ca1b` — 0 conflicts (80 files from base)

---

### Fixed

| Finding | Commit |
|---|---|
| phase7-bt-subclasses-inside-test-methods-0: hoist 4 BT subclasses to module level | `2dd5dbb` |
| phase8-setup-tree-escapes-unclassified-0: `setup_tree()` was outside every `try` | `2dd5dbb` |
| phase10-allow-internal-sites-assert-no-reason-0: `allow_internal=True` now requires `reason=` | `2dd5dbb` |
| phase10-case-absent-test-reason-unpinned-0: pin the absent case | `2dd5dbb` |
| phase10-embargo-test-substring-too-weak-0: assert `EmbargoEvent`, not bare `not found` | `2dd5dbb` |
| phase10-identical-assertions-two-branches-0: pin each internal-error branch | `2dd5dbb` |
| phase8-assert-failure-reason-unused-0: note prescribes the form that exists and is used | `2dd5dbb` |
| phase9-bt-pitfalls-note-not-updated-0: production-side write-up + AGENTS.md cell | `2dd5dbb` |
| phase5-execute-tree-docstring-omits-classification-0: both bridge docstrings | `2dd5dbb` |
| phase3-verification-claim-overstated-0: PR body Verification corrected | *(PR body)* |
| phase2-issue-3019-actor-py-claim-wrong-0: correction posted on #3019 | *(issue comment)* |

### Filed as Issues

| Finding | Issue |
|---|---|
| phase8-duplicate-test-two-files-0: `test_create_case_activity_missing_case_id` duplicated byte-for-byte | [#3048](https://github.com/CERTCC/Vultron/issues/3048) |
| phase8-db-record-validation-error-misclassified-0: two more `VultronError`-laundering sites | appended to [#3019](https://github.com/CERTCC/Vultron/issues/3019#issuecomment-5513243262) — that issue already owns the edge-case list, so no duplicate was filed |

---

### The one structural fix worth reading

`execute_with_setup` called `setup_tree()` **outside every `try`**. So the one
call that does store cloning, port construction and `setattr` over arbitrary
`context_data` was the single fallible step still outside the net this PR built:
a wiring bug produced no `BTExecutionResult` at all and reached the FastAPI
background task as a bare exception — the exact invariant the new
`except Exception` comment asserts. Now classified like any other internal error,
so **`BTBridge` never raises**, which step 3 on #3019 can now assume.

### The harness change is a rule, not advice

`allow_internal=True` switches off the only automatic check these tests have, so
it is now accepted *only* together with `reason=`. That turns "name your reason"
from a note into something the harness enforces — and it immediately caught the
last assertion-free opt-in in the suite, `test_accepts_an_internal_error_when_opted_in`,
which was itself introduced by this PR.

### Also repaired (not a triage finding)

`test_incoming_learnings.py::TestRealCorpus::test_every_committed_learning_file_is_archivable`
was **already red on `origin/main`** — four `20260901-2789-*.md` learning files
carry a bare `timestamp: 2026-09-01`, which BW-02-004 rejects as naive. Proven
pre-existing (fails at `origin/main` in a detached worktree; validator, test and
all four files entered via merge `ea384ca1b` and are absent from the pre-merge tip;
nothing on this branch touches `plan/incoming/learnings/`). Mechanical, so fixed
in `ae57cf6` rather than deferred.

### Left deliberately unchanged

Both exception branches still log via `logger.exception`, so a routine
`VultronNotFoundError` protocol outcome still emits an ERROR traceback. #3019
records why — log levels touch the SL-04 conventions and deserve their own
decision. Triage flagged it and did not re-litigate it.

---

*Execute artifact: `.claude/pr-3026-execute.json`*
*Next step: `/pr-ship` continues to `/pr-verify` automatically.*

sei-ahouseholder (commented):
## PR Verify: #3026 — fix(bt): separate a programming error from a protocol failure in BT execution

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, not a draft, re-checked live at verdict time
**CI status**: ✅ passing — 21/21 checks, including all four Demo Integration and Invariant Harness jobs
**Integrity**: 13 triage findings / 13 execute results, IDs match exactly
**Conflict markers at HEAD**: clean

---

### Per-finding verdicts

| Finding | Outcome | Verdict | Checked at HEAD |
|---|---|---|---|
| phase7-bt-subclasses-inside-test-methods-0 | fixed | ✅ CONFIRMED | AST walk of `test_bridge.py`: `_BrokenNode`, `_SilentBugNode`, `_BadSetupNode`, `_BadShutdownNode` all module-level |
| phase8-setup-tree-escapes-unclassified-0 | fixed | ✅ CONFIRMED | AST: `setup_tree()` is inside a `try` in `execute_with_setup`, handlers `['VultronError', 'Exception']` |
| phase10-allow-internal-sites-assert-no-reason-0 | fixed | ✅ CONFIRMED | All 7 production `allow_internal=True` sites carry `reason=`; the harness now *enforces* it, so a bare one fails the suite |
| phase10-case-absent-test-reason-unpinned-0 | fixed | ✅ CONFIRMED | `reason=f"case '{CASE_ID}' not found"` |
| phase10-embargo-test-substring-too-weak-0 | fixed | ✅ CONFIRMED | `reason="EmbargoEvent"` |
| phase10-identical-assertions-two-branches-0 | fixed | ✅ CONFIRMED | `"exceeded max iterations (5)"` and `"entered INVALID state"` |
| phase8-assert-failure-reason-unused-0 | fixed | ✅ CONFIRMED | Note prescribes `assert_failure(result, reason=…)`; `assert_failure_reason` docstring states it cannot see status or classification |
| phase3-verification-claim-overstated-0 | fixed | ✅ CONFIRMED | PR body now says the class is **not** uniformly load-bearing; counts refreshed |
| phase9-bt-pitfalls-note-not-updated-0 | fixed | ✅ CONFIRMED | `notes/bt-pitfalls.md:445`, cross-linked; `AGENTS.md` and `test/AGENTS.md` cells extended |
| phase5-execute-tree-docstring-omits-classification-0 | fixed | ✅ CONFIRMED | Both `Returns:` sections enumerate the classification |
| phase2-issue-3019-actor-py-claim-wrong-0 | fixed | ✅ CONFIRMED | Correction comment present on #3019 |
| phase8-db-record-validation-error-misclassified-0 | filed | 📝 NOTED | #3019 OPEN — appended to its edge-case record rather than duplicated |
| phase8-duplicate-test-two-files-0 | filed | 📝 NOTED | #3048 OPEN, milestone Project Health, `size:S`, on Project #24 |

**13/13 verified.** 11 CONFIRMED at HEAD · 2 NOTED with live open issues · 0 UNRESOLVED · 0 MISSING-COMMIT.

---

### Claim checks

The PR body's counts were checked against the authoritative CI run rather than
taken on trust. `Tests (pytest)` reports **9388 passed, 390 skipped, 29 xfailed,
5557 subtests** — which reconciles exactly with the claimed 8134 unit + 1254
integration, and 27 + 2 xfailed. Every `xfail` reason resolves to an open issue
(#2593, #2824, #2885, #2670, #2944, #2505, #1993).

The pre-existing `main` breakage execute repaired
(`test_every_committed_learning_file_is_archivable`) is green here, and the four
learning files now carry quoted tz-aware timestamps per BW-02-004.

### Noted, not blocking

- `test_bridge.py:515` still defines `SilentFail` inside a test function — the
  pre-existing instance of the same CONCERN-2321 violation, which triage
  deliberately scoped out of this PR. Worth a follow-up sweep of the file, not a
  gap in this fix.
- Both bridge exception branches still log via `logger.exception`, so a routine
  `VultronNotFoundError` emits an ERROR traceback. Deferred by explicit decision
  recorded on #3019 (log levels touch the SL-04 conventions). Not a defect in
  this PR.

---

*Artifacts cleaned up. Pipeline complete.*

sei-ahouseholder (commented):
## PR Verify (follow-up round) — ✅ READY-TO-MERGE

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, re-checked live
**CI status**: ✅ 21/21 passing — `Tests (pytest)` reports **9438 passed**, which reconciles exactly with 8184 unit + 1254 integration
**Conflict markers**: clean
**Duplicate test names across `report/test_nodes.py` and `report/nodes/`**: **0** (was 20)

#3048 was fixed in place rather than deferred, and the PR now carries `Closes #3048`.

### What the investigation changed

Filing it would have been the wrong call, and not for scheduling reasons. The
finding was reported as one duplicated test with a *future* divergence risk. It
was **20** duplicated tests, and **16 had already diverged** — the `nodes/` copies
had accumulated `@pytest.mark.spec` markers (BT-03-001, BT-03-004, BT-10-001,
RMB-09-001, RMB-15-001) that the flat copies never got. So the flat file was
shadowing the spec-marked copies with unmarked stale ancestors, which is a
coverage-accounting problem (SR-05-005), not just untidiness.

Deleting was verified safe per test by AST-comparing every assertion and call, not
by eyeballing. The one apparent coverage loss — `test_create_case_activity`'s
`finder.id_ in activity_data['to']` — turned out to be a local rename to
`reporter`. The flat copies' better assertion *messages* were restored in
`nodes/test_case_creation.py` before deletion, so nothing of value went with them.
`8184 = 8204 - 20` with no other failure is the proof they were redundant.

### The new guard earned its keep again

While this branch was open, `main` added
`test_junk_on_key_fails_tree_with_type_error` in two typed-port files, both
asserting a bare FAILURE on a wrong-typed port value. The guard rejected them —
correctly, since a wrong-typed port value is a programming error and is precisely
the #2907 case that raised #3019. Both now declare
`allow_internal=True, reason="not of type"`.

This is the expected cost of the rule: concurrent branches writing crash-path
tests will hit it. The failure is loud and the assertion message names the fix, so
the trade looks right — it is the same check that found the eight originally-wrong
tests.

Issue #2961, opened 2026-09-01
Concern: SYNC-00-007/008 proactive gap detection absent — per-peer replication state exists but is never swept


sei-ahouseholder:
## Summary

`VultronReplicationState` records `last_acknowledged_hash` and `updated_at` per peer, keyed by `{case_id}/replication/{peer_id}`. The CaseActor knows its own ledger tail hash at all times. The comparison needed to detect a silently-lagging replica is therefore trivially available: if a peer's `last_acknowledged_hash != current_tail_hash`, that peer is behind.

What is absent is anything that performs this check proactively. `replay_missing_entries_trigger()` (`vultron/core/use_cases/triggers/sync.py`) can push the missing suffix to a named peer, but it is only ever invoked reactively from `RejectLedgerEntryReceivedUseCase` — i.e. only when a peer sends a `Reject(CaseLedgerEntry)`. If a replayed `Announce(CaseLedgerEntry)` batch is lost in transit, the peer stops sending Rejects (it received nothing new to reject), the CaseActor has no signal to retry, and the replica stays permanently diverged.

SYNC-00-003 requires a "full synchronisation loop, retry, and backoff". SYNC-00-007 requires gap detection. SYNC-00-008 requires retrying until the peer's tail hash converges. The reactive model satisfies these under normal packet loss but not under persistent delivery failure where Rejects also stop arriving.

## Design decision needed

Two viable approaches:

**Option A — opportunistic on-commit sweep**: After `FanOutLogEntryNode` commits and fans out entry N, iterate all `VultronReplicationState` objects for the case and call `replay_missing_entries_trigger()` for any peer whose `last_acknowledged_hash != entry_N.entry_hash`. Rate-limit via the existing `replay_guard` cooldown. Simple, no new infrastructure.

**Option B — Sentinel-shaped watcher**: A persistent background Sentinel that periodically reads all per-peer replication states for active cases, compares each peer's `last_acknowledged_hash` to the current ledger tail, and replays to lagging peers. More powerful (catches gaps even when no new entries are being committed), but requires a scheduler or heartbeat mechanism — Vultron has no current scheduler.

The existing `updated_at` field on `VultronReplicationState` already provides the staleness signal needed by either option.

## Acceptance Criteria

- [ ] AC-1: A design decision is recorded — Option A (opportunistic) or Option B (Sentinel) — in a spec entry or ADR amendment covering SYNC-00-003/007/008.
[...]

Issue #2687, opened 2026-08-26
Participant lifecycle, roster integrity & message delivery


sei-ahouseholder:
## Design idea

Specify and implement reliable participant admission, departure, and rejoin mechanics alongside correct error surfaces for message delivery failures.

## Open work

- #1918 — Rejoin semantics for departed case participants
- #2255 — Received-side BT failures cannot reach InboxOutcome; senders always get 202/processed
- #2257 — Failed Add/Remove(CaseParticipant) receipt silently diverges the case roster and drops the participant from all fan-out
- #2087 — CV (vendor aware) has no self-report path — design intends sender-asserted vendor awareness, implementation has none
- #2215 — Pre-submission embargo negotiation is documented as permitted but has no implemented mechanics
- #2367 — CP-05-005: no recovery path when Accept(CaseProposal) is sent but silently lost
- #2514 — Remove seed-offer-record demo scaffold (superseded by ledger backfill)
- #2679 — impl(2334): runtime enforcement of CaseLedger timestamp invariants + conformance tests

## Rationale

Calved 2026-08-26 from **#1935 Protocol authority & lifecycle semantics** to isolate the participant-lifecycle and message-delivery reliability design thread.

sei-ahouseholder:
## Completed — remaining open work calved into focused epics

Substantial work was delivered under this epic. The 25 open issues that remained have been re-homed across five thematically consistent epics (all at Schedule=Now) to allow each design area to progress independently:

| New Epic | Design idea | Issues |
|----------|-------------|--------|
| #3408 — Embargo negotiation, defaults & lifecycle | Complete embargo negotiation flow — pre-submission proposals, actor vs. protocol defaults, RSVP deadline clamping, shortest-wins at case creation | #2215 #2666 #2667 #3390 #3391 #3392 #3404 |
| #3409 — Participant admission, roster integrity & addressing | Correct participant admission, roster consistency, and activity routing mechanics | #2257 #2624 #2714 #2996 #3047 #3059 #3363 #3365 |
| #3410 — Inbox delivery error surfaces & retry safety | Inbox message processing surfaces BT failures correctly and eliminates unbounded retry loops | #2255 #2865 #2901 |
| #3411 — BTBridge execution hardening | Consolidate exception handlers, separate setup/execution error labels, fix context key leakage | #3084 #3085 #3161 |
[...]

Issue #2685, opened 2026-08-26
Authority model & conservative gate defaults


sei-ahouseholder:
## Design idea

Define who may trigger case-state changes and harden all authorization gate nodes to conservative, require-owner-approval defaults.

## Open work

- #2675 — Flip StatusAuthorizationCallOutBundle defaults to RequireCaseOwnerApproval and add RequireCaseOwnerApproval tree factory
- #2676 — Audit all call-out gates for security-significant permissive defaults and flip them
- #2042 — append_rm_state lacks per-session call-count guard; future participant-creation nodes can silently over-advance RM state
- #1752 — No mechanism for participant to request additional roles after joining
- #1894 — Post-owner-close message boundary is unspecified: is close a kill switch, a selective gate, or unconstrained?
- #1902 — report-phase closure (SvcCloseCaseUseCase) is unreachable — needs the non-auto-create-case branch designed

## Rationale

Calved 2026-08-26 from **#1935 Protocol authority & lifecycle semantics** to isolate the authority-model and gate-defaults design thread.

Issue #2567, opened 2026-08-25
Protocol vocabulary extension (Part 2: speculative additions)


sei-ahouseholder:
## Design idea

Vocabulary and schema additions that require explicit scope decisions before specification can begin: error signaling, invitation lifecycle extensions, actor trust metadata, and protocol vocabulary for new CVD workflows.

## Rationale

Split from **#2088 Protocol vocabulary extension (Part 1)** to separate speculative future vocabulary ideas from concrete near-term implementation work. Each item here needs a design or scope decision before decomposition is possible.

PR #3056, opened 2026-09-02
docs: plan issue #2833 — G05 authority and roster lifecycle

### LINKED ISSUE #1752: Concern: no mechanism for participant to request additional roles after joining
## Observation

During the ISSUE-1745 bugfix (suggest-actor role assignment), it was noted that
the current protocol has no mechanism for a participant to request additional
CVD roles after they have already joined a case.

## Current behaviour

Roles are fixed at the time the `Offer(CaseParticipant)` is created by the
CaseActor and accepted by the Case Owner. The acceptor (participant) can only
accept or implicitly decline offered roles by rejecting the offer entirely — they
cannot add roles not present in the original Offer.

## Risk / fragility

A participant who initially joins as `VENDOR` but later needs `COORDINATOR`
access has no protocol path to request that upgrade. Operators must manually
re-run the suggest-actor flow with the new roles. This is undocumented and
error-prone.

## Possible approach

A new activity type (e.g. `Request(CaseParticipant)` or `Update(CaseParticipant)`)
that allows a participant to request additional roles; the Case Owner
Accept/Reject pattern from ADR-0026 could be reused.

## References

- ISSUE-1745 — suggest-actor roles fix (surfaced this gap)
- ADR-0026 — CaseActor-Routed Actor Suggestion and Invitation Flow
- CM-16 — suggest-actor protocol spec

### LINKED ISSUE #1894: Post-owner-close message boundary is unspecified: is close a kill switch, a selective gate, or unconstrained?
## Summary

The spec has no rule governing whether **any** further case-scoped message may be sent or accepted after the case owner closes the case (the `Leave(VulnerabilityCase)` / RC that broadcasts a terminal close to the ledger). Owner-close is defined as *global and terminal*, but the protocol never states whether the terminal close-ledger announcement is a **kill switch** (no further messages accepted at all), a **selective gate** (some activity types — e.g. `Add(Note)` — remain permitted while others do not), or simply an unconstrained state where messages may still arrive and be handled ad hoc. The concrete case that surfaced this: can a finder send `Add(Note, target=Case)` after the owner has closed the case? The specs neither permit nor forbid it directly — the answer must be inferred, and the inference is incomplete.

## Surface Symptom vs. Underlying Problem

**Surface symptom:** There is no explicit `MUST NOT` telling a finder they cannot send `Add(Note, target=Case)` to a closed case, so it reads like an isolated missing rule for one activity type.

**Underlying problem:** The protocol has no general model for the post-close message boundary. Owner-close is specified as terminally closing the case *for all participants* (a "permanent Leave"), but nothing defines what that terminal state means for **inbound activities**. Is the close-ledger entry a hard boundary after which the CaseActor rejects everything, or does terminality only constrain *RM protocol (R\*) messages* while leaving `MsgTypeGI` activities (notes, general inquiries) and other activity classes unspecified? Without a single stated rule, three implementations could each behave differently and all claim spec conformance.

## Category

Protocol correctness / spec completeness.

## Severity

Medium. Not a live crash, but a genuine conformance ambiguity: independent implementations can diverge on whether post-close messages are dropped, accepted, or partially accepted, and the reference implementation currently has no closed-case guard on the note handler at all.

## Evidence (spec investigation)

**1. Owner closure is global and terminal**
- `ontology/vultron_activitystreams.ttl`, `RmCloseCase` (~lines 787–801): *"Case closure is considered a permanent Leave from the case. If the case owner (attributedTo) is the actor, then the case is closed for all participants. If the actor is not the case owner, then the actor should be removed from the case participants."*
- `specs/case-management.yaml` `CM-04-007` (MUST): `RM.CLOSED` is terminal; repeated `CLOSED -> CLOSED` is a terminal no-op that does not append status records, broadcast, or create ledger entries.
- `notes/case-state-model.md`: RM state is participant-specific, but owner-close is the global-close branch; "Terminal = RM.CLOSED (regardless of path)".

**2. The only post-close message rule is permissive and scoped to R\* only**
- `specs/vultron-protocol-spec.yaml` `VP-03-013` (**MAY**): "Recipients MAY ignore messages received on Closed cases."
- `specs/rm-behavior.yaml` `RMB-01-005`, `RMB-14-002` (MAY): a Participant in RM Closed "MAY ignore any R* message without sending a response... The protocol permits silent discard." Both `satisfies VP-03-013`.
- **Scope gap:** these are explicitly about **R\* (RM) messages**. `Add(Note)` maps to **MsgTypeGI** (General Inquiry), not R\* — so even the permissive "MAY ignore" rules do not literally cover it. There is *no* rule at all addressing a `MsgTypeGI` activity received on a terminally-closed case.

**3. `Add(Note)` is a participant activity any participant may originate — no owner restriction, no close guard**
- `ontology/vultron_activitystreams.ttl` `AddNoteToCase` (~197–213): subclass of `CaseParticipantActivity` and `MsgTypeGI` — contrast owner-only `AddStatusToCase` / `AddParticipantToCase` / `RemoveCaseParticipantFromCase` ("should only be performed by the case owner").
- `specs/vultron-as2-mapping.yaml` (MUST): `ADD_NOTE_TO_CASE` MUST be `Add(Note)[target=VulnerabilityCase]`; `CLOSE_CASE` MUST be `Leave(VulnerabilityCase)`.
- `specs/handler-protocol.yaml`: the `AddNoteToCase` handler MUST persist a Note and append its ref to `VulnerabilityCase.notes` — **no closed-case / RM-status guard**. The note-idempotency rule checks presence in `case.notes`, not case status. This absence is itself evidence the model does not gate `Add(Note)` on closure.

**4. Communication flow routes everything through the CaseActor, which has general reject authority**
- `notes/case-communication-model.md` (PCR-08): after case creation, all case-scoped participant messages flow through the CaseActor (`Participant -> CaseActor -> CaseLedgerEntry -> Announce -> all Participants`); the originator addresses `Add(Note)` only to the CaseActor.
- `notes/case-ledger-authority.md` (CLP-09): the CaseActor is the sole authority deciding whether an inbound assertion becomes canonical history; rejected assertions stay local and are reported to the sender, not broadcast. `Add(Note)` is a listed canonical `payloadSnapshot` type, conditioned on CaseActor acceptance. This is the mechanism by which a post-close note *could* be dropped — but it is a general authorization model, not a close-specific rule.

**Bottom line of the investigation:** No explicit `MUST NOT` forbids sending `Add(Note)` after close; owner-close is terminal-and-global; the only post-close rules are MAY-ignore rules scoped to R\* messages, leaving GI (and other non-R\* activity classes) unspecified. So the finder is not *protocol-violating* by sending, and the CaseActor *may* silently reject — but "may reject" is not "must reject," and nothing distinguishes notes from other post-close activity.

## Impact if Ignored

- Implementations diverge on post-close behavior (drop-all vs. accept-all vs. selective) while all claiming conformance.
- The reference implementation accepts and persists post-close notes today (no handler guard), which may contradict the "terminal / permanent Leave" intent — canonical history could grow after a case is closed.
- The R\*-only scoping of the MAY-ignore rules is a silent coverage hole for the entire `MsgTypeGI` (and other non-R\*) activity families.

## Suggested Action

Decide and specify the post-owner-close message boundary as a general rule, then propagate:

1. Define what the close-ledger announcement means for **inbound activities** — pick a model: (a) hard kill switch (CaseActor MUST reject all case-scoped activities after owner-close), (b) selective gate (enumerate which activity classes remain permitted, e.g. read-only inquiries or notes, and which are rejected), or (c) explicit "MAY ignore" generalized beyond R\* to all message types.
2. If any activities remain permitted post-close, state the CaseActor's obligation (accept/broadcast vs. reject-and-notify) per class rather than leaving it to CLP-09's general discretion.
3. Broaden `VP-03-013` (and the RM-behavior satisfiers) beyond R\* messages, or add a companion rule covering `MsgTypeGI` / non-R\* activities.
4. Add the corresponding closed-case guard (or explicit no-guard rationale) to `specs/handler-protocol.yaml` for `AddNoteToCase` and analogous handlers.

### LINKED ISSUE #1902: concern: report-phase closure (SvcCloseCaseUseCase) is unreachable — needs the non-auto-create-case branch designed
## Concern

`SvcCloseCaseUseCase` (aliased as `SvcCloseReportUseCase`, in
`vultron/core/use_cases/triggers/report.py:178`) and its service method
`TriggerService.close_case(actor_id, offer_id, note)`
(`vultron/core/use_cases/triggers/service.py:199`) close a case *via the report
RM lifecycle*: they resolve an offer → report → case and run
`create_close_case_trigger_tree`.

Nothing in the current process reaches that path. Because the process runs with
"auto create case on report received" set, a case exists before the receiver
evaluates the report as valid or invalid. So there is no point at which we close
a *report* — we always close a *case*, and #1858 made that the uniform path via
`Leave(VulnerabilityCase)`.

**This is not a "delete the dead function" task.** The path is unreachable only
because we have not built the other branch: the case where the receiver does
*not* auto-create a case on report receipt, and we must therefore capture the
receiver's *intent* to create a case from a report. That branch is a genuine
protocol design question, and `SvcCloseCaseUseCase` is what a report-phase
closure would look like if it existed. Removing it would discard the shape of a
path we may need.

## What needs deciding

- Do we support a non-auto-create mode at all, where a received report sits in
  the report phase (RM.RECEIVED → VALID/INVALID) before any case exists?
- If so, how is the receiver's intent to create a case from a report
  represented and triggered? Is that an explicit trigger use case, a
  configuration-driven branch inside the report-received path, or something
  else?
- In that mode, what closes a report that never becomes a case — the existing
  report-RM closure (`SvcCloseCaseUseCase` / `TransitionRMtoClosed`), or
  something new?
- Consequently: does `SvcCloseCaseUseCase` stay as the report-phase closure for
  that branch, get relocated/renamed to make its scope explicit, or get retired?
- What does the `SvcCloseReportUseCase` alias mean once this is settled? It
  currently points at a use case named for cases, which is confusing in both
  readings.

## Why it matters

While unresolved, the codebase carries two closure mechanisms with overlapping
names and no documented boundary:

- `TriggerService.close_case(actor_id, offer_id, note)` — report-lifecycle
  closure, currently unreachable.
- `TriggerService.leave_case(actor_id, case_id)` — the uniform
  `Leave(VulnerabilityCase)` path added in #1858, which is what the demo and
  every scenario now use.

A contributor reading `close_case` reasonably assumes it is the way to close a
case. It is not. Either the boundary gets documented or the API gets
consolidated, and which one depends on the auto-create question above.

## Notes

- The auto-create flag behavior is the premise here; confirm its current
  default and whether a non-auto-create mode is actually in scope before
  designing.
- Related: #1858 (uniform sender-side Leave), and the receiver-side role
  semantics follow-up filed alongside this issue.

### LINKED ISSUE #1918: Rejoin semantics for departed case participants
## Summary

When a participant sends `Leave(VulnerabilityCase)`, their RM state advances to
`RM.CLOSED` (ADR-0050, CM-23-003). Currently `RM.CLOSED` is a terminal state.
This idea explores what should happen if a departed participant wants to rejoin
a case.

## Motivation

During #1910 planning, it was noted that there is currently no mechanism for a
participant to rejoin a case after leaving. This may be a valid use case (e.g., a
vendor who initially declined coordination later wants to participate), but the
current RM state machine has no transition out of `RM.CLOSED`.

## Design Questions

- Should `RM.CLOSED` ever be non-terminal? One framing: `RM.DEFERRED` is a
  "shallow sleep" (easily resumed); `RM.CLOSED` could be a "deep sleep"
  (resumption is more like a resurrection than a reawakening — same participant,
  new lifecycle arc).
- What message type triggers a rejoin request? A new `Join(VulnerabilityCase)`?
  A re-invitation via the existing actor-suggestion flow?
- Does the Case Actor need to explicitly re-accept a rejoining participant?
- How does the ledger record a rejoin? New `CaseLedgerEntry` type, or reuse of
  existing types?
- Are there security/trust implications of allowing a departed participant to rejoin?

## Notes

- Deferred from #1910 planning — out of scope for the CM-23 Leave-closure work.
- If rejoining is supported, `RM.CLOSED` may need to become non-terminal (or a
  new terminal state `RM.TERMINATED` may be needed to distinguish "left and
  never returning" from "left but potentially returnable").
- The `RM.DEFERRED` vs `RM.CLOSED` distinction may inform the design.

### LINKED ISSUE #2087: Concern: CV (vendor aware) has no self-report path — design intends sender-asserted vendor awareness, implementation has none
## Concern

The `CV` (Vendor Aware) message and the `v→V` VFD transition have **no trigger-side self-report path in the implementation**, and the design intent behind them is not captured in any spec. This is a gap between design and implementation.

### What the design intended

Per project author (2026-08-07 review session):

> Any participant who delivers a report to a vendor, or invites a vendor to a case, MAY choose to add a participant status on their behalf indicating that they are aware of the case. In our initial formulation this doesn't require vendor acknowledgement — just that the sender asserts: *A is a vendor, and I gave them an opportunity to know about the report/case.*

So `CV` is a **third-party assertion about another participant's awareness**, not a self-report by the vendor. That makes it structurally unlike the other VFD transitions.

### What exists today

- **Receive-side only**: CSB-01-001 covers receiving `CV` — the receiver updates its model of the *sender's* CS state to `Vfd···` and emits `CK`. Its description ("a report has been delivered to a specific Vendor") is consistent with the third-party-assertion reading.
- **No trigger-side spec**: CSB-15 specifies trigger-side role preconditions for `f→F` (CSB-15-001, requires VENDOR) and `d→D` (CSB-15-002, requires DEPLOYER), plus `notify-published` (CSB-15-003). There is **no CSB-15 item for `v→V`**.
- **No trigger endpoint**: no `notify-vendor-aware` equivalent found under `vultron/core/use_cases/triggers/` or the FastAPI trigger routers. Vendor awareness is currently implicit in report receipt.

Consequence: the draft protocol spec (#2078) lists `vfd → Vfd → VFd → VFD` as VFD states but has no normative statement about who may drive the first transition or how.

### Why this is a Concern and not just a doc gap

Resolving it requires a design decision with contentious consequences.

**Option A — no state changes.** Treat sender assertion as sufficient. A participant who notified a vendor posts a `ParticipantStatus` on the vendor's behalf setting `Vfd`. Simplest, matches original formulation.

**Option B — split notified vs. aware.** Track `notified` (sent) separately from `aware` (positively acknowledged receipt). More honest, but adds a state distinction to VFD and changes the CS model.

**Option C — transport-level evidence.** Use delivery success as the trigger: "I posted the invite to their server and got a 2xx, good enough to call `v→V`."

### The contentious part

Options A and C both apply **behavioral pressure** to the vendor:

> If I made a good-faith effort to notify you, it's not my fault you didn't read your mail — and we know your server received it. So we're going to mark you down as aware.

That is a defensible coordination stance and it is also likely to be disputed by vendors, especially where `v→V` feeds disclosure-timing decisions or public case state. Worth deciding deliberately rather than by implementation default. Flagging now because it will be easier to revisit before there are conformance claims resting on it.

### Open questions

1. May *any* participant assert `v→V` for a vendor, or only the one who did the notifying?
2. Does the asserted-about vendor have a mechanism to dispute or correct it?
3. Does a 2xx on delivery constitute awareness (Option C), and if so, what about shared-inbox or store-and-forward paths where 2xx means "queued", not "seen"?
4. Does `v→V` interact with the two-seam adoption model (RSH-01/RSH-02, ADR-0046)? A third-party assertion about another participant looks like exactly the case Seam 1's approval gate exists for — but the default is auto-adopt.
5. If Option B, does the CS model gain a state, or does `notified` live outside VFD?

### Source

Surfaced during critical review of the draft Vultron Protocol specification (PR #2078) while verifying VFD role-gating claims. Original review item was a suspected code bug; investigation showed no code bug — the transition has no implementation at all, by omission rather than defect.

Relevant: `specs/cs-behavior.yaml` CSB-01, CSB-15; `vultron/core/states/cs.py` (`VFD_Trigger.V = "vendor_becomes_aware"`, `CS_vfd`); ADR-0046; `notes/received-status-authorization.md`.

Adjacent issues: #1667 (CaseStatus authority model undesigned) — related but concerns unilateral state changes generally, not third-party assertion specifically.

### LINKED ISSUE #2534: CreateAndPersistCaseActivityNode: create_case_addressees port call lacks NoDataAvailable guard
## Problem

`CreateAndPersistCaseActivityNode.initialise()` catches `NoDataAvailable` only for the `create_case_obj` port (test-driven: the fail-fast test asserts `feedback_message` contains `"create_case_obj"`). The `create_case_addressees` port is called without a try/except, so if it is absent the exception propagates unhandled — the BT execution fails with an exception rather than returning `Status.FAILURE` with a diagnostic `feedback_message`.

This is inconsistent with the ARCH-15-001 pattern (all required-port reads either set `feedback_message` on missing data or are guarded in `update()`).

## Suggested fix

Wrap the `create_case_addressees` port call with a `try/except NoDataAvailable` block that sets `self.feedback_message` and returns `Status.FAILURE`, matching the pattern already used for `create_case_obj`.

*Surfaced during ISSUE-1886.*

### LINKED ISSUE #2794: [Docs/Explanation] Document Case Actor authority, trust bootstrap, and routing model
## Summary

Enrich the public docs/topics/ Explanation quadrant with content explaining
the Case Actor's role as the authoritative case coordinator: how trust is
established with new participants (trust bootstrap), why all messages route
through the Case Actor, and the authority model for case ledger entries.

## Motivation

The CaseActor authority model and routing constraint are critical for external
implementers to understand before implementing any part of the protocol that
involves multi-party coordination. These concepts are currently only documented
in contributor-facing notes.

## Scope

Source notes:
- `notes/case-communication-model.md` — All messages route through the Case Actor;
  routing rationale and implications
- `notes/case-ledger-authority.md` — CaseActor as authority; trusted timestamps;
  assertion recording model
- `notes/case-bootstrap-trust.md` — How trust is established between CaseActor
  and new Participants via the Invite/Accept handshake

These three are tightly coupled (same actor, same authority chain) and may
produce one enriched page under case_model.md or a new dedicated page.

## Acceptance Criteria

- [ ] AC-1: Read all three source notes files; check docs/topics/case_model.md for overlap
- [ ] AC-2: Decide: extend case_model.md / new page / separate sections (document decision)
- [ ] AC-3: Strip Python-specific content (BT node names, class references, code patterns)
- [ ] AC-4: Keep authority model rationale, routing design, and trust bootstrap protocol flow
- [ ] AC-5: Add framing accessible to external implementers building any Vultron-compatible node
- [ ] AC-6: Run /simplified-technical-english on any new or modified page

## Reference

Source: #2526

### LINKED ISSUE #2833: G05: Authority and roster lifecycle
Planning group **G05** of 19. Parent: #2828 — read its body for the planning protocol that applies to every group (umbrella-is-not-a-defect-claim, Tasks parent to the domain epic, nothing closes before it is in `origin/main`, lose nothing).

**Members**: #1752, #1894, #1902, #1918, #2087, #2534, #1673, #2794, #3020, #2961
**Domain epics for Tasks**: #2685 (authority model), #2687 (participant lifecycle & roster), #607 (docs)
**Blocked by**: G01.

## Why these are grouped

One theme: **who may assert what, and what changes after a participant joins or a case
closes.** Members currently split across #2685 and #2687, which is an artificial cut —
they are two halves of one authority model.

Note honestly: this group has two sub-clusters (post-join role changes: #1752, #1918,
#2087; close semantics: #1894, #1902). If the interview shows they do not share a
decision, splitting into two Task sets is fine — but assess that *after* reading all
eight, not before.

## Blocked by G01

The Offer/Accept/Reject authorization round-trip (**#2812/#2809 are owned by G01 — do
not close them here**). This group's authority rules must be expressible with whatever
gate mechanism G01 defines.

## Members

- **#1752 — no mechanism for a participant to request additional roles after joining.**
  Tier `Next`.
- **#1894 — post-owner-close message boundary is unspecified: is close a kill switch, a
  selective gate, or unconstrained?** Tier `Next`.
- **#1902 — report-phase closure (`SvcCloseCaseUseCase`) is unreachable; the
  non-auto-create-case branch needs designing.**
- **#1918 — rejoin semantics for departed case participants.**
- **#2087 — CV (vendor aware) has no self-report path: the design intends
  sender-asserted vendor awareness, the implementation has none.**
- **#2534 — `CreateAndPersistCaseActivityNode`: `create_case_addressees` port call lacks
  a `NoDataAvailable` guard.** Smallest member; a concrete defect the authority model
  explains.
- **#1673 — actor-level production policy configuration for call-out points.** The
  configuration surface through which authority policy is actually expressed.
- **#2794 — [Docs/Explanation] Case Actor authority, trust bootstrap, and routing
  model.** Folded in: it documents exactly what this session decides.
- **#3020 — authority rules written in many places instead of one: no single place to
  read a rule or ask whether an action is authorized for a role.** Maps directly to the
  authority model this session consolidates; the fragmentation it names is what the ADR
  output eliminates.
- **#2961 — SYNC-00-007/008: proactive gap detection absent — per-peer
  `VultronReplicationState` exists but is never swept for lagging replicas.** Weak fit;
  included because authority over which actor initiates replication repair touches the
  authority model. Triage during session: if it does not share the authority-model
  decision, spin it out as a standalone Task under the SYNC epic.

## Desired output

- **One ADR** on the authority model: what a participant may assert about itself versus
  what requires owner approval; whether roles are requestable post-join and how; what
  close means for inbound and outbound messages; rejoin rules.
- Spec amendments in the case-management / participant-lifecycle files.
- `docs/` explanation page for #2794.
- Task set parented to #2685 / #2687.

## Constraints

- ADR-0076's conservative default (deny unless owner-approved) is the baseline. Any
  loosening must be argued explicitly.
- #1752 and #1894 are `Next` tier — if the session's ADR grows, make sure those two get
  Tasks that can proceed independently.

## Acceptance criteria

- [ ] AC-1: An ADR records the authority model: self-assertable vs owner-approved,
      post-join role requests, close semantics, rejoin rules.
- [ ] AC-2: A verdict on whether the two sub-clusters share one decision or need
      splitting, decided after reading all eight members.
- [ ] AC-3: #1752 and #1894 have Tasks that can proceed independently of the full ADR.
- [ ] AC-4: Spec amendments drafted for case-management / participant-lifecycle.
- [ ] AC-5: `docs/` explanation page written for #2794.
- [ ] AC-6: Tasks parented to #2685 / #2687; #2812/#2809 referenced, not closed.
- [ ] AC-7: #2961 triaged — either its replication-gap question shares the authority ADR
      or it is spun out as a standalone Task under the SYNC epic.

### LINKED ISSUE #3020: The rules about who may act on a case are written in many places instead of one
## Summary

Vultron's rules about who may act on a case — which role may report a fix, who may
start an embargo, who may add to the shared case history — are written many times over,
each time next to the code that fetches the data. There is no single place to read a
rule or to ask whether an action is permitted. The case state rules already live in one
place (`vultron/core/states/`); the role and embargo-eligibility rules do not.

## Surface Symptom vs. Underlying Problem

**Surface reading.** Two packages look like they do the same job.
`vultron/core/predicates/` holds one function in 55 lines, and
`vultron/core/states/` holds many. It looks like duplicated structure that
should be merged or deleted.

**Underlying problem.** The two packages are correctly separated, and the split
should be kept. `states/` answers questions about state values and imports only
`vultron.enums`, `vultron.core.states`, and `vultron.types` — it never reads a case
or a participant. `predicates/` answers questions about cases and participants, and
may ask `states/`. That is why `all_participants_rm_closed` does not belong in
`rm.py`.

The real problem is that a whole class of rules is in neither package. Every party
to a case must reach the same answer to "was this action permitted?" — if the vendor
and the coordinator disagree, the shared case history means different things at the
two sites. A rule written in twelve places cannot be shown to give one answer, and a
rule that is private to one service class cannot be asked at all.

**Already correct, leave alone.** The `states/` package and its import rule. The
behavior trees that decide what an actor *wants* to do — whether to accept a report,
whether the fix is ready, whether to join an embargo. Those are each organization's
own choices and must stay where they are (ADR-0025).

## Category

- [ ] Top risk
- [x] Technical debt
- [ ] Security
- [ ] Performance / scaling
- [ ] Fragile / high-churn area
- [ ] Other

## Severity

medium

## Evidence

**The role test is written at least twelve times.**
`vultron/core/models/case_participant.py:353` already has `has_role()`. Almost nothing
calls it. The membership test is written inline instead at:

- `vultron/core/behaviors/case/nodes/vfd_role_guards.py:117`, `:174`, `:328`
- `vultron/core/behaviors/status/nodes/lifecycle.py:125`
- `vultron/core/behaviors/status/nodes/_adjudication.py:82`, `:114`
- `vultron/core/behaviors/report/nodes/develop_fix_conditions.py:89`
- `vultron/core/behaviors/report/assign_cve_id_tree.py:134`
- `vultron/core/models/participant_status.py:189`, `:196`
- `vultron/core/predicates/participants.py:48`
- `vultron/demo/helpers/milestones.py:163`, `:266`, `:432`

**An existing rule function is bypassed.**
`vultron/core/services/embargo_lifecycle.py:54` imports `EM`, `EM_Trigger`, `EMAdapter`,
and `create_em_machine` from `vultron/core/states/em.py`, but not
`is_valid_em_transition` (`em.py:120`). `_drive_em_transition`
(`embargo_lifecycle.py:908`) instead builds a state machine, calls the trigger, and
catches `MachineError` to decide whether the move was allowed. The answer already
exists in the file it imports from.

**One embargo rule is in the right shape but the wrong place.**
`embargo_lifecycle.py:887` `_assert_pxa_embargo_eligible` is a `@staticmethod` taking
`(pxa_state, case_id, operation)`. It reads no stored data, so it could move into the
rule layer with no change in behavior. It is private to a service class, so nothing
else can ask "may this case still start an embargo?" (EMB-01-002, EMB-02-002).

**The write rule for the case history is a table, not a rule.**
`vultron/core/behaviors/sync/nodes/canonical_entry.py:80` `_CASE_AUTHORED_SIGNATURES`
lists the activity and object pairs the case manager may author.
`notes/case-ledger-authority.md:344` already records this as
"Gap 1: Authorization Was a Convention, Not a Gate."

**The role gates are tied to how the code runs, not only to the case.**
`vultron/core/behaviors/case/nodes/role_gates.py` `create_case_manager_gated_tree`
compares the acting actor identifier held in behavior tree working memory against the
case manager named on the case (BTND-07-005). Its own docstring warns that the acting
actor, its store, and the role holder must all be one actor, and that the gate
silently skips the work if any of the three differ.
`vfd_role_guards.py:46` `_resolve_actor_roles` shows the usual shape: fetch the
participant's roles from storage, then apply a one-line test.

**The package meant for this is unused and undocumented.**
`vultron/core/predicates/` is 55 lines and one function,
`all_participants_rm_closed` (`participants.py:29`), written so a demo check could run
against objects in memory. Its only caller is `vultron/demo/helpers/verification.py:27`.
Nothing on the protocol path uses it. The root `AGENTS.md` names `core/states/` at
lines 146, 147, and 671 and never names `core/predicates/`; `vultron/core/AGENTS.md`
names neither. So there is no written rule for where a new rule goes, and in practice
new ones have been going into behavior tree nodes.

**This is a pattern, not one instance.** Four open concerns are the same shape seen one
rule at a time: #3016 (`violation_rm_vfd_entailment` has zero callers), #3008
(`violation_pxa_em_entailment` has no enforcement path), #3009 (the replica-apply path
checks only the RM ratchet), and #2701 (`ValidateCaseStatusTransitionNode` is wired
into no tree). A rule with no single home is a rule that can quietly stop being asked.

## Impact if Ignored

- Two parties to the same case can reach different answers about whether an action was
  permitted, and nothing detects the disagreement.
- Each new rule gets written in a new place, so the count of sites grows and the cost of
  changing a rule grows with it.
- Rules can only be tested through a running container and a database, because they are
  welded to the code that fetches data. That is why the concerns listed above were found
  one at a time by inspection rather than by a test.
- Three findings above are defects in waiting: a bypassed rule function, a rule that
  cannot be reused because it is private, and a gate that skips silently when three
  identifiers drift apart.

## Suggested Action

Decide where the role rules and the embargo eligibility rules belong. Three options:

**A. Move them, and keep two packages.** `states/` keeps rules about state values.
`predicates/` grows to hold rules about cases, participants, and roles. Write the
boundary and the import direction into `vultron/core/AGENTS.md`. The behavior tree nodes
keep the fetching and call the rule.

**B. Move them into `states/` and remove `predicates/`.** One place for every rule. The
cost is that `states/` starts reading domain objects, so its current clean import rule
is lost.

**C. Write down the boundary only.** Accept the repeated rules, but add the rule for
where new ones go so the repetition stops growing. This part is small enough to ride
along with the next change that touches either package, whichever option is chosen.

Weigh this on testability and single-answer grounds. A rule written once can be tested
once, without a container, and can be shown to give every party the same answer. It also
closes the gap the project has already recorded in `notes/case-ledger-authority.md:344`.
This is not preparation for any change of authority model and should not be scheduled as
though it were.

**Out of scope.** What any rule says — this is about where rules live, not what they
permit. The behavior trees that decide what an actor wants to do (ADR-0025).
`_CASE_AUTHORED_SIGNATURES` as a design; finding 4 records it as evidence, and changing
it is its own decision.

## References

- Specs: CLP-07, CLP-09, CLP-12-001, CM-25-005, CSB-15-001, CSB-15-002, RSH-01-002,
  EMB-01-002, EMB-02-002, BTND-07-005
- ADRs: ADR-0025, ADR-0076
- Notes: `notes/case-ledger-authority.md:344`
- Same-shape concerns: #3016, #3008, #3009, #2701
- Nearest sibling in this epic: #2042

sei-ahouseholder:
- Closes #2833
- Closes #2087
- Closes #3020
- Closes #1752
- Closes #1894
- Closes #1902
- Closes #1918
- Closes #2534
- Closes #2794

## Summary

Planning output for G05 (authority and roster lifecycle). Records two ADRs
capturing the authority model decided in the grill-me session and routes the
ten member concerns into implementation issues, Ideas, and triage outcomes.

## Changes

- **`docs/adr/0084-participant-assertion-authority.md`**: participant status is
  self-declaratory by default; `v→V` (Vendor) and `d→D` (Deployer) are
  assertable on-behalf by Case Manager/Owner when externally evidenced; `f→F`
  is Vendor-only; role/roster changes need owner approval (ADR-0026);
  Vendor-participation-implies-`V` invariant (valid Vendor VF ∈ {Vf, VF}).
- **`docs/adr/0085-case-lifecycle-boundaries.md`**: owner-close is a hard write
  boundary (no external ledger writes after the closure `Announce`); `RM.CLOSED`
  is terminal, rejoin unsupported with a defer-not-close workaround; reopen is
  owner-only (mechanics deferred); CONCERN-1902 resolves to a spec clarification
  that auto-create-case is an implementation choice.
- **`docs/adr/index.md`**, **`mkdocs.yml`**: register ADR-0084 and ADR-0085.

## Sub-cluster verdict (umbrella AC-2)

The two flagged sub-clusters resolve into two distinct decisions sharing one
through-line: **ADR-0084** (assertion authority: who may assert what) and
**ADR-0085** (lifecycle boundaries: close & rejoin).

## Implementation Issues

Authority epic (#2685):
- #3057 — On-behalf `v→V`/`d→D` assertion + Vendor-implies-V invariant (from #2087) [size:M]
- #3058 — Centralize enforceable authority/eligibility rules into the rule layer (from #3020) [size:L]

Participant lifecycle epic (#2687):
- #3059 — Note-based post-join role-change workaround (from #1752) [size:S]
- #3060 — Owner-close write boundary + closed-case guard (from #1894) [size:M]
- #3061 — RM.CLOSED terminal + defer-not-close; auto-create-case is impl choice (from #1918, #1902) [size:S]
- #3062 — Bugfix: NoDataAvailable guard on `create_case_addressees` (from #2534) [size:S]

Docs epic (#607):
- #3063 — Case Actor authority/trust-bootstrap Explanation page (from #2794); blocked-by #3057, #3060, #3061 [size:M]

## Deferred to Ideas (Epic #2567)

- #3065 — `Update(CaseParticipant)` post-join role-change protocol message (replaces the #3059 workaround)
- #3066 — Case reopen mechanics (needs the G01 ask primitive #2885/#2883)

## Triaged out

- #2961 (proactive replication gap detection) — does not share the authority decision; remains under the SYNC epic #1158 for its own planning (umbrella AC-7).
- #1673 (production call-out policy config) — out of scope; stays in the Deployable-in-Production track.


sei-ahouseholder:
## 🔎 PR Triage — #3056

**docs: plan issue #2833 — G05 authority and roster lifecycle**

| | |
|---|---|
| Base | `main` |
| Merge state | ✅ MERGEABLE (CLEAN) — verified via live 3-way trial merge |
| CI | ✅ all checks passing (Analyze, CodeQL, docs-build-check, lint) |
| Type | Docs / planning (`specs-notes`) |

### Findings: 1 IMPROVE, 0 FAIL

#### IMPROVE — `phase2-deferred-ideas-not-filed-0` (fix-now)
The PR body states two items are **"Deferred to Ideas (Epic #2567)"** — the
`Update(CaseParticipant)` post-join role-change protocol message and case-reopen
mechanics — but **no Idea issues were actually created**. Epic #2567's sub-issues
do not include them and no open Idea matches. This trips the planning protocol's
explicit *"lose nothing"* rule: the deferred design directions live only as prose
in the PR body and ADR-0085's *Deferred* section.

**Resolution (fix-now, same family):** file the two Idea issues under Epic #2567
(matching the PR's own claim), or reword the body/ADR to say they are still to be
captured.

### What passed
- ✅ **Merge state** — trial merge deduplicates the identical ADR-0083 index/nav lines both sides add; merged `index.md` and `mkdocs.yml` contain ADR-0083/0084/0085 exactly once, no duplicates.
- ✅ **ADR conformance** — 0084 & 0085 follow `_adr-template.md`; `status: accepted` agrees with decisive prose (MS-14). Deferred sub-points are explicitly scoped out, not confidence hedges.
- ✅ **Registration** — both ADRs registered in `docs/adr/index.md` and `mkdocs.yml` nav.
- ✅ **Issue linkage** — all 7 impl issues (#3057–#3063) exist and are OPEN; 8 group members + umbrella #2833 closed; #1673 and #2961 correctly left open (triaged out).
- ✅ **History archive** — `plan/history/2609/learning/CONCERN-2833.md` well-formed, frontmatter valid, covers AC-1…AC-7.
- ✅ **PR body / commit trailers** — closing refs at top, Summary + Changes present; both commits carry `Co-authored-by`.

*Note: Phase 8 code-review sub-agent skipped — diff is 100% Markdown/YAML (ADRs, index, nav, history); no code to review. ADR content reviewed manually instead.*

<sub>🤖 pr-triage — no code changed in this phase.</sub>

sei-ahouseholder:
## 🔧 PR Execute — #3056

**1 finding processed · 1 fixed · 0 skipped · 0 deferred**

### ✅ Fixed — `phase2-deferred-ideas-not-filed-0` (IMPROVE, fix-now) — `45304e718`
The PR promised two items "Deferred to Ideas (Epic #2567)" but no Idea issues
existed. Filed them so the promise (and ADR-0085's "tracked as Ideas under Epic
#2567" statement) is concrete — **lose nothing**:

- **#3065** — `[Idea] Update(CaseParticipant) post-join role-change protocol message`
- **#3066** — `[Idea] Case reopen mechanics for closed vulnerability cases`

Both typed **Idea**, wired as sub-issues of **Epic #2567**, and added to Project
#24 (Schedule = Someday). ADR-0085's *Deferred* reference, the planning history
archive, and this PR's body were updated to cite #3065 / #3066.

### CI loop (1 iteration)
- 🔀 Synced with `main` — merged clean (`b9be9a628`), no conflicts.
- 🧪 Local unit suite — **passing** (expected XFAILs only).
- 📄 `mkdocs build --strict` — the 101 strict-mode warnings are a **known issue tracked separately**, not introduced here; CI `docs-build-check` passes.
- ✅ **CI green** — all 5 checks pass (Analyze ×2, CodeQL, docs-build-check, lint).
- 🔗 Merge state: **MERGEABLE (CLEAN)**, base `main`.

<sub>🤖 pr-execute — handing off to pr-verify.</sub>

sei-ahouseholder:
## ✅ PR Verify — #3056

### Overall verdict: **READY-TO-MERGE**

| Gate | Result |
|---|---|
| Merge state | ✅ **MERGEABLE (CLEAN)** — live check, base `main` |
| Conflict markers at HEAD | ✅ none |
| CI | ✅ all 5 checks pass (Analyze ×2, CodeQL, docs-build-check, lint) |
| Execute integrity | ✅ 1 finding / 1 result — counts match |

### Per-finding verdict

|---|---|---|---|
| `phase2-deferred-ideas-not-filed-0` | IMPROVE | fixed (`45304e718`) | ✅ **CONFIRMED** |

**Spot-check at HEAD:** ADR-0085's *Deferred* line and the planning history
archive both cite **#3065** / **#3066**; both issues are **OPEN**, typed
**Idea**, and wired as sub-issues of **Epic #2567**. The "lose nothing" gap is
closed — the two deferred design directions are now tracked, not just prose.

No deferred-ask items. No FAIL findings. Nothing blocks the merge.

<sub>🤖 pr-verify — read-only. Session artifacts cleaned up.</sub>

Issue #3070, opened 2026-09-02
docs-build-check never runs when only the executable-example source changes (DOCBW-02-001 trigger paths omit vultron/)


sei-ahouseholder:
## Symptoms

A pull request that changes only Python under `vultron/` can break every
executable example on the published docs site without the docs workflow ever
running. Expected: any change that can break a rendered page triggers the docs
build. Observed: it does not.

## Root cause

`.github/workflows/docs-build-check.yml` triggers on `docs/**`,
`overrides/**`, `references.bib`, `mkdocs.yml`, `pyproject.toml`, `uv.lock`,
a few root markdown files, and the workflow file itself. It does **not**
trigger on `vultron/**`.

But ~101 `markdown_exec` blocks under `docs/howto/activitypub/` are of the form:

```python
print(json2md(some_example()))
```

Their content is produced at build time by code in
`vultron/wire/as2/vocab/examples/`. That directory is the actual source of
those pages, and it is outside the trigger set.

`specs/docs-build-workflow.yaml` DOCBW-02-001 encodes the same wrong premise
in its rationale:

> These are the files that can break the site build or introduce broken links.

That statement is false as written. #2904 is the proof: the frozen-model
assignment in `vultron/wire/as2/vocab/examples/_base.py` broke all 101 blocks,
and no listed trigger path was involved.

Note that DOCBW-03-002 already recognises the general shape of this problem
("The Build Site step MUST run on every workflow trigger regardless of which
paths changed") — but a step that always runs is useless if the workflow
itself never fires.

## Components involved

- `.github/workflows/docs-build-check.yml`
- `specs/docs-build-workflow.yaml` — DOCBW-02-001 statement and rationale
- `vultron/wire/as2/vocab/examples/` — the untracked source of ~101 pages

## Relationship to #3051

Distinct defect, same remediation area. #3051 is about the docs build not
running `--strict`, so warnings never fail. This is about the docs build not
*running at all* for the changes most likely to break executable examples.
Fixing only #3051 leaves this hole open; fixing only this leaves warnings
non-fatal. They should probably be resolved together.

## Source

Discovered during analysis of #2904.

sei-ahouseholder:
## Correction: the trigger gap is wider than filed above

I scoped this issue around the ~101 executable example blocks fed by
`vultron/wire/as2/vocab/examples/`. That is the narrowest case, not the whole
one. Two more page families are generated at build time from sources outside
the trigger set:

[...]

Issue #3051, opened 2026-09-02
docs-build-check does not run mkdocs --strict, so link/anchor validation never gates CI


sei-ahouseholder:
## Problem

`mkdocs.yml` declares validation settings that are meant to fail the build —
`validation.nav.omitted_files: warn` (#1304) and, as of PR #3027,
`validation.links.anchors: warn`. Neither can actually fail CI, because no
workflow runs `mkdocs build --strict`:

- `.github/workflows/docs-build-check.yml:52` — `uv run mkdocs build --verbose --clean --config-file mkdocs.yml`
- `.github/workflows/deploy_site.yml:50` — `uv run mkdocs build --clean --config-file mkdocs.yml`

`grep -rn 'strict' .github/workflows/*.yml` returns only an unrelated
`sha256sum --check --strict`. So warnings are printed and the job stays green.

The local wrapper `.github/scripts/mkdocs-build-strict.sh` does pass `--strict`,
but it currently aborts on the 101 pre-existing `markdown_exec` warnings
(#2904) regardless of anything else, and its warning-count arithmetic is broken
in both directions (see `plan/incoming/learnings/20260827-mkdocs-build-strict-arithmetic-bug.md`
and `20260827-mkdocs-build-strict-false-negative.md`). So there is no gate
locally either.

## Why it matters

The `anchors` setting was added in #3027 specifically because MkDocs reports dead
in-page `#fragment` links at `info` level, which is invisible — that PR's first
draft shipped 31 dead in-page links and looked clean. Promoting them to `warn`
makes them visible, which is worth having on its own. But the stated intent, "so
this fails `--strict` in future", does not hold until something runs `--strict`.
The same is true of the nav `omitted_files` guardrail from #1304, which has been
inert for the same reason.

## Acceptance criteria

- [ ] AC-1: A CI job fails when `docs/` introduces a dead in-page anchor link
- [ ] AC-2: A CI job fails when a page is added without a nav entry or `not_in_nav` declaration
- [ ] AC-3: `.github/scripts/mkdocs-build-strict.sh` warning-count arithmetic is fixed so exit code and printed message agree
- [ ] AC-4: The chosen mechanism tolerates a documented baseline of known-bad warnings, or the baseline is zero because #2904 is fixed first

## Blocked by

#2904 — until the 101 `markdown_exec` failures are resolved, a bare
`--strict` gate fails on debt unrelated to any given PR. Either fix #2904 first
or land an explicit baseline mechanism.

## Source

[...]

Issue #1190, opened 2026-06-26
Project Infrastructure Upgrades


sei-ahouseholder:
Tracks improvements to project tooling and automation infrastructure — build
tooling, documentation tooling, CI/CD, and developer workflow automation.

## Scope
- #294 MkDocs → Zensical migration (MkDocs 2.0 compatibility)
- #526 Automated monthly/quarterly project reports (GitHub Action)
- #1862 add push:main trigger + concurrency to demo-integration.yml; remove
  demo-image-cache-warm.yml (routed from #1794 2026-08-03)

Issue #2288, opened 2026-08-13
Remove alias_generator from the six core actor classes; render actor AS2 responses from as_* wire types


sei-ahouseholder:
## Summary

Remove `alias_generator=to_camel` from the six core actor classes and `CoreActorCollection`, and render AS2 actor documents from the `as_*` wire classes instead of from core.

This half of the #1991 set is independent of `ParticipantStatus`/`CaseStatus` (#2289): the actor classes have no flat-field shim, and their only real consumer is the HTTP actor router. The blast radius is externally visible, though — `GET /actors/{id}` currently shapes its AS2 output using the core alias generator — so the route change and the alias removal must land together.

## Acceptance Criteria

- [ ] AC-1: `alias_generator=to_camel` is removed from `VultronPerson`, `VultronOrganization`, `VultronService`, `VultronApplication`, `VultronGroup`, and `CoreActorCollection` in `vultron/core/models/actor.py` (lines 85/115/135/155/175/195 as of the planning baseline). (ARCH-12-003)
- [ ] AC-2: `CoreActor.to_json()` (`vultron/core/models/actor.py:76-77`) is deleted. It dumps `by_alias=True`, has no callers in `vultron/`, and is a standing bypass of the port seam. (ARCH-20-005)
- [ ] AC-3: `GET /actors/{id}` and its sibling routes in `vultron/adapters/driving/fastapi/routers/actors/_routes.py` (lines 108, 166, 176, 212, 481 as of the planning baseline) render responses from the `as_*` actor classes — `as_VultronPerson`, `as_VultronOrganization`, `as_CaseActor` already exist — rather than dumping a core actor class with `by_alias=True`. (ARCH-20-006)
- [ ] AC-4: An HTTP-level test asserts the served actor document still carries camelCase AS2 keys and the correct `@context`, i.e. the external contract is unchanged by the refactor.
- [ ] AC-5: The stale docstring on `VultronPerson` claiming it is 'Registered in `VOCABULARY["Person"]`' is corrected — the six core actor classes are registered in `CORE_VOCABULARY` only. (Verified during planning: their `WIRE_VOCABULARY` entries are empty.)
- [ ] AC-6: `vultron/demo/utils.py` no longer relies on core-side aliasing for its `by_alias=True` dump of core objects.

## Reference

Source: #2260
Docs PR: https://github.com/CERTCC/Vultron/pull/2285
ADR: `docs/adr/0063-wire-rendering-port-for-core-objects.md`
[...]

PR #3071, opened 2026-09-02
fix: clear example timestamps with model_copy so all 101 markdown_exec vocabulary blocks render

### LINKED ISSUE #1328: [Bug] mkdocs build: markdown_exec code blocks fail due to _CASE singleton reuse in examples
## Summary

Three executable Python code blocks in the docs fail during `mkdocs build --strict`,
producing `markdown_exec: Execution of python code block exited with errors` warnings
that fail strict mode.

## Root Cause

`vultron/wire/as2/vocab/examples/_base.py` defines `_CASE` as a module-level
singleton. `create_case()` in `case.py` calls `case()` (which returns `_CASE`)
and then calls `add_participant()` on it. When multiple doc code blocks execute
`create_case()` in the same build process, the second call finds the actor already
mapped to a participant UUID from the first call, triggering:

```
VultronValidationError: Participant-index divergence: actor
'https://vultron.example/organizations/vendorco' already mapped to
'urn:uuid:<first-id>' but add_participant received 'urn:uuid:<second-id>'.
```

The three failing code blocks are at positions n43, n58, and n82 in the build,
all tracing to `vultron/wire/as2/vocab/examples/case.py:58`.

## Acceptance Criteria

- [ ] `mkdocs build --strict` produces zero `markdown_exec` execution errors
- [ ] `create_case()` can be called multiple times in the same process without error
- [ ] Fix uses `case(random_id=True)` or resets state between calls rather than
      reusing the shared singleton

## Fix Approach

The simplest fix is to have `create_case()` call `case(random_id=True)` instead
of `case()` so each execution gets a fresh `VulnerabilityCase` with a new ID and
empty `actor_participant_index`.

### LINKED ISSUE #2904: All 101 markdown_exec vocabulary-example blocks fail: frozen-model assignment in _strip_published_udpated
## Summary

Every `markdown_exec` Python block that renders an AS2 vocabulary example fails at docs
build time. The published pages show a Python traceback where the example JSON should be.
101 blocks are affected.

## Root cause

One line. `vultron/wire/as2/vocab/examples/_base.py:136`, inside
`_strip_published_udpated`:

```python
cast(Any, obj).published = None
```

The target model is frozen — `vultron/wire/as2/vocab/base/objects/base.py:43`:

```python
model_config = ConfigDict(validate_assignment=False, frozen=True)
```

So every call raises:

```text
pydantic_core._pydantic_core.ValidationError: 1 validation error for as_VulnerabilityReport
published
  Instance is frozen [type=frozen_instance, input_value=None, input_type=NoneType]
```

`markdown_exec` catches the exception, emits `WARNING - markdown_exec: Execution of
python code block exited with errors`, and renders the traceback as the block's output.

The `cast(Any, obj)` is doing exactly what a cast does — silencing the type checker
without changing runtime behaviour. `mypy` and `pyright` cannot see the problem because
the cast is the thing hiding it.

## Why CI never caught it

`.github/workflows/docs-build-check.yml` runs:

```bash
uv run mkdocs build --verbose --clean --config-file mkdocs.yml
```

No `--strict`. Warnings therefore never fail the job. With `--strict` the build exits 1:

```text
Aborted with 101 warnings in strict mode!
```

## Suggested fix

Replace the assignment with a copy, since the model is frozen by design:

```python
obj = obj.model_copy(update={"published": None, "updated": None})
```

Check the surrounding function — it appears to mutate and return, so callers may rely on
identity. Drop the `cast(Any, ...)` as part of the fix; if the type checker then objects,
that objection is the real signal.

Once the 101 warnings are gone, consider making `docs-build-check` strict so this class
of regression fails on the PR that introduces it. That is the change with lasting value —
without it, the next silent docs breakage is equally invisible.

## Acceptance criteria

- [ ] AC-1: `PYTHONPATH= uv run mkdocs build --strict` exits 0 with no `markdown_exec`
  warnings.
- [ ] AC-2: A rendered vocabulary example page shows example JSON, not a traceback —
  verified by inspecting built output, not just by the absence of warnings.
- [ ] AC-3: A test covers `_strip_published_udpated` against a frozen model, so the
  regression cannot return silently.
- [ ] AC-4: Decide (and record) whether `docs-build-check` becomes `--strict`. If yes, it
  is a separate commit so the mechanical fix and the policy change are reviewable apart.

## Reference

- Found during `/pr-ship` triage of PR #2880 while verifying that PR's docs claim
- Pre-existing on `main`; unrelated to #2880's contents
- `vultron/wire/as2/vocab/examples/_base.py:136`
- `vultron/wire/as2/vocab/base/objects/base.py:43`
- `.github/workflows/docs-build-check.yml`

sei-ahouseholder:
- Closes #2904

## Summary

Every `markdown_exec` block that renders an AS2 vocabulary example published a
Python traceback instead of the example JSON — 101 blocks. The timestamp-stripping
helper cleared `published`/`updated` by assignment on a wire object, and wire
objects are frozen by design (ADR-0074), so the call raised on every input. The
helper now returns a `model_copy` with those fields cleared instead of mutating.

`mkdocs build --strict` goes from **exit 1 with 101 warnings** to **exit 0 with 0
warnings**.

## Root cause

`vultron/wire/as2/vocab/examples/_base.py:136` did:

```python
cast(Any, obj).published = None
```

`as_Object` sets `ConfigDict(validate_assignment=False, frozen=True)`
(`vocab/base/objects/base.py:43`), so Pydantic raised
`ValidationError(type=frozen_instance)` on every call. `markdown_exec` caught it
and rendered the traceback as the block's output.

The `cast(Any, obj)` is why no type checker ever flagged this: the cast erases the
target type, so neither mypy nor pyright could see the write was forbidden.

`model_copy` is the right fix rather than forcing the mutation through
(`object.__setattr__`, or relaxing `frozen=True`). Beyond respecting ADR-0074:
`_REPORT`, `_CASE` and the actors are shared module-level singletons, and
`adapters/driving/fastapi/routers/examples.py` serves those same instances over
HTTP. Stripping in place would have permanently removed the timestamps from live
API responses as a side effect of a docs build — the shared-singleton hazard that
closed #1328.

## The 101st block was a separate defect

`docs/howto/activitypub/objects.md` called `_case.add_report(...)` and
`_case.add_participant(...)`. Those methods exist only on the **core** case model
(`vultron/core/models/case.py:180,188`) and were never on the wire
`as_VulnerabilityCase`; the block also passed `add_vendor_participant_to_case()`
(an `as_Add` activity) where an `as_CaseParticipant` belongs. It dated to #67,
before the core/wire split.

It was also the only docs block that assembled an object inline across several
statements instead of calling one example function — which is exactly why it rotted
unnoticed. It is now `populated_case()`, so the test suite covers it.

## Why the test suite missed all of this

`test/wire/as2/vocab/test_vocab_utils.py` has always had a `test_json2md`, and it
always passed. Its fixture was `class Foo(as_Base)` — `as_Base` is not frozen and
declares neither `published` nor `updated`, so both `hasattr` guards were False,
the buggy lines never executed, and the test asserted that nothing went wrong.
Covered-but-never-exercised, which suppresses the signal rather than merely
missing it.

## Changes

- **`vultron/wire/as2/vocab/examples/_base.py`**: `_strip_published_udpated`
  returns `obj.model_copy(update=...)` built from `type(obj).model_fields`;
  `cast`/`Any` dropped. Docstring records why it copies rather than mutates.
- **`vultron/wire/as2/vocab/examples/participant.py`**: extracted
  `_participant_for()` — the three `add_*_participant_to_case()` functions each
  rebuilt the same participant inline (CS-22-001). Adds `vendor_participant()`
  and `finder_participant()`; `coordinator_participant()` now composes on the
  helper. Construction is byte-identical, so all existing example output is
  unchanged.
- **`vultron/wire/as2/vocab/examples/case.py`**: adds `populated_case()` — a case
  carrying its report plus finder and vendor participants, with a consistent
  `actor_participant_index`.
- **`vultron/wire/as2/vocab/examples/vocab_examples.py`**: re-exports the three
  new functions.
- **`docs/howto/activitypub/objects.md`**: the `VulnerabilityCase` block is now
  `print(json2md(populated_case()))`, matching every other block on the page.
- **`test/wire/as2/vocab/test_vocab_utils.py`**: frozen-model regression tests
  (stripping works, and the argument is not mutated), plus a parametrized ratchet
  over all 60 zero-arg example functions asserting each renders parseable JSON —
  one test that would have caught all 100 failures without a docs build. Includes
  a guard so the ratchet cannot silently collect zero functions.
- **`notes/wire-artifact-immutability.md`**: corrects the claim that `frozen=True`
  raises `TypeError`; under Pydantic v2 it is a `ValidationError` with
  `type=frozen_instance`.

## Verification

- AC-1: `PYTHONPATH= uv run mkdocs build --strict` → **exit 0, 0 warnings**
  (baseline: exit 1, `Aborted with 101 warnings in strict mode!`).
- AC-2: verified against **built HTML**, not just warning absence — no
  `Traceback`/`frozen_instance` anywhere under `site/howto/activitypub/`, and the
  rendered `VulnerabilityCase`, `Create` and `Invite` examples parse as JSON with
  `published`/`updated` absent at the top level.
- AC-3: covered by the frozen-model tests and the example ratchet above.
- AC-4: recorded below — no CI change here.
- Unit suite: 8262 passed, 390 skipped, 27 xfailed. Integration: 1254 passed,
  2 xfailed. Both re-run after freshening onto `origin/main`.
- Black, flake8, mypy, pyright clean. markdownlint clean.

## AC-4 decision: the `--strict` CI gate stays out of scope

`docs-build-check.yml` still does not pass `--strict`, deliberately. That work is
already tracked as #3051 (parent epic #1190, milestone Project Health), whose AC-3
also covers the `mkdocs-build-strict.sh` warning-count arithmetic bug. #3051 is
wired as `blockedBy` #2904 and its body says to fix #2904 first so a bare
`--strict` gate does not fail on unrelated debt — merging this unblocks it with a
zero baseline. Keeping the policy change separate also matches AC-4's own
"separate commit so the mechanical fix and the policy change are reviewable apart".

## Also filed: #3070

While confirming why CI never caught this: `docs-build-check.yml` triggers on
`docs/**`, `mkdocs.yml`, `pyproject.toml` and friends — but **not** `vultron/**`.
Since ~101 pages are generated by code in `vultron/wire/as2/vocab/examples/`, a PR
touching only that directory never starts the docs workflow at all. This bug is the
proof. `specs/docs-build-workflow.yaml` DOCBW-02-001 encodes the same wrong premise
in its rationale ("These are the files that can break the site build").

Distinct from #3051: that one is about warnings not being fatal, this one is about
the workflow not running. Fixing either alone leaves a hole.


sei-ahouseholder (commented):
## PR Triage: #3071 — fix: clear example timestamps with model_copy so all 101 markdown_exec vocabulary blocks render

**Linked issues**: #2904 (All 101 markdown_exec vocabulary-example blocks fail: frozen-model assignment in _strip_published_udpated)
**Changed files**: 11 files — `wire/as2/vocab/examples`, `docs`, `notes`, `test/wire`
**CI status**: ✅ passing (22 checks green)
**Base branch**: `main`

Verified independently during triage: `PYTHONPATH= uv run mkdocs build --strict` → exit 0, 0 warnings (AC-1); no `Traceback`/`frozen_instance` anywhere under `site/howto/activitypub/` (AC-2); the participant refactor is byte-identical in serialized output vs. `origin/main` for all four affected example functions; the new ratchet reproduces #2904 when run against `origin/main` sources; AC-4's deferral to #3051 is recorded in the PR body.

---

### Findings

|---|---|---|---|---|
| phase8-populated-case-participant-id-mismatch-0 | code-review | ❌ FAIL | `populated_case()` renders a case whose participants and `actorParticipantIndex` reference a *different* case id (`urn:uuid:…` vs the case's `https://demo.vultron.local/…`) — visible on the built page | fix-now |
| phase8-case-kwargs-silently-ignored-0 | code-review | ⚠️ IMPROVE | `case(random_id=False, **kwargs)` silently discards kwargs and returns the `_CASE` singleton | fix-now |
| phase8-stripped-case-genesis-hash-unverifiable-0 | code-review | ⚠️ IMPROVE | Stripping `published` leaves `genesisHash` behind (derived from it, CLP-08-003), so the published example shows a hash nobody can recompute | fix-now |
| phase8-participant-for-type-widened-0 | code-review | ⚠️ IMPROVE | `_participant_for(actor: as_Base)` should be `as_Actor`; the new `as_Base` import only widens the signature | fix-now |
| phase10-ratchet-can-shed-coverage-silently-0 | test-coverage | ⚠️ IMPROVE | The ratchet drops any example that grows a required argument, with no signal | fix-now |
| phase10-not-examples-list-mostly-dead-0 | test-coverage | ⚠️ IMPROVE | 3 of 4 `_NOT_EXAMPLES` entries are already covered by the required-args filter; the comment misdescribes `main` | fix-now |

**Total**: 1 FAIL · 5 IMPROVE · 0 NEW-ISSUE

#### On the FAIL

`_participant_for()` contexts its participants against the singleton `case()`, but `populated_case()` builds its enclosing case with `case(random_id=True, …)`. Rendered output:

```json
"id": "https://demo.vultron.local/VulnerabilityCase/3d016371-…",
"caseParticipants": [{"id": "urn:uuid:6a4aca2e-…/participants/finndervul",
                      "context": "urn:uuid:6a4aca2e-…"}],
"actorParticipantIndex": {"https://vultron.example/users/finndervul":
                          "urn:uuid:6a4aca2e-…/participants/finndervul"}
```

The fix must mint the case id first and thread it into the participant helpers. Dropping `random_id=True` does **not** work — `case()` ignores `**kwargs` when `random_id` is False, so the block would render an empty case instead (hence the second finding).

#### Confirmed correct, for the record

- The `objects.md` rewrite fixes a second, independent defect: the old block called `add_report`/`add_participant`, which exist only on the **core** case model, and mutated the shared `_CASE` singleton — it would have raised `AttributeError` even with #2904 fixed.
- `model_copy(update=...)` is output-identical to a validated construction here: `to_json` uses `exclude_none=True` and not `exclude_unset`, so the added `model_fields_set` entries are invisible.
- No import cycle introduced between `examples/case.py` and `examples/participant.py`.
- The retained `Foo(as_Base)` test is no longer vacuous — it now exercises the `if not updates: return obj` early return.
- Pre-existing and untouched: `choose_preferred_embargo()` does not round-trip through `from_json` (fails on the unstripped object too); nested sub-object timestamps are left in place across all examples, as the new docstring says.

---

*Triage artifact: `.claude/pr-3071-triage.json`*

sei-ahouseholder (commented):
## PR Execute: #3071

All 6 triage findings fixed inline in `2675789d0`. No issues filed, nothing skipped, no user decisions deferred.

| # | Finding | Outcome | Commit |
|---|---|---|---|
| phase8-populated-case-participant-id-mismatch-0 | ❌ FAIL — case id disagreed with its own participants | ✅ fixed | `2675789d0` |
| phase8-case-kwargs-silently-ignored-0 | ⚠️ IMPROVE — `case(**kwargs)` discarded kwargs | ✅ fixed | `2675789d0` |
| phase8-stripped-case-genesis-hash-unverifiable-0 | ⚠️ IMPROVE — `genesisHash` left behind by the strip | ✅ fixed (docstring) | `2675789d0` |
| phase8-participant-for-type-widened-0 | ⚠️ IMPROVE — `as_Base` → `as_Actor` | ✅ fixed | `2675789d0` |
| phase10-ratchet-can-shed-coverage-silently-0 | ⚠️ IMPROVE — ratchet could shed coverage silently | ✅ fixed | `2675789d0` |
| phase10-not-examples-list-mostly-dead-0 | ⚠️ IMPROVE — misleading `_NOT_EXAMPLES` comment | ✅ fixed | `2675789d0` |

### What changed

- **`populated_case()` keeps the shared `case()` id.** It previously minted a fresh one with `random_id=True` while its participants were built against the singleton, so the published block showed a case whose `caseParticipants[].context` and `actorParticipantIndex` pointed at a *different* case — and in a different URI scheme. The rendered block now carries one case id throughout.
- **`case(**kwargs)` honours its kwargs** when `random_id` is False, applying them on top of the shared case identity instead of silently returning the bare singleton. Raising would also have closed the footgun, but honouring is what keeps the participants and their case in agreement.
- **`_strip_published_udpated`'s caveat now covers derived fields** — `as_VulnerabilityCase.genesis_hash` is computed from `published` (CLP-08-003), so a stripped case renders a hash the reader cannot recompute. The behaviour is intentional; the silence about it was not.
- **`_participant_for(actor: as_Actor, ...)`**, dropping the `as_Base` import that only widened the signature past what the helper can handle.
- **The ratchet can no longer shed coverage in silence**: `_candidate_example_funcs()` + `_requires_arguments()` are now separate, and a new test asserts nothing is excluded, so an example that grows a required argument fails loudly.

### Verification

- Sync: merged `origin/main` cleanly (`8587fee72`), no conflicts.
- Unit suite: **8263 passed, 390 skipped, 27 xfailed** (129s). Integration suite not required — no `demo/`, `adapters/`, `behaviors/`, `use_cases/` or extractor changes.
- `PYTHONPATH= uv run mkdocs build --strict` → **exit 0, 0 warnings**, re-run after the fixes.
- Black, flake8, mypy, pyright clean on the touched files.
- xfail ratchet: every `XFAIL` reason resolves to a live open issue (the `#1991` reference in `test_hierarchy_invariants.py` is explicitly annotated as superseded by open #2288/#2289).
- CI after push: **22/22 checks passing**. Merge state: `MERGEABLE` / `CLEAN` against `main`.

---

*Execute artifact: `.claude/pr-3071-execute.json`*
*Next step: `/pr-verify`.*

sei-ahouseholder (commented):
## PR Verify: #3071 — fix: clear example timestamps with model_copy so all 101 markdown_exec vocabulary blocks render

**Overall verdict**: ✅ READY-TO-MERGE
**CI status**: ✅ passing (20 success, 2 skipped, 0 failing)
**Base sync in execute**: ✅ merged @ `8587fee72` (0 conflicts resolved)
**Integrity check**: ✅ all 6 findings accounted for

---

### Finding Verdicts

|---|---|---|---|
| phase8-populated-case-participant-id-mismatch-0 | ❌ FAIL | fixed @ `2675789d0` | ✅ CONFIRMED |
| phase8-case-kwargs-silently-ignored-0 | ⚠️ IMPROVE | fixed @ `2675789d0` | ✅ CONFIRMED |
| phase8-stripped-case-genesis-hash-unverifiable-0 | ⚠️ IMPROVE | fixed @ `2675789d0` | ✅ CONFIRMED |
| phase8-participant-for-type-widened-0 | ⚠️ IMPROVE | fixed @ `2675789d0` | ✅ CONFIRMED |
| phase10-ratchet-can-shed-coverage-silently-0 | ⚠️ IMPROVE | fixed @ `2675789d0` | ✅ CONFIRMED |
| phase10-not-examples-list-mostly-dead-0 | ⚠️ IMPROVE | fixed @ `2675789d0` | ✅ CONFIRMED |

### Evidence at HEAD

The FAIL finding was verified functionally, not just by diff: at HEAD,
`populated_case().id_ == case().id_`, every participant's `context` equals the case id, and every
`actor_participant_index` value is prefixed by it — with the report and both participants still
present. The `case(**kwargs)` path returns a populated case instead of silently discarding its
kwargs. `as_Actor` annotation, the `genesis_hash` docstring caveat and the corrected
`_NOT_EXAMPLES` comment are all present in the working tree, and
`test_no_example_func_is_excluded_from_the_ratchet` passes.

No conflict markers anywhere in the tree. No deferred items, no filed issues, nothing skipped.

---

*Verified against HEAD `8587fee72`. Session artifacts cleaned up.*

Issue #3156, opened 2026-09-03
Guards with no production caller pass their own unit tests: two confirmed witnesses, no structural check


sei-ahouseholder:
## The pattern

A guard whose only production call site never invokes it, while its unit tests
pass by calling it directly. The tests report the requirement as covered; the
code path is dead. This is the *second* confirmed instance, which is why it is
being filed rather than left as a learning entry.

### Witness 1 — `plan/incoming/learnings/20260902-2904-vacuous-test-wrong-base-class.md`

`test_json2md` passed continuously while the function was broken for every real
input, because the fixture subclassed `as_Base` (no `published`/`updated`) and
both `hasattr` guards evaluated False. 101 published pages rendered tracebacks
while the test was green.

### Witness 2 — #2824 (this session)

`_validate_entry_timestamps` (`vultron/core/behaviors/sync/nodes/canonical_entry.py`)
implemented CLP-14-002/003/006/007/008. `_validate_canonical_entry` ran it only
under `if case_published is not None:`, and `CreateLogEntryNode.update()` —
the sole production caller — never passed `case_published`. The guard never
executed in production for its entire life. Meanwhile
`test_canonical_entry.py` had **14** passing tests for it, every one calling
`_validate_entry_timestamps` directly with hand-supplied arguments.

Compounding it: `check_clp14_timestamp_invariants` in
`test/ci/invariants/common.py` was never imported by `universal_harness.py`, so
no scenario ran it either, and ADR-0079 § "Validation" asserted both were
running.

## Why this shape is worse than missing coverage

Missing coverage is visible in a coverage report and reads as a known gap. These
tests *suppress* the signal: the spec-coverage ratchet counted CLP-14-002/003/006
as covered, so nothing anywhere reported a problem. The defect surfaced only
because a stale `xfail` reference sent someone to read the call site.

## Suggested guard

An architecture ratchet in the shape of the existing
`test/architecture/test_no_dl_mutations_in_execute.py`: for each private
validator/guard in `vultron/`, assert at least one non-test caller exists, and
that its parameters that gate behaviour are actually supplied at that call site.
A narrower and cheaper version: for guards that raise a spec-cited error, require
at least one test that reaches the guard **through** its production entry point,
not by direct call. `test/core/behaviors/sync/nodes/test_chain_timestamp_guard.py`
(added in the #2824 fix) is the shape that test should take.

Worth considering as part of the existing test-correctness epic (#2619).

## Provenance

[...]

Issue #2898, opened 2026-08-31
bug(flaky): CaseActor outbox congested by Phase 1 LedgerFanout — demo drain workaround masks protocol-layer race


sei-ahouseholder:
## Summary

`fcvcv_demo`, `fccv_extension_demo`, and `fvcv_extension_demo` intermittently fail because the CaseActor's outbox is congested by Phase 1 `LedgerFanout` when Phase 2 submits `suggest-actor-to-case`. The resulting `Offer(CaseParticipant)` arrives late and the `find_cp_offer_for_case` poll times out.

## Current workaround (commit `dffc78f2`, PR closes #2819)

Added `wait_for_contiguous_ledger_coverage` drain calls at the end of `_phase_report_submission` in all three demo files, plus a timeout increase from 15s → 40s. The commit message labels this "Option A" — a workaround, not a root fix.

## Root cause (unaddressed)

The CaseActor's outbox has no back-pressure or ordering guarantee relative to new submissions arriving while it is processing a fanout. When Phase 1 ledger replication is in flight, Phase 2's `suggest-actor-to-case` submission competes with the fanout for outbox capacity. The drain workaround moves the ordering responsibility into each demo scenario, meaning:

1. Any scenario that doesn't add the drain is still vulnerable.
2. The workaround is load-sensitive — under slower CI it may still time out.
3. The underlying protocol delivery ordering problem (ADR-0026, ADR-0058) is not enforced by the actor layer.

## What a real fix looks like

The CaseActor (or the LedgerFanout node) should ensure outbox drains complete before accepting new case-participant submissions, OR `suggest-actor-to-case` should be gated on ledger sync at the protocol level rather than at each demo scenario's discretion.

## History

This race class has recurred multiple times: #2376 (closed Aug 2026), #2819 (closed Aug 2026), now back again. Each closure added a demo-layer workaround rather than fixing the delivery ordering guarantee.

## References

- `vultron/demo/scenario/fcvcv_demo.py` — `_phase_report_submission` drain block
- `notes/flaky-tests.md` rows for `fcvcv Demo Integration` / `fcvcv Invariant Harness`
- ADR-0026, ADR-0058

sei-ahouseholder:
## Another CI datapoint — `fcvcv`, `_phase_report_submission`, passed on rerun with identical code

Observed on PR #3159 (`bug/2824-clp-14-15-timestamp-invariants`), commit `22973f736`:

[...]

PR #3159, opened 2026-09-03
fix(ledger): wire the CLP-14/15 timestamp guard into the commit boundary

### LINKED ISSUE #2824: xfail cleanup: CLP-14-003/006/009 and CLP-15-001-004 timestamp/causal-order invariants not yet enforced (stale #2679 reference)
## Summary

Seven `@pytest.mark.xfail(strict=True)` tests in `test/core/models/test_clp_14_15_invariants.py` reference issue #2679, which was closed by commit `fcc836ce` (`impl(2514,2679)`). The xfail markers were not removed when #2679 closed, and the referenced invariants are still not enforced at runtime.

## Stale xfail markers

| Test | Spec ID | Description |
|---|---|---|
| `test_clp_14_003_published_timestamps_monotonic` | CLP-14-003 | No cross-entry monotonic timestamp check |
| `test_clp_14_006_entry_not_before_case_creation` | CLP-14-006 | No `entry.published >= case.published` validation |
| `test_clp_14_009_clock_skew_tolerance_configurable` | CLP-14-009 | Configurable clock-skew tolerance not exposed |
| `test_clp_15_001_participant_emits_in_causal_order` | CLP-15-001 | Participant causal-order enforcement not implemented |
| `test_clp_15_002_participant_must_not_batch_in_arbitrary_order` | CLP-15-002 | Out-of-order batch detection not implemented |
| `test_clp_15_003_participant_published_timestamps_nondecreasing` | CLP-15-003 | Participant timestamp monotonicity not enforced |
| `test_clp_15_004_participant_timestamp_reflects_event_time` | CLP-15-004 | Timestamp accuracy SHOULD-check not implemented |

## Action needed

1. Update the `reason=` strings in all 7 markers to reference this issue.
2. Implement the 7 remaining CLP-14/15 timestamp and causal-order invariants to make the tests pass, then remove the markers.

## Evidence

The xfail ratchet in `pr-execute` detected these during PR #2822 (docs-only, no relation to the invariants). Surfaced because the post-sync test run found stale xfail references after the main-branch merge that closed #2679.

### LINKED ISSUE #3043: Concern: CreateLogEntryNode does not pass case_published, so CLP-14 enforcement is opt-in on the chain commit path
`_validate_canonical_entry` accepts `case_published` and enforces CLP-14, but
`CreateLogEntryNode.update()` (`chain.py` ~line 300) does not pass it. Runtime
enforcement therefore fires only when a caller explicitly opts in, leaving
CLP-14-002, -003 and -006 unenforced on the chain commit path — the path that
actually mints ledger entries.

## Acceptance criteria

- [ ] AC-1: `CreateLogEntryNode.update()` passes `case_published` through to
  `_validate_canonical_entry`.
- [ ] AC-2: A test asserts CLP-14-002/-003/-006 are enforced via the chain
  commit path, not only via a direct validator call.

## Provenance

Recovered from `plan/incoming/learnings/20260828-clp14-chain-callsite-unwired.md`
during the 2026-09-02 learnings-queue audit; never filed.

### LINKED ISSUE #3149: AS2 extractor drops the sender's published timestamp, making CLP-14-007/008 vacuous
## Summary

`_build_activity_snapshot()` in `vultron/wire/as2/extractor/_builders.py:135`
builds the core `VultronActivity` from an inbound wire activity but does not
carry the wire activity's `published` field across the boundary. The sender's
claimed "sent at" timestamp is silently discarded, and
`VultronActivity.published` falls back to its
`default_factory=_now_utc` (`vultron/core/models/base.py:210`) — the
*receiving* actor's clock at extraction time.

## Evidence

`_build_activity_snapshot` copies 15 fields (`id_`, `type_`, `actor`,
`attributed_to`, `object_`, `target`, `origin`, `context`, `in_reply_to`,
`to`, `cc`, `summary`, `content`, `suggested_roles`, `roles`) and omits
`published`. The nested-object builder 30 lines below it,
`_build_report_object` (`_builders.py:183`), *does* carry it:
`published=_get_timestamp(obj, "published")`. The top-level activity builder
was simply missed.

## Impact

Two CLP-14 SHOULD-level guards are structurally vacuous as a result:

- **CLP-14-007** — reject a payload assertion whose `payloadSnapshot.published`
  is implausibly far in the *future* relative to the CaseActor's clock.
- **CLP-14-008** — reject one that is implausibly *stale*.

Both are implemented in `_validate_entry_timestamps`
(`vultron/core/behaviors/sync/nodes/canonical_entry.py:218-234`) and compare
`payloadSnapshot.published` against `datetime.now()`. Because that field is
stamped by the receiving actor milliseconds earlier, neither branch can ever
fire on a real inbound message. Their tests
(`test_clp_14_007_future_timestamp_payload_rejected`,
`test_clp_14_008_stale_timestamp_payload_rejected` in
`test/core/models/test_clp_14_15_invariants.py`) pass only because they call
`_validate_canonical_entry` directly with a hand-crafted payload, bypassing
extraction.

More broadly, the CaseActor cannot detect participant clock error or timestamp
spoofing at all — ADR-0079 § "Residual Uncertainty" assumes it can flag
implausible timestamps at SHOULD level.

Note that `_VOLATILE_SNAPSHOT_KEYS` in `vultron/core/sync_helpers.py:181`
correctly documents `published`/`updated` as build-time-stamped; the docstring
on `_validate_entry_timestamps` (`canonical_entry.py:182`) asserts the
opposite — that the field is "the AS2 activity's *claimed* timestamp" — and is
wrong.

## Fix

Carry `published=_get_timestamp(activity, "published")` in
`_build_activity_snapshot`, matching `_build_report_object`, and correct the
`_validate_entry_timestamps` docstring.

Idempotency is unaffected: `_semantic_payload` already strips `published` from
the equivalence comparison via `_VOLATILE_SNAPSHOT_KEYS`.

## Provenance

Found during #2824 (CLP-14/15 timestamp invariant enforcement) while
determining why the CLP-14 commit-boundary guard had never fired.

### LINKED ISSUE #3158: AGENTS.md size ratchet is red on main: 405 lines against a 400-line target
## Summary

`test/metadata/test_agents_md_size_ratchet.py::test_root_agents_md_within_target`
fails on `origin/main`: the root `AGENTS.md` is 405 lines against a 400-line
target. Verified by checking out `origin/main:AGENTS.md` in isolation and running
the ratchet — the failure reproduces with no local changes.

## Impact

`main` is red, and the failure lands on every unrelated PR whose branch is
freshened onto it. `AGENTS.md` is a high-traffic file, so the noise is
continuous, and the specific failure mode teaches agents to ignore a red
ratchet.

## Cause

Growth accumulated across several merges. The rule in `AGENTS.md` itself
("this file is at its 400-line budget, so trim as you add, never append") and in
`notes/agents-md-structure.md` is only advisory at authoring time; the ratchet is
the enforcement, and it went red without anyone owning the trim.

## Fix

Recovered the 5 lines by condensing prose in the "Cross-cutting rules with no
other home" section and the skill-interaction and miscellaneous-tips lists. No
rules were dropped — only wording tightened.

Fixed in the #2824 PR because that PR touches `AGENTS.md` (one line, in the
case-ledger pitfall row) and would otherwise ship red for a failure it did not
cause.

## Follow-up worth considering

The ratchet fires only after the budget is already blown. A pre-commit hook on
`AGENTS.md` line count would catch it at authoring time, where the person adding
the line still has the context to decide what to trim.

### LINKED ISSUE #3169: CLP-14/CLP-15 do not name which of an entry's two timestamps they constrain, and CLP-15-001/002 read as CaseActor duties CLP-15-005 forbids
## Summary

CLP-14 and CLP-15 never said *which* of a `CaseLedgerEntry`'s two timestamps each
requirement constrains, and two CLP-15 entries read as CaseActor enforcement
duties that CLP-15-005 forbids. Both gaps produced working code that was wrong,
and seven `xfail(strict=True)` stubs that could never have passed.

An entry carries two timestamps:

- `CaseLedgerEntry.published` — the CaseActor's **commit** stamp.
- `payloadSnapshot.published` — the asserting actor's **claimed** event time.

## What was wrong

1. **CLP-14-002 / -003 / -006 did not name the field.** Applied to the wrong one,
   each check is either vacuous or falsely rejecting. CLP-14-003 in particular:
   applied to *claimed* timestamps it compares two participants' unsynchronised
   clocks, which is the wall-clock ordering ADR-0079 explicitly rejected as
   option C.

2. **CLP-15-001 / -002 read as CaseActor duties.** They sit in a group titled
   *Participant Assertion Timestamp Obligations*, but neither says who checks
   them, and CLP-15-005 — four entries later — forbids the CaseActor from
   reconstructing participant-internal causal order it cannot verify. Stubs
   written against the enforcement reading were structurally unpassable.

3. **CLP-15-003 was treated as per-assertion enforceable.** The CaseActor
   observes *arrival* order, not causal order. Refusing an assertion on this
   signal drops well-formed messages under ordinary transport reordering
   (ADR-0037), and drops them entirely, because the guarded commit precedes the
   effect nodes (CLP-10-006).

4. **CLP-14-009 authorised only two of three tolerances.** It named CLP-14-007
   and CLP-14-008, so the clock-skew tolerance applied to CLP-14-006 had no
   normative basis.

5. **No requirement covered inbound `published` presence.** AS2 makes the field
   optional and the wire base class defaults it to `now_utc`, so a receiver
   could not distinguish a claimed time from its own clock. That is what left
   CLP-14-007 and CLP-14-008 vacuous (#3149). The commit-boundary guard was
   citing CLP-07-011 for a presence rule CLP-07-011 does not state.

## Resolution

Amended in `specs/case-ledger-processing.yaml`:

- CLP-14-002, -003, -006 now name the envelope stamp vs. the claimed timestamp
  explicitly, with `verification` pointing at the layer that owns each half.
- CLP-14-006 gains the claimed-timestamp clause and its clock-skew tolerance.
- CLP-14-009 covers all three tolerances.
- CLP-15-001, -002 gain a `note` that they bind the participant only and are not
  per-assertion enforceable, plus `verification` pointing at `check_causal_edges`
  (DEMOMA-22-005).
- CLP-15-003 gains a `note` that a CaseActor MAY report a regression at a
  severity tracking the configured tolerance, but MUST NOT refuse on it.
- New **CLP-15-006**: a participant MUST include `published`; a receiver MUST
  reject an inbound activity without one at message validation and MUST NOT
  substitute its own clock.

## Provenance

Surfaced by the PR that wired the CLP-14/15 guard into the commit boundary, and
recorded in `plan/incoming/learnings/20260903-2824-clp14-15-do-not-name-their-timestamp.md`.
Filed rather than left in the learnings queue because a learnings file is an
input queue, not a trackable record.

Related: #2752 (ADR-0079 vs. SYNC-01-002 gapless-index), #3149, #2824.

### LINKED ISSUE #3170: DEFER_CASE omits include_activity=True: a received Ignore(VulnerabilityCase) never ledgers and loses the RM→DEFERRED transition
## Summary

The `DEFER_CASE` `SemanticEntry` in `vultron/semantic_registry/case.py` omits
`include_activity=True`, while its `ENGAGE_CASE` sibling has it. As a result a
received `Ignore(VulnerabilityCase)` never reaches the case ledger, and when the
receiving actor holds `CASE_MANAGER` the RM→DEFERRED transition is lost with it.

## Mechanism

- `include_activity` defaults to `False` (`vultron/semantic_registry/_entry.py`).
- `DeferCaseReceivedEvent` does not narrow the base
  `activity: VultronActivity | None = None`, so it stays `None`.
- `create_receive_activity_tree` is called with a non-`None` `case_id` from
  `vultron/core/behaviors/report/prioritize_tree.py`, so the guarded commit *is*
  in the tree.
- With no activity, the ledger snapshot is dumped from the `VultronEvent` itself,
  which has neither a `type` nor a `published` field.
- The commit therefore aborts at the canonical-signature check in
  `_validate_canonical_entry` — `('', 'VulnerabilityCase')` is not canonical —
  even though `("Ignore", "VulnerabilityCase")` **is** in
  `_CANONICAL_PAYLOAD_SIGNATURES` and so is clearly meant to be committable.
- Because the guarded commit precedes the effect nodes (CLP-10-006), the abort
  also skips `TransitionParticipantRMtoDeferred`.

## Why the tests did not catch it

Every existing test in `test/core/use_cases/received/case/test_engage_defer.py`
constructs its event with `activity=VultronActivity(...)` by hand — supplying
exactly the thing the registry was failing to supply. They pass with or without
`include_activity`. Nothing exercised `extract_event`, which is the path
production uses.

## Fix

`include_activity=True` on the `DEFER_CASE` entry, plus a parametrised test that
goes through `extract_event` for both defer and engage and asserts
`event.activity` is populated and carries `type_` and `published`. Verified
non-vacuous: with the flag removed the defer case fails and the engage case still
passes.

## Provenance

Found during the PR that wired the CLP-14/15 timestamp guard into the commit
boundary, while auditing every `payload_snapshot` producer for the newly required
`published` field. Pre-existing — not a regression from that PR, since the abort
already happened one check earlier and that ordering is unchanged.

## Adjacent latent item

`CaseLedger.append` (`vultron/core/models/case_ledger.py`) accepts
`payload_snapshot or {}` with no CLP-07 validation at all, so it would silently
accept a snapshot with no `published`. Currently unreachable in production —
`CaseLedger` is instantiated only in its own docstring example — but worth
closing off if that class ever becomes live.

### LINKED ISSUE #3185: spec-dump drops the note field, hiding caveats on 52 requirements from every agent that loads specs the documented way
## Summary

`spec-dump` (`vultron/metadata/specs/llm_export.py`) never mapped the spec-level
`note` field, so **52 requirements** across the corpus carried guidance that no
agent following the documented workflow could see.

## Why it matters

`AGENTS.md` says:

> **Load specs first**: `PYTHONPATH= uv run spec-dump` — never read raw
> `specs/*.yaml`.

`note` is where a requirement records the caveats that stop it being read the
wrong way — which side an obligation binds, and what a receiver must *not* do
about it. Dropping it is worse than a missing field: the requirement still reads
as complete, so nothing signals that the reader is missing the part that decides
correct implementation.

That is not hypothetical. CLP-15-001/002 were implemented as Case Actor
enforcement duties that CLP-15-005 forbids, producing seven permanently-red
`xfail(strict=True)` stubs (#2824, #3169). The amendment that fixes that reading
puts the correction in `note` — which, before this fix, would have been invisible
to the next agent through exactly the same channel.

## Evidence

- Exported keys before the fix: `adr, group, group_title, id, kind,
  postconditions, preconditions, priority, rationale, relationships, scope,
  statement, steps, tags, testable, topic, type, verification`.
- Requirements with an exported `note`: **0**. With the fix: **52**.
- `_rel_record` in the same module *does* export a relationship's `note`
  (`llm_export.py`), which is why this reads as an oversight rather than a
  deliberate exclusion.

## Affected requirements

52 across BT, CLP, CM, CP, EMB, EP, EDF, HM, MV, RSH, SYNC and TRIG. The RSH
group alone has 14.

## Fix

- Map `spec.note` in `_spec_record`.
- Extract `_behavioral_fields` so `_spec_record` stays inside the flake8
  complexity gate rather than suppressing it.
- Three tests in `test/metadata/specs/test_llm_export.py`: `note` present when
  set, omitted when unset (matching `rationale`), and a field-completeness guard
  that compares `StatementSpec.model_fields` against the exporter's output with
  an explicit `dropped` set — so the next field added to the schema forces a
  decision instead of being silently unmapped.

Field parity verified before and after: `rationale` 1430, `preconditions` 152,
`postconditions` 116, `steps` 115 unchanged. Non-vacuity verified by removing the
mapping and watching the new test fail with `KeyError: 'note'`.

## Provenance

Found by the verify phase of the PR review pipeline on #3159, checking whether
that PR's own spec amendment actually reached its audience. It did not.

sei-ahouseholder:
- Closes #2824
- Closes #3043
- Closes #3149
- Closes #3158
- Closes #3169
- Closes #3170
- Closes #3185

## Summary

`_validate_entry_timestamps` implemented CLP-14-002/003/006/007/008 but ran only under `if case_published is not None:`, and `CreateLogEntryNode.update()` — its sole production caller — never passed that argument, so the guard was dead code from the commit that introduced it. `check_clp14_timestamp_invariants` was likewise never imported by `universal_harness.py`, so no scenario ran it either, while ADR-0079 § "Validation" asserted both were running. This wires both, and splits the two timestamps an entry carries to the layers that can actually own them.

## What #2824 got wrong

The issue asked to *implement* seven invariants and delete seven `xfail(strict=True)` stubs. The invariants were already implemented; what was missing was the wiring. Three of the seven stubs could never have gone green as written:

- **CLP-14-003 / CLP-14-006 stubs** asserted cross-entry invariants against a single bare `CaseLedgerEntry`. That layer never sees its predecessor or its parent case, so it structurally cannot enforce them.
- **CLP-15-001 / CLP-15-002 stubs** expected the CaseActor to reject a participant's out-of-causal-order emission per assertion. CLP-15-005 explicitly forbids that ("MUST NOT attempt to reconstruct participant-internal causal order it cannot verify"), as does ADR-0079 § "Residual Uncertainty". Their real verification is `check_causal_edges` (DEMOMA-22-005), which is the only vantage point from which an emission-order obligation is observable.

## The two-timestamp split

An entry carries a **commit** stamp (`CaseLedgerEntry.published`, written by the CaseActor) and a **claimed** time (`payloadSnapshot.published`, chosen by the asserting actor). The old code applied CLP-14-003 and CLP-14-006 to the claimed time, which is wrong in both directions: monotonicity across *different* actors' claimed times is exactly the wall-clock ordering ADR-0079 rejected as option C, so participant A stamping T5 and B stamping T3 would have had a well-formed assertion rejected because two clocks disagreed.

Now: commit-stamp invariants go to the CI harness (true by construction, one writer one clock); claimed-time invariants go to the commit boundary, with monotonicity scoped **per snapshot actor**, which is what CLP-15-003 actually says.

## Changes

- **`vultron/core/behaviors/sync/nodes/canonical_entry.py`**: `_validate_entry_timestamps` no longer runs behind a single optional argument — each check gates on the context *it* needs. Drops the cross-actor CLP-14-003 comparison in favour of per-actor CLP-15-003; adds a clock-skew tolerance to CLP-14-006; re-cites the required-`published` check as CLP-07-011 (the verbatim-AS2-activity requirement) rather than CLP-14-002, which constrains the commit stamp the model already enforces.
- **`vultron/core/behaviors/sync/nodes/chain.py`**: resolves `case_published` and the per-actor predecessor from the DataLayer and passes them plus the configured tolerances. A missing case is expected, not an error — the genesis entry commits alongside case creation.
- **`vultron/core/sync_helpers.py`**: adds `_find_prev_actor_published`. Returns `None` when this exact assertion is already recorded, so a redelivery is not mistaken for a regression — otherwise the ordering check and the idempotency path contradict each other and out-of-order delivery (ADR-0037) becomes a hard failure.
- **`vultron/wire/as2/extractor/_builders.py`**: carries `published` across the wire→core boundary. Without it the sender's claimed time was replaced by the receiver's clock, making CLP-14-007/008 compare that clock against itself — they could never fire. Closes #3149.
- **`vultron/config/ledger.py`** (new), **`config/app.py`**, **`config/__init__.py`**: `LedgerConfig` under `AppConfig.ledger` exposes the skew tolerance, future tolerance and staleness window (CLP-14-009), settable as `VULTRON_LEDGER__*`. Nested to match the `server`/`database`/`actor` convention.
- **`vultron/core/behaviors/case/ledger_snapshots.py`**, **`receive_close_case_tree.py`**, **`use_cases/received/embargo.py`**, **`routers/demo_triggers.py`**: seven hand-built CaseActor snapshots never set `published`. Wiring the guard without this fails CLP-07-011 on every case creation, so it is not an optional part of the fix.
- **`vultron/core/models/_helpers.py`**: `parse_published` shared by the guard and the predecessor lookup (CS-22-001); replaces a private copy in `canonical_entry.py` and reuses the existing `_as_utc` rather than adding a second one.
- **`test/ci/invariants/universal_harness.py`**: runs `check_clp14_timestamp_invariants` in every scenario.
- **`test/core/behaviors/sync/nodes/test_chain_timestamp_guard.py`** (new, 8 tests): exercises the guard *through* `CreateLogEntryNode`. The old tests all called the private validator directly, which is why 14 of them passed against a guard that never ran in production.
- **`test/core/models/test_clp_14_15_invariants.py`**: all 7 `xfail` markers removed; every stub retargeted at the layer that owns its requirement, all `@pytest.mark.spec` markers preserved.
- **`docs/adr/0079-case-ledger-causal-ordering.md`**, **`notes/case-ledger-authority.md`**, **`AGENTS.md`**: the ADR's Validation section claimed checks were running that were not; the note and pitfall index now record the two-timestamp split and the three traps.
- **`AGENTS.md`**: recovered 5 lines to clear the size ratchet, which was already red on `main`. Closes #3158.

## Verification

- Unit suite: **14,115 passed**, 0 failed (419 skipped/xfail)
- Full suite (`-m ""`, includes demo integration + invariant harness): **15,372 passed**, 0 failed (421 skipped/xfail)
- **41 new tests** across 13 test files
- Black, flake8, mypy, pyright clean; `mkdocs build --strict` clean; markdownlint clean (2,174 files)
- Spec-coverage ratchet passes with zero CLP-14/CLP-15 requirements uncovered; `AGENTS.md` size ratchet green
- Each new negative test verified non-vacuous. Two checked by deliberately reverting the fix and watching the test go red: the `DEFER_CASE` registry wiring (#3170) and the redelivery carve-out. The parse-time check has an explicit anti-vacuity test pinning the sender's timestamp against the receiver's clock.

## Review round 2 — three FAIL findings and how they were resolved

PR triage found that wiring the guard, as originally written, introduced a
silent-data-loss path. Resolved with the maintainer on each point.

### CLP-15-003 is now reported, not refused

The guard refused an assertion whose claimed timestamp fell behind the same
actor's previous one — and the refusal destroyed it: the guarded commit precedes
the effect nodes (CLP-10-006), so the raise aborted the whole receive sequence
with no ledger record, no retry, and a 202 already returned to the sender. Two
distinct activities from one participant can arrive in either order (two close
POSTs become two asyncio tasks; the per-actor lock serialises but does not
order), so this fired on well-formed traffic.

It now records the entry and reports: `INFO` within the configured clock-skew
tolerance, `WARNING` beyond it. `disposition="rejected"` was considered and
rejected — in this codebase that value already means a *local-only correlation
marker with an empty payloadSnapshot* (CLP-04-001, DRPT-02-007), so reusing it
would have overloaded the field and rendered empty rows in the report builder.

### Two snapshots mixed the Case Actor's clock into a participant's stream

`receive_close_case_tree.py` and `received/embargo.py` stamped `_now_utc()` on
snapshots attributed to a *participant*, manufacturing regressions on their own
and leaving CLP-14-007/008 comparing the receiver's clock against itself on
those paths. Both now take the claimed time from the triggering activity via a
shared `claimed_published_iso()`.

### A missing inbound `published` is now a validity failure

`#3149`'s fix was only half a fix: `as_Base` declares `published` with
`default_factory=now_utc`, so a sender that simply omitted the field still got
the receiver's clock stored as its "claim", and both guards stayed vacuous and
bypassable. `parse_activity` now refuses such an activity on the raw body
(new **CLP-15-006**), which the inbox adapter maps to HTTP 422. Checked at the
parser rather than as a model validator on purpose: stored-object rehydration
goes through `model_validate` directly, and outbound authoring *should* stamp
the local clock — a class-level validator would have broken both.

The presence check the guard previously raised under CLP-07-011 was mis-cited.
CLP-07-011 requires the snapshot be the *verbatim* activity and says nothing
about `published`; AS2 makes the field optional. That rule now has a real
requirement behind it instead.

### Spec text amended (#3169)

The specs were the root cause and were left untouched by the first round:
CLP-14-002/003/006 never named *which* of an entry's two timestamps they
constrain, CLP-15-001/002 read as Case Actor duties CLP-15-005 forbids, and
CLP-14-009 authorised only two of the three tolerances implemented. All amended
in `specs/case-ledger-processing.yaml`, with `verification` pointers to the layer
that owns each half.

### Also fixed

- **#3043** — a previously-filed Concern describing this exact defect, with
  acceptance criteria this PR satisfies. Found during review; now closed here.
- **#3170** — `DEFER_CASE` omitted `include_activity=True`, so a received
  `Ignore(VulnerabilityCase)` never reached the ledger and lost the RM→DEFERRED
  transition. Pre-existing, found while auditing every snapshot producer for the
  newly required `published`. Its existing tests hand-built the activity, so they
  passed either way — the same vacuity shape this PR is about.
- `CreateLogEntryNode.update()` made two full unfiltered `list_objects` scans per
  commit, both eager even for a `disposition="rejected"` marker. Now one shared
  scan via `recorded_entries_for_case`, resolved lazily.
- `test_ledger_snapshots.py` — the test CLP-07-011 names as its verification —
  had zero `published` assertions. Now asserts every builder emits a parseable,
  UTC-aware `published`, with an exhaustiveness check so a new builder cannot
  escape it.
- `notes/configuration.md` env-var table and `AppConfig` sketch updated for the
  new `ledger` section; the cross-link to `case-ledger-authority.md` is now
  two-way.
- `parse_published` now actually returns UTC, matching its docstring.
- Added the missing CLP-14-008 boundary test and env-binding tests for the two
  config knobs that were only asserted at their defaults. Both new CLP-14-009
  tests assert the override binds *and* changes the outcome.

### Caught by verify: the spec amendment was not reaching its audience (#3185)

The verify phase checked whether this PR's own spec amendment was visible through
the channel agents actually use, and it was not. `spec-dump` never mapped the
spec-level `note` field, so the load-bearing part of the CLP-15-001/002/003
amendment — which side each obligation binds, and that a Case Actor MUST NOT
refuse on CLP-15-003 — was invisible to any agent following `AGENTS.md`'s
instruction to load specs through the exporter and never read raw `specs/*.yaml`.

That is the same failure mode #3169 was filed about, one layer down: the
requirement still reads as complete, so nothing signals the reader is missing the
part that decides correct implementation.

The gap was wider than this PR: **52 requirements** across BT, CLP, CM, CP, EMB,
EP, EDF, HM, MV, RSH, SYNC and TRIG were affected. `_rel_record` in the same
module already exported a *relationship's* `note`, so the spec-level omission was
an oversight rather than a policy. Fixed, with a field-completeness guard that
compares `StatementSpec.model_fields` against the exporter's output so the next
schema field forces a decision instead of being silently dropped.

## Follow-ups filed

- #3156 (Concern) — guards with no production caller pass their own unit tests: second confirmed witness, no structural check exists
- #3169 (Concern) — the CLP-14/CLP-15 spec-text gaps, now amended in this PR. Filed rather than left as a learnings entry, which is an input queue and not a trackable record.
- `plan/incoming/learnings/20260903-2824-clp14-15-do-not-name-their-timestamp.md` — retained as the narrative write-up; #3169 is the tracked record.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -269,7 +269,7 @@ linked file before touching that area. New pitfalls MUST be routed per
 | Domain object validation | [domain-validation](notes/domain-validation.md) | assignment and `append` bypass validation (CM-27-001, PRM-03-003); no `self` assignment in `mode="after"` validators (ARCH-21-004); silent `None` == fake `SUCCESS` (ARCH-15); `getattr` misses `ValueError`; `model_fields` is empty in `__init_subclass__`; rejecting atomically does **not** license reporting only the first violation (EH-07-001, ADR-0086); every `ParticipantStatus`-write validator calls `participant_transition_violations()`, never the individual `is_valid_*`/`violation_*` predicates (BTND-10-002), and `force_rm_state` is a pinned exemption — do not add users (#3106); trigger path fails closed while receive path partial-accepts — Postel's maxim, do not "reconcile" them |
 | Behavior tree nodes | [bt-pitfalls](notes/bt-pitfalls.md) | write nodes validate their own transitions (CSB-16, EMB-18-001); guarded commits run as CASE_MANAGER (BT-17-005/006); a BT's store follows its executing actor (BT-05-005/006); never clear a blackboard key you don't own; guard names name the transition, not the symptom |
 | BT integration / concurrency | [bt-integration](notes/bt-integration.md), [bt-pitfalls](notes/bt-pitfalls.md) | blackboard needs a module-level `RLock` under `BackgroundTasks`; trigger-side `execute()` delegates SM transitions (BT-15-001); `internal_error is False` means "nothing escaped the bridge", not "no bug" — a node's own `except Exception` and every nested-`BTBridge` hop defeat it |
-| Case ledger | [case-ledger-authority](notes/case-ledger-authority.md), [ownership-transfer](notes/ownership-transfer.md) | ledger is not a process log (CLP-07); commits are role-gated (CLP-09); `create_receive_activity_tree` already injects the guarded commit — a second one in `effect_nodes` forks the chain (CLP-09-001) |
+| Case ledger | [case-ledger-authority](notes/case-ledger-authority.md), [ownership-transfer](notes/ownership-transfer.md) | ledger is not a process log (CLP-07); commits are role-gated (CLP-09); `create_receive_activity_tree` already injects the guarded commit — a second one in `effect_nodes` forks the chain (CLP-09-001); an entry has **two** timestamps and they belong to different layers — commit stamp to the CI harness, claimed `payloadSnapshot.published` to the commit boundary, monotonic **per snapshot actor** only; every hand-built snapshot dict sets `published` (CLP-07-011); never gate a whole guard on one optional argument (#2824) |
 | Who sends what to whom | [case-communication-model](notes/case-communication-model.md) | `case_addressees()` is the wrong recipient (PCR-08-001/002); no identity spoofing or foreign CaseActor IDs on the received side; Invite/Accept routes through the Case Actor (PCR-08-007/008); delegated emit sets `actor=case_actor_id`, `attributed_to=requesting_actor_id` and gates on the **role** (CM-24) |
 | Pattern matching / semantics | [activitystreams-semantics](notes/activitystreams-semantics.md), [activitystreams-state-update](notes/activitystreams-state-update.md) | registry patterns must match the inbound wire format; `target_` is permissive unless `strict=True` (SE-08); `Reject(Invite(…))` carries the case in `inner_target` (CM-11-003); `SemanticEntry` phrases use only `{actor}`/`{object}`/`{target}` (SE-07-005) |
 | Persistence / stores | [datalayer-design](notes/datalayer-design.md) | `dl.read()` returns core objects (ADR-0034); core must not re-read wire activities for semantics (ADR-0035); an actor id **is** a store name (DL-07-004); `_dehydrate_data` deliberately keeps inline Activity sub-fields as snapshots |
@@ -298,17 +298,15 @@ linked file before touching that area. New pitfalls MUST be routed per
   or `test/`.
 - **`dl.save/create/update/delete()` in `execute()` bypasses the BT audit trail** —
   ratchet: `test/architecture/test_no_dl_mutations_in_execute.py` (#1071).
-- **Receive-side ordering is guards → commit → effects** (CLP-10-006), and
-  received-side `execute()` never calls `commit_log_entry_trigger()` directly
-  (BT-06-006, SYNC-02-002).
-- **Stub adapter files must raise `NotImplementedError`** — docstring-only stubs
-  hide integration gaps (OX-10-004, OX-11-004).
+- **Receive-side ordering is guards → commit → effects** (CLP-10-006); received
+  `execute()` never calls `commit_log_entry_trigger()` (BT-06-006, SYNC-02-002).
+  **Stub adapters raise `NotImplementedError`** — docstring-only stubs hide
+  integration gaps (OX-10-004, OX-11-004).
 - **Protocol-declared fields must stay in sync with concrete classes**, and
   `TypeGuard` discriminators may `hasattr`-check only Protocol-declared attributes
   (CS-20-001/002).
 - **Emit nodes in case-scoped trigger BTs fail fast on a missing CaseActor**
-  (PCR-08-011); **peer broadcast nodes must not mask delivery failure with
-  SUCCESS** (BT-14-001).
+  (PCR-08-011); **peer broadcast nodes must not mask delivery failure** (BT-14-001).
 - **Small habits**: mypy infers a type from the first branch assignment (use
   distinct names per `except`/`if`-else branch); pre-build dedup sets before
   fallback loops (`seen = set(d.values())`, O(n×m) → O(n+m)); walrus for
@@ -319,34 +317,31 @@ linked file before touching that area. New pitfalls MUST be routed per
 - **Superseded notes sections are archived via `append-history note`** (PD-03-002,
   PD-03-004); **large migrations partition by node shape, then domain**
   ([notes/agentic-workflow.md](notes/agentic-workflow.md)); **MkDocs `not_in_nav`
-  and `exclude_docs` are not the same**
-  ([notes/documentation-strategy.md](notes/documentation-strategy.md)).
+  ≠ `exclude_docs`** ([notes/documentation-strategy.md](notes/documentation-strategy.md)).
 - **Transport-role naming must stay explicit** — core ports docs, adapter notes,
   ADR refs and codebase reference pages change together. Likewise
-  `HashChainLedgerRecord` (in-memory) vs. `CaseLedgerEntry` (wire-serializable):
-  distinct types, imported by full module path (ARCH-12-007).
+  `HashChainLedgerRecord` (in-memory) vs. `CaseLedgerEntry` (wire): distinct
+  types, imported by full module path (ARCH-12-007).
 - **Idempotency responsibility chain** —
   [`vultron/core/AGENTS.md`](vultron/core/AGENTS.md); **avoid `BaseModel` in
   ports** — [`vultron/core/ports/AGENTS.md`](vultron/core/ports/AGENTS.md);
-  **ledger commit precedes outbox write** and **local-only correlation markers use
-  `disposition="rejected"`** —
-  [`vultron/core/behaviors/case/AGENTS.md`](vultron/core/behaviors/case/AGENTS.md).
+  **ledger commit precedes outbox write**, **local-only correlation markers use
+  `disposition="rejected"`** — [`vultron/core/behaviors/case/AGENTS.md`](vultron/core/behaviors/case/AGENTS.md).
 
 ---
 
 ## Skill Interaction Rules
 
-- Always use the `ask_user` tool for user questions — never plain text.
-- Provide a recommended answer on every `ask_user` call.
-- Rule applies transitively when skills compose (`learn` → `grill-me`, etc.).
+- Always use the `ask_user` tool for user questions — never plain text; provide a
+  recommended answer on every call. Applies transitively when skills compose
+  (`learn` → `grill-me`, etc.).
 
 ---
 
 ## Governance note for agents
 
-Agents MAY update `AGENTS.md` to correct/clarify rules, but substantive
-changes SHOULD be discussed via Issue or PR. Include rationale in the commit
-message.
+Agents MAY update `AGENTS.md` to correct/clarify rules, but substantive changes
+SHOULD be discussed via Issue or PR. Include rationale in the commit message.
 
 ---
 
@@ -359,15 +354,15 @@ message.
   note, update its `status`, `related_specs`, and `related_notes` in the same
   change; cross-links SHOULD be two-way. Full write-up + schema:
   [notes/notes-frontmatter.md](notes/notes-frontmatter.md).
-- **Docs links must be relative**: links in `docs/` MUST be relative and MUST NOT
-  go above `docs/`. Run `uv run mkdocs build --strict` before committing docs.
-  `docs/developer/` pages are draft docs — visible in `mkdocs serve` but excluded from production builds.
+- **Docs links must be relative** and MUST NOT go above `docs/`. Run
+  `uv run mkdocs build --strict` before committing docs. `docs/developer/` pages
+  are draft docs — visible in `mkdocs serve`, excluded from production builds.
 - **Demo script lifecycle logging**: see
   [`vultron/adapters/AGENTS.md`](vultron/adapters/AGENTS.md) for `demo_step` /
   `demo_check` pattern.
-- **Project history entries**: use `uv run append-history` — never write
-  directly to `plan/history/`. See HM-01–HM-05 and
-  `notes/history-management.md`. During `orient-agent`, read only `plan/*.md`.
+- **Project history entries**: use `uv run append-history` — never write directly
+  to `plan/history/`. See HM-01–HM-05 and `notes/history-management.md`. During
+  `orient-agent`, read only `plan/*.md`.
 
 ---
 
--- a/docs/adr/0079-case-ledger-causal-ordering.md
+++ b/docs/adr/0079-case-ledger-causal-ordering.md
@@ -152,17 +152,46 @@ from a malicious or malfunctioning participant.
 
 ## Validation
 
-Conformance tests in `test/ci/invariants/common.py` verify:
+Enforcement splits along the two timestamps an entry carries, and the split is
+load-bearing — conflating them produces either a vacuous check or a false
+rejection (Issue #2824).
 
+**Commit timestamp** (`CaseLedgerEntry.published`, stamped by the CaseActor).
+`check_clp14_timestamp_invariants` in `test/ci/invariants/common.py`, run by
+every scenario through `universal_harness.py`, verifies:
+
+- `entry.published` is non-null for all entries.
 - `entry[N].published >= entry[N-1].published` for all consecutive recorded
   entries within a case.
-- `entry.published` is non-null for all entries.
 - `log_index` values are unique within a case and form a gapless sequence.
 - All entries in a case share the same `case_id`.
-- No entry's `published` precedes the case's own `published` timestamp.
+- No entry's `published` precedes the case-creation entry's timestamp.
 
-These are tracked as a separate implementation issue (runtime enforcement via
-extension of `_validate_canonical_entry` and conformance tests).
+A single writer stamping from one clock satisfies these by construction; the
+harness is what catches a replica that does not.
+
+**Claimed timestamp** (`payloadSnapshot.published`, chosen by the asserting
+actor). `_validate_entry_timestamps`, reached from every commit through
+`CreateLogEntryNode`, rejects an assertion that:
+
+- carries no parseable `published` at all (CLP-07-011);
+- predates the case by more than the configured clock-skew tolerance
+  (CLP-14-006);
+- regresses behind that *same actor's* previous assertion (CLP-15-003);
+- sits beyond the future tolerance or staleness window (CLP-14-007,
+  CLP-14-008).
+
+Monotonicity of the claimed timestamp is scoped to one actor's stream. Applying
+it across actors would reintroduce option C: two participants' clocks are not
+comparable, so a later-committed assertion from one legitimately carries an
+earlier claimed time than an earlier-committed one from another. Thresholds are
+tunable via `AppConfig.ledger` (CLP-14-009).
+
+The participant emission obligations CLP-15-001 and CLP-15-002 are not
+enforceable per-assertion — that is the point of the residual uncertainty above,
+and CLP-15-005 forbids the attempt. They are validated at the conformance layer
+by `check_causal_edges`, which compares declared causal edges against observed
+ledger order across a whole scenario (DEMOMA-22-005).
 
 ## More Information
 
--- a/docs/topics/case_ledger_sync.md
+++ b/docs/topics/case_ledger_sync.md
@@ -157,9 +157,30 @@ The Case Actor should also refuse an assertion whose own timestamp is far in
 the future or far in the past compared to its clock — by default, more than
 five minutes ahead or more than seven days old
 ([CLP-14-007](../reference/specs/protocol.md#clp-14),
-[CLP-14-008](../reference/specs/protocol.md#clp-14)). These are sanity
+[CLP-14-008](../reference/specs/protocol.md#clp-14)). A deployment that knows
+its own clock conditions may tune both thresholds
+([CLP-14-009](../reference/specs/protocol.md#clp-14)). These are sanity
 checks, not proof of honesty; a participant with a badly wrong clock can still
 produce a well-formed but misleading assertion.
+
+### Two timestamps, two different rules
+
+An entry carries two timestamps, and the rules above do not all apply to the
+same one:
+
+| Timestamp | Written by | Rules it answers to |
+|---|---|---|
+| `published` on the entry | the Case Actor, at commit time | CLP-14-002, CLP-14-003, CLP-14-006 read across the whole ledger |
+| `published` inside `payloadSnapshot` | the asserting participant | CLP-14-006 at the commit boundary, CLP-14-007, CLP-14-008, CLP-15-003 |
+
+The distinction matters most for the "never move backwards" rule. Across the
+Case Actor's own commit timestamps it holds because one writer stamps every
+entry from one clock. Applied to *participants'* timestamps it would be wrong:
+two participants' clocks are not synchronized, so an assertion recorded later by
+one of them can legitimately claim an earlier time than an assertion recorded
+earlier by the other. The Case Actor therefore compares a participant's claimed
+timestamp only against that same participant's previous assertion, and allows a
+small tolerance when comparing it against the case's creation time.
 
 ---
 
@@ -180,6 +201,15 @@ each participant: send events in the order they happened.
 
 A participant that breaks these rules produces a malformed assertion. That is
 a fault on the sending side, not a Case Actor failure.
+
+The Case Actor cannot police all four at the moment an assertion arrives, and it
+must not try: it may not reconstruct an order it did not observe
+([CLP-15-005](../reference/specs/protocol.md#clp-15)). What it can and does
+check is the timestamp rule — an assertion whose claimed time falls behind the
+same participant's previous assertion is refused. The two ordering rules are
+visible only across a whole case, by comparing the causal links a scenario
+declares against the order the ledger recorded; the conformance suite
+checks them there.
 
 ---
 
--- a/notes/case-ledger-authority.md
+++ b/notes/case-ledger-authority.md
@@ -9,6 +9,7 @@ related_specs:
 related_notes:
   - notes/activitystreams-semantics.md
   - notes/case-state-model.md
+  - notes/configuration.md
   - notes/ownership-transfer.md
   - notes/sync-ledger-replication.md
 relevant_packages:
@@ -332,6 +333,47 @@ that rejects entries violating CLP-07-001 through CLP-07-004 *before*
 they enter the hash chain. Failing fast at commit time keeps the
 canonical chain clean and surfaces bugs immediately, rather than allowing
 silent pollution that's discovered only when replicas diverge.
+
+### An Entry Has Two Timestamps, and They Belong to Different Layers
+
+*Spec: CLP-14, CLP-15; ADR-0079 § "Validation". Issue #2824.*
+
+`CaseLedgerEntry.published` is the CaseActor's **commit** stamp.
+`payloadSnapshot.published` is the asserting actor's **claimed** event time.
+Deciding which one an invariant is about is not a detail — get it wrong and the
+check is either vacuous or falsely rejecting:
+
+- Commit-timestamp invariants (CLP-14-002/003/006 read against the envelope)
+  hold **by construction** — one writer, one clock. They belong to the
+  conformance harness (`check_clp14_timestamp_invariants`), which is the only
+  vantage point that sees a whole ledger. An entry never sees its predecessor,
+  so the *model* layer can never enforce them; asserting that it should is what
+  produced seven permanently-red `xfail(strict=True)` stubs.
+- Claimed-timestamp invariants (CLP-14-006/007/008, CLP-15-003) cannot hold by
+  construction, because a participant chose the value. They belong to
+  `_validate_entry_timestamps` at the commit boundary.
+
+Three traps, all found the hard way:
+
+1. **Claimed-timestamp monotonicity MUST be scoped per snapshot actor.**
+   Comparing claimed times across actors is exactly the wall-clock ordering
+   ADR-0079 rejected as option C. CLP-15-003 says "within the same
+   participant's event stream" for this reason.
+2. **A CaseActor-authored snapshot needs a `published` too.** Hand-built
+   snapshot dicts kept omitting it, which makes the snapshot something other
+   than the verbatim AS2 activity CLP-07-011 requires. New commit call sites
+   must set it (`_now_utc().isoformat()`).
+3. **Never gate a whole guard on one optional argument.** The CLP-14 guard sat
+   behind `if case_published is not None:` and the sole production call site
+   never passed it, so nothing was checked for the guard's entire life while
+   its unit tests passed by calling it directly. Gate each check on the context
+   *it* needs, and test enforcement through the production node, not the
+   private validator.
+
+CLP-15-001 and CLP-15-002 bind the *participant*, and CLP-15-005 forbids the
+CaseActor from reconstructing participant-internal causal order. Their
+verification is `check_causal_edges` (DEMOMA-22-005), not a per-assertion
+check.
 
 ---
 
--- a/plan/incoming/learnings/20260903-2824-clp14-15-do-not-name-their-timestamp.md
+++ b/plan/incoming/learnings/20260903-2824-clp14-15-do-not-name-their-timestamp.md
@@ -1,0 +1,63 @@
+---
+title: "CLP-14/CLP-15 constrain two different timestamps without naming which, and CLP-15-001/002 are written as if the CaseActor could enforce what CLP-15-005 forbids it from reconstructing"
+type: learning
+timestamp: "2026-09-03T20:40:00Z"
+source: ISSUE-2824
+signal: spec-contradiction
+---
+
+A `CaseLedgerEntry` carries two timestamps, and CLP-14/CLP-15 never say which
+one each requirement is about:
+
+- `CaseLedgerEntry.published` — the CaseActor's **commit** stamp.
+- `payloadSnapshot.published` — the asserting actor's **claimed** event time.
+
+The distinction is not cosmetic. It decides whether a requirement holds by
+construction or has to be enforced, and picking the wrong field produces a check
+that is either vacuous or falsely rejecting:
+
+| Requirement | Statement says | Which field it can only mean |
+|---|---|---|
+| CLP-14-002 | "The published timestamp on a `CaseLedgerEntry`" | commit stamp |
+| CLP-14-003 | "published timestamps on consecutive recorded entries" | commit stamp — one writer, one clock, so true by construction |
+| CLP-14-006 | "Every CaseLedgerEntry's published timestamp" | ambiguous; enforceable only against the claimed time, since the commit stamp is `now()` and trivially satisfies it |
+| CLP-14-007/008 | "`payloadSnapshot.published`" | claimed time — the only two that say so explicitly |
+| CLP-15-003 | "the published timestamp on activity B" | claimed time |
+
+Applied to the wrong field, CLP-14-003 is actively harmful: comparing *claimed*
+timestamps across two participants is exactly the wall-clock ordering ADR-0079
+rejected as option C. Participant A stamps T5, B stamps T3, the CaseActor
+receives A then B, and a well-formed assertion is rejected because two clocks
+disagree. CLP-15-003 is the same property stated correctly — "within the same
+participant's event stream" — which is what makes it enforceable where
+CLP-14-003 is not.
+
+**The contradiction.** CLP-15-001 ("a participant MUST emit ... in causal
+order") and CLP-15-002 ("MUST NOT batch ... in arbitrary order") are written in
+a group whose title is *Participant Assertion Timestamp Obligations*, but nothing
+in either says who checks them. CLP-15-005, four entries later, says the
+CaseActor "MUST NOT attempt to reconstruct participant-internal causal order it
+cannot verify." A reader implementing CLP-15-001 per-assertion is implementing
+the thing CLP-15-005 prohibits — and it cannot be done anyway, because
+`_validate_canonical_entry` is stateless. Seven `xfail(strict=True)` stubs were
+written against exactly that reading and could never have gone green.
+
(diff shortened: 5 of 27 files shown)


sei-ahouseholder (commented):
## PR Triage: #3159 — fix(ledger): wire the CLP-14/15 timestamp guard into the commit boundary

**Linked issues**: #2824, #3149, #3158
**Changed files**: 27 files (+1182 / −254) — core/behaviors/sync, core/behaviors/case, core/use_cases/received, core/models, wire/as2/extractor, config, adapters/driving/fastapi, test/ci/invariants
**CI status**: ✅ passing (22 checks green on `0d8b2a7`, incl. 4 demo-integration + 4 invariant-harness scenarios)
**Base branch**: `main`

---

### Findings

|---|---|---|---|---|
| `phase8-clp15003-reorder-silent-loss-0` | code-review | ❌ FAIL | CLP-15-003 hard-rejects a legitimate assertion when two distinct activities from one participant arrive out of emission order — and the activity is silently, permanently lost (no `rejected` entry, no retry, effect nodes skipped) | fix-now |
| `phase8-clp15003-cross-clock-zero-tolerance-1` | code-review | ❌ FAIL | CLP-15-003 compares with zero tolerance across two clocks in one actor's stream: `receive_close_case_tree.py:159` and `embargo.py:487` stamp the CaseActor's clock under a participant's actor URI. Any forward participant skew fails the owner-Leave close commit | fix-now |
| `phase5-clp14006-tolerance-unauthorised-2` | spec-conformance | ❌ FAIL | A 5-min tolerance weakens MUST-level CLP-14-006, which no spec entry authorises (CLP-14-009 is MAY and covers only CLP-14-007/008); ADR-0079 amended but `specs/case-ledger-processing.yaml` untouched | fix-now |
| `phase5-clp14-15-spec-text-unamended-3` | spec-conformance | 🎫 NEW-ISSUE | Spec-level premise inversion tracked only in a `plan/incoming/learnings/` file, which the completeness doctrine says is not tracking | inversion-halt |
| `phase8-clp14-007-008-still-bypassable-4` | code-review | ⚠️ IMPROVE | The #3149 fix is partial: `as_Base.published` has `default_factory=now_utc`, so a sender that omits `published` still gets the receiver's clock stored as its "claim" — CLP-14-007/008 stay vacuous and bypassable | fix-now |
| `phase10-ledger-snapshots-published-untested-5` | test-coverage | ⚠️ IMPROVE | `test/core/behaviors/case/test_ledger_snapshots.py` (CLP-07-011's declared verification) has zero `published` assertions and was not touched; no boundary test for CLP-14-008; 2 of 3 new env vars unverified | fix-now |
| `phase9-notes-configuration-stale-6` | notes-currency | ⚠️ IMPROVE | `notes/configuration.md` env-var table and `AppConfig` sketch omit the new `ledger` section; the new cross-link from `case-ledger-authority.md` is one-way | fix-now |
| `phase8-double-full-ledger-scan-7` | code-review | ⚠️ IMPROVE | `CreateLogEntryNode.update()` now makes two eager full `list_objects("CaseLedgerEntry")` scans per commit (three with `ReconstructChainTail`), unfiltered by case, paid even for `disposition="rejected"` | fix-now |
| `phase8-defer-case-missing-include-activity-8` | code-review | 🎫 NEW-ISSUE | `DEFER_CASE` omits `include_activity=True`, so a received `Ignore(VulnerabilityCase)` never ledgers and loses the RM→DEFERRED transition. Pre-existing, **not** a regression here | fix-now-file |
| `phase3-verification-missing-test-counts-9` | pr-body-format | ⚠️ IMPROVE | Verification section reports "exit 0" instead of the real total/new test counts required by `pr-body-guide.md` | fix-now |
| `phase8-clp14006-case-published-cross-clock-10` | code-review | ⚠️ IMPROVE | `_build_case_object` keeps a wire-sourced originator's clock on `case.published`, so CLP-14-006 can mix clocks too (candidate exposure: ownership-transfer transferee) — PLAUSIBLE, unconfirmed | fix-now |
| `phase8-parse-published-docstring-inaccurate-11` | code-review | ⚠️ IMPROVE | `parse_published` promises a UTC datetime but `_as_utc` leaves an aware non-UTC offset unchanged. Harmless for comparisons; wrong contract on a newly shared helper | fix-now |

**Total**: 3 FAIL · 7 IMPROVE · 2 NEW-ISSUE

---

### What's solid

Worth stating plainly, because the diff is large and the findings above are concentrated in one area:

- The core diagnosis is right and well evidenced. `_validate_entry_timestamps` really was dead code gated behind one optional argument, `check_clp14_timestamp_invariants` really was never imported by `universal_harness.py`, and ADR-0079 really did assert both were running.
- Removing all 7 `xfail(strict=True)` markers while retargeting each stub at the layer that owns its requirement is the correct call, and every `@pytest.mark.spec` marker survived.
- `test_chain_timestamp_guard.py` exercising the guard **through** `CreateLogEntryNode` is the right shape, and the `config_override` test asserting the override actually bound is exactly the anti-vacuity discipline this PR is about.
- An exhaustive AST audit of every `payload_snapshot` dict literal in `vultron/` carrying both `type`/`type_` and `actor` returns 8 producers, all with `published`. The PR's claim of seven missed sites is accurate.
- CI is fully green on the head SHA, and `test/metadata` + `test/architecture` pass locally (AGENTS.md is back to exactly 400 lines, clearing #3158).

### The through-line in the three FAILs

All three are the same root shape: the PR correctly identifies that comparing *claimed* timestamps across actors is the wall-clock ordering ADR-0079 rejected, then applies that insight only to the cross-actor case. Within one actor's stream the same hazard is still live — arrival order is used as a proxy for causal order (which CLP-15-005 forbids and CLP-15-003's own "causally related" qualifier excludes), the comparison has no skew tolerance, and two of the PR's own new `published` stamps mix the CaseActor's clock into a participant's stream. The failure mode is worse than a false rejection: the activity is dropped with no ledger record and no retry.

---

*Triage artifact: `.claude/pr-3159-triage.json`*
*Next step: `/pr-execute` (or `/pr-ship`) — note that finding `phase5-clp14-15-spec-text-unamended-3` is an `inversion-halt`, so execute will pause for a human decision.*

sei-ahouseholder (commented):
## PR Execute: #3159

**Findings**: 12 from triage — 11 fixed, 1 `no_change_needed`
**Issues filed**: #3169 (Concern), #3170 (Bug) — both closed by this PR
**Closing refs added**: #3043, #3169, #3170
**CI**: ✅ 25/25 passing on `d3f553d7`
**Suite**: 14,112 unit passed · 15,369 full-suite passed · 0 failed · 38 new tests

---

### The three FAILs

All three were the same root shape: the original diff correctly identified that comparing *claimed* timestamps across actors is the wall-clock ordering ADR-0079 rejected, then applied that insight only to the cross-actor case. Within one actor's stream the hazard was still live.

| # | Resolution |
|---|---|
| `phase8-clp15003-reorder-silent-loss-0` | CLP-15-003 no longer raises. The entry is recorded and the regression reported — `INFO` within the configured skew tolerance, `WARNING` beyond it. This removes the silent-loss path: the assertion is always committed, so the effect nodes and `StoreActivityNode` still run. |
| `phase8-clp15003-cross-clock-zero-tolerance-1` | Both sites that stamped the Case Actor's clock under a participant's actor URI now take the claimed time from the triggering activity, via one shared `claimed_published_iso()`. |
| `phase5-clp14006-tolerance-unauthorised-2` | `CLP-14-009` amended to cover all three tolerances actually implemented; `CLP-14-006` amended to state its claimed-timestamp clause explicitly. |

`disposition="rejected"` was considered for the first and **rejected on review**: in this codebase that value already means a *local-only correlation marker with an empty `payloadSnapshot`* (CLP-04-001, DRPT-02-007), so reusing it would have overloaded an existing term of art and rendered empty rows in the report builder.

### Gate 2 — inversion

Put to the maintainer, who chose **file it and amend the spec here**. #3169 filed and the spec amended in this PR, so nothing is left inconsistent at merge. The learnings entry stays as the narrative write-up; #3169 is the tracked record.

### Scope the maintainer widened

`phase8-clp14-007-008-still-bypassable-4` was resolved more thoroughly than triage proposed. Rather than skipping checks when a claimed time is absent, **a missing inbound `published` is now a message-validity failure** at `parse_activity`, grounded in a new **CLP-15-006**.

Placed at the parser rather than as a validator on `as_Base`, and that placement was verified rather than assumed: stored-object rehydration goes through `model_validate` directly and never through `parse_activity`, so a class-level validator would have broken every stored record and every outbound factory. Blast radius was measured empirically — make the change, run the full suite — giving 10 hand-written fixtures plus the tutorial example, all fixed. Only `published` is required, not `updated`: `published` is the field with protocol meaning.

### Two things triage missed

- **#3043** — a Concern already on file describing this PR's exact defect, with two acceptance criteria the PR satisfies. Now in the closing list.
- The presence check the guard raised under **CLP-07-011** was mis-cited. CLP-07-011 requires the snapshot be the *verbatim* activity and says nothing about `published`; AS2 makes the field optional. That rule now has a real requirement behind it.

### Non-vacuity

Given this PR exists because a guard passed its own tests while never running, each new negative test was checked:

- `include_activity=True` (#3170) — removed it, confirmed the defer parametrisation fails and the engage one still passes.
- The CLP-15-003 bands — a third test asserts **no** report is emitted when timestamps do not regress, so the band tests cannot pass against a guard that logs unconditionally.
- The parse-time check — pins the sender's value against the receiver's clock.
- The two CLP-14-009 tests assert the override binds **and** changes the outcome.
- `test_ledger_snapshots.py` — introspects the module for `build_*_snapshot` names, so a new builder cannot escape the `published` assertion.

One test I wrote was itself vacuous (a year comparison that short-circuited to `True`); it was removed rather than kept.

### Not fixed, deliberately

`phase8-clp14006-case-published-cross-clock-10` — triage marked it PLAUSIBLE and the reviewer could not substantiate the ownership-transfer-transferee path by reading code. The main Case Actor path is safe because `case_proposal_received_tree.py` mints the case locally. Fixing blind would mean guessing at a path nobody has demonstrated; it is recorded in #3169's scope.

### Also, from main mid-run

`test_case_resolution_uses_helpers` (ADR-0087) landed on `main` during the sync and flagged this PR's `read_case`. Added to `KNOWN_ALLOWLIST` under **Regime 3** with a category comment: the genesis `create_case` entry commits alongside case creation, so absence is expected and `_require_case` would turn the genesis commit into a `FAILURE`.

---

*Execute artifact: `.claude/pr-3159-execute.json`*
*Next step: `/pr-verify`.*

sei-ahouseholder (commented):
## PR Verify: #3159 — verdict `GAPS-FOUND`

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`; `origin/main` is an ancestor of HEAD; no conflict markers
**CI**: ✅ 23 pass / 2 skipping / 0 failing on `d3f553d7` — the exact PR head sha; all 8 Demo Integration + Invariant Harness jobs green
**Artifact integrity**: 12 triage findings ↔ 12 execute findings, all IDs matched

---

### Per-finding verdicts

| Finding | Severity | Verdict |
|---|---|---|
| `phase8-clp15003-reorder-silent-loss-0` | FAIL | ✅ CONFIRMED — no `raise` cites CLP-15-003 at HEAD; the two-band `logger.info`/`logger.warning` selector is present |
| `phase8-clp15003-cross-clock-zero-tolerance-1` | FAIL | ✅ CONFIRMED — both snapshots call `claimed_published_iso()`; no `_now_utc()`/`datetime.now()` remains on either |
| `phase5-clp14006-tolerance-unauthorised-2` | FAIL | ✅ CONFIRMED — CLP-14-006/14-009 amended and name the field |
| `phase5-clp14-15-spec-text-unamended-3` | NEW-ISSUE | ⚠️ **PARTIAL** — see gap below |
| `phase8-clp14-007-008-still-bypassable-4` | IMPROVE | ✅ CONFIRMED — `VultronParseMissingPublishedError` raised in `parse_activity` |
| `phase10-ledger-snapshots-published-untested-5` | IMPROVE | ✅ CONFIRMED — 10 `published` references in the CLP-07-011 verification test |
| `phase9-notes-configuration-stale-6` | IMPROVE | ✅ CONFIRMED — 3 `VULTRON_LEDGER__*` rows, cross-link two-way |
| `phase8-double-full-ledger-scan-7` | IMPROVE | ✅ CONFIRMED — one `recorded_entries_for_case` scan shared by both lookups |
| `phase8-defer-case-missing-include-activity-8` | NEW-ISSUE | ✅ CONFIRMED — `include_activity=True` at HEAD, registry test green |
| `phase3-verification-missing-test-counts-9` | IMPROVE | ✅ CONFIRMED — real counts in the PR body |
| `phase8-clp14006-case-published-cross-clock-10` | IMPROVE | 📝 NOTED — `no_change_needed`; PLAUSIBLE and unsubstantiated, recorded in #3169's scope |
| `phase8-parse-published-docstring-inaccurate-11` | IMPROVE | ✅ CONFIRMED — `astimezone(timezone.utc)` present |

Issues real and open: #3169, #3170, #3043.

---

### The gap

**`spec-dump` does not export the `note:` field, so part of the #3169 amendment is invisible via the documented path.**

The spec amendment put three pieces of load-bearing guidance into `note:` — that CLP-15-001/002 bind the participant only and are **not** per-assertion enforceable, and that a Case Actor MAY report a CLP-15-003 regression but MUST NOT refuse on it. Verification against the exporter's own output shows:

- `spec-dump` exports these keys and no others: `adr, group, group_title, id, kind, postconditions, preconditions, priority, rationale, relationships, scope, statement, steps, tags, testable, topic, type, verification`.
- Requirements repo-wide with an exported `note`: **0**.
- `vultron/metadata/specs/llm_export.py` never maps `spec.note`. Notably `_rel_record` *does* export a relationship's `note`, so this reads as an oversight rather than a deliberate exclusion. Three pre-existing requirements are affected the same way.

`AGENTS.md` instructs agents to load specs via `PYTHONPATH= uv run spec-dump` and **never** to read raw `specs/*.yaml`. So an agent following the documented workflow cannot see this guidance — which is the exact failure mode #3169 was filed to fix: a requirement whose reader never receives the part that matters. The amendment is correct in the YAML; it just does not reach its audience.

This is a gap in the fix, not in the diagnosis, and it was introduced by this session rather than inherited.

### Not treated as gaps

- The `merge_state.synced` field was missing from the execute artifact — my own bookkeeping omission, not a real unsynced branch. The sync is independently verifiable: `origin/main` is an ancestor of HEAD via merge commit `7ee18a163`, and `sync-with-main.sh` returned exit 0 on a clean merge. Recorded with that evidence rather than left to produce a false `UNSYNCED-EXECUTE`.
- `phase8-clp14006-case-published-cross-clock-10` is `no_change_needed` by design, not an unresolved FAIL.

---

Artifacts preserved at `.claude/pr-3159-{triage,execute}.json`. Fixing the export gap next, then re-verifying.

sei-ahouseholder (commented):
## PR Verify (re-run): #3159 — verdict `READY-TO-MERGE`

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`; `origin/main` is an ancestor of HEAD; no conflict markers; not a draft
**CI**: ✅ 23 pass / 2 skipping / 0 failing on `22973f73` — the exact PR head sha
**Artifact integrity**: 12 triage findings ↔ 12 execute findings, all IDs matched; `merge_state.synced` recorded
**Suite**: 14,115 unit passed · 15,372 full-suite passed · 0 failed · 41 new tests
**Working tree**: clean

---

### The gap from the previous verify is closed

`spec-dump` now exports the `note` field, confirmed at HEAD:

- requirements exporting `note`: **52** (was 0)
- CLP-15-001, CLP-15-002, CLP-15-003 — all three present

So the load-bearing part of the #3169 amendment now reaches agents through the channel `AGENTS.md` tells them to use. Filed as #3185 and closed by this PR, with a field-completeness guard that compares `StatementSpec.model_fields` against the exporter's output so the next unmapped field fails at authoring time.

Field parity verified unchanged either side of the fix: `rationale` 1430, `preconditions` 152, `postconditions` 116, `steps` 115. Non-vacuity verified by removing the mapping and watching the new test fail with `KeyError: 'note'`.

### All findings confirmed at HEAD

| Finding | Severity | Verdict |
|---|---|---|
| `phase8-clp15003-reorder-silent-loss-0` | FAIL | ✅ CONFIRMED |
| `phase8-clp15003-cross-clock-zero-tolerance-1` | FAIL | ✅ CONFIRMED |
| `phase5-clp14006-tolerance-unauthorised-2` | FAIL | ✅ CONFIRMED |
| `phase5-clp14-15-spec-text-unamended-3` | NEW-ISSUE | ✅ CONFIRMED — spec amended **and** now reachable via `spec-dump` |
| `phase8-clp14-007-008-still-bypassable-4` | IMPROVE | ✅ CONFIRMED |
| `phase10-ledger-snapshots-published-untested-5` | IMPROVE | ✅ CONFIRMED |
| `phase9-notes-configuration-stale-6` | IMPROVE | ✅ CONFIRMED |
| `phase8-double-full-ledger-scan-7` | IMPROVE | ✅ CONFIRMED |
| `phase8-defer-case-missing-include-activity-8` | NEW-ISSUE | ✅ CONFIRMED |
| `phase3-verification-missing-test-counts-9` | IMPROVE | ✅ CONFIRMED |
| `phase8-clp14006-case-published-cross-clock-10` | IMPROVE | 📝 NOTED — `no_change_needed`, PLAUSIBLE and unsubstantiated; recorded in #3169's scope |
| `phase8-parse-published-docstring-inaccurate-11` | IMPROVE | ✅ CONFIRMED |

Closes: #2824, #3043, #3149, #3158, #3169, #3170, #3185 — all real and open.

### One CI flake, discriminated not assumed

`fcvcv Demo Integration` and `fcvcv Invariant Harness` failed on run `34862390556`, then **passed on rerun of the identical commit**. The commit under test touched only `vultron/metadata/specs/llm_export.py` and its test, neither imported by the demo runtime.

The signature — `wait_for_case_participants` timing out in `_phase_report_submission` waiting for the late joiner's case replica — matches **#2898**, which predicts the drain workaround is load-sensitive. Recorded there as a datapoint rather than filed as a duplicate: [comment](https://github.com/CERTCC/Vultron/issues/2898#issuecomment-5666680282).

That comment also notes a secondary problem worth a maintainer's eye: the Invariant Harness reports `"This is a real invariant failure, not missing test data"` when the *demo* aborted upstream, so a delivery flake presents as a broken ledger invariant. Not touched here — it is unrelated to this PR.

### Verdict

Every FAIL is `CONFIRMED` at HEAD, CI is green on the head sha, the branch is live-`MERGEABLE`, and no `UNSYNCED-EXECUTE` / `BRANCH-BEHIND` / `PR-IS-DRAFT` flag is set.

`READY-TO-MERGE`

Issue #2684, opened 2026-08-26
VFD state-machine split completion & CS transition correctness


sei-ahouseholder:
## Design idea

Complete the VFD→VF+D split and enforce correct CaseStatus state-machine transitions, ordering invariants, and notification wiring.

## Open work

- #2662 — add CS_vf, CS_d, VfDimension, DDimension alongside existing CS_vfd types
- #2663 — split ParticipantStatus.vfd into vf/d fields and migrate all call sites
- #2664 — remove CS_vfd, VfdDimension and retired VFD guard nodes
- #2524 — wire cs_invariants ephemeral-state and history guards into CS receive path
- #2175 — add EmitCaseStatusUpdateNode; wire after all EM lifecycle BT nodes
- #2527 — CS ordering constraints under out-of-order ledger delivery (ADR-0037) are unspecified
- #2112 — _validate_transitions returns only the first error when multiple dimensions are simultaneously invalid

## Rationale

Calved 2026-08-26 from **#1935 Protocol authority & lifecycle semantics** to isolate the concrete VFD-split implementation thread from authority-model and participant-lifecycle design work.

PR #3203, opened 2026-09-14
docs: plan concern #3111 — one ParticipantStatus writer

### LINKED ISSUE #3111: Consolidate the two ParticipantStatus writers outside the composed evaluator
## Summary

BTND-10-002 requires the rules governing a `ParticipantStatus` write to be
composed into one evaluator that every validating node calls. #3050 did that for
the two BT nodes on the case-participant emit path
(`participant_transition_violations()`). Two other writers validate a transition
and write a `ParticipantStatus` without going through it, each enforcing a
different subset.

This issue is the **design question about consolidating them**. The data-loss
defect originally described here — `append_rm_state()` resetting a vendor's `vf`
by omitting it — was a separate bug and is fixed in PR #3110.

| Writer | Rules enforced | Scope |
|---|---|---|
| `participant_transition_violations()` | RM + VF + D + PXA transitions, VENDOR/DEPLOYER gates, cross-machine entailments, compound CS | case participant, emit path |
| `CaseParticipant.append_rm_state()` | RM adjacency only | case participant, **model method** |
| `_ReportPhaseRMTransition._guard_transition()` | RM adjacency only | report phase, pre-case |

The receive path (`_adjudicate_dimensions()`) is deliberately different — that is
Postel's maxim (ADR-0086, ADR-0061) — and is not in question.

## 1. `CaseParticipant.append_rm_state()` — the open question

It writes a `ParticipantStatus` to the **same** case-participant object the
composed evaluator governs, but validates RM adjacency and nothing else. It is
live: three call sites in `vultron/core/use_cases/_helpers.py`, including
`update_participant_rm_state()`, reached from `rm_transitions.py`, `owner.py` and
`received/case/validate.py`.

So the same domain write is subject to two different rule sets depending on
whether it goes through the BT node or the model method.

**Two candidate designs, neither obviously right:**

**A — route it through the composed evaluator.** Three problems found while
scoping this:

- It would newly enforce the cross-machine entailments on RM-only appends,
  because the entailments read the *effective* `vf`/`d` even when only `rm` is
  asserted. Unmeasured blast radius, and the same shape of surprise that the RM
  rule produced in #3050 (see #3106).
- The model cannot see the case, so it cannot resolve the PXA baseline the
  evaluator expects.
- Probable import cycle: `models` → `states.participant_transitions` →
  `predicates.roles`, on top of the `models` ↔ `states` cycle already documented
  in `notes/domain-validation.md`.

**B — move the callers to the BT write path** and reduce or delete the model
mutator. Architecturally cleaner: the write then lands in the BT audit trail,
which is the standing rule (`test/architecture/test_no_dl_mutations_in_execute.py`,
#1071). Larger refactor, and it changes three call sites in the received-case and
report-phase paths.

## 2. `_ReportPhaseRMTransition` — resolved

Confirmed a deliberate separate lifecycle and now recorded as such in the class
docstring: it operates on a *report* before a case exists, so there is no case
participant, no VF/D/PXA dimension and no role to gate on, and its current state
is report-scoped (`_current_report_phase_rm_state`) rather than case-scoped. RM
adjacency is the whole rule set that applies.

Still open only as part of question 1: whether the report-phase and
case-participant lifecycles should eventually share one evaluator.

## Current state

`test/architecture/test_participant_status_validation.py` discovers
`ParticipantStatus`-write validators **structurally** — any module that names a
transition predicate and constructs a dimension object — and fails when one
appears that is not declared. Both writers above are declared exclusions with
reasons, so the divergence is visible and cannot silently grow. It is not
resolved.

## Questions to settle

1. Design A or B for `append_rm_state()`?
2. If the model method stays, what stops it drifting from the composed rule set?
   It can see roles but not the case.
3. Should the report-phase and case-participant lifecycles share one evaluator,
   or stay separate with the scoping recorded (as now)?

## Reference

- Specs: BTND-10-002, BTND-10-003, ARCH-15-004
- ADR: `docs/adr/0086-report-every-violation-reject-the-batch.md`
- Notes: `notes/domain-validation.md` § "Compose the rule set, don't just share the predicates"
- Ratchet: `test/architecture/test_participant_status_validation.py` (`_DECLARED_EXCLUSIONS`)
- Corroborates the #2906 learning: predicates with callers on only one path are
  either a deliberate scoping decision or a gap
- Surfaced during #3050 / PR #3110; related #3106

sei-ahouseholder:
- Closes #3111

## Summary

Settles CONCERN-3111. Scoping it found **seven** `ParticipantStatus` writers
outside the composed evaluator, not two — and four of them were structurally
invisible to the ratchet built to discover the population.

## Changes

- **`docs/adr/0089-one-participant-status-writer.md`** (new): `ParticipantStatus`
  becomes ladder-only with a single writer, reached only as a real tree child,
  with one mechanism per input. Pre-case RM state relocates to `ReportCaseLink`,
  which ADR-0041 already created for that window. Records the rejected
  alternatives — reviving the proto-case (ADR-0015, superseded by ADR-0041), a
  fourth ADR-0087 disposition, a separate marker type, and nested delegation.
- **`specs/behavior-tree-node-design.yaml`**: amend BTND-10-002 to bind every
  *writer* rather than every validating *node* — three unvalidated writers and
  one non-node writer escaped through the old wording. Add BTND-10-004 (tree
  child, never nested), BTND-10-005 (one mechanism per input), BTND-10-006
  (pre-case RM state belongs to `ReportCaseLink`).
- **`notes/domain-validation.md`**: the seven-writer inventory; why the ratchet's
  gate inverted the incentive (it keyed on evidence of validating badly, so a
  writer that validated nothing was invisible); the two entangled jobs of
  `ParticipantStatus` and the id-reuse promotion between them; and the measured
  proof that the RM-coupled entailments cannot fire on an RM-only advance in any
  legal state.
- **`docs/adr/index.md`**, **`mkdocs.yml`**: register ADR-0089.
- **`docs/adr/archived/README.md`**: drop the stale "currently empty" claim —
  ADR-0015 is archived there (MS-16-001).

## Findings worth calling out

- **The ratchet had a blind spot.** `test_no_undeclared_participant_status_validator`
  flagged a module only when it *both* named a member predicate *and* built a
  dimension. A writer validating nothing names no predicate, so it was never
  flagged — the detector caught partial validators and missed wholly-unvalidated
  ones. Three such writers were live: `_get_or_create_accepted_status`,
  `_build_owner_initial_status`, `_build_bootstrap_statuses`.
- **A fourth writer escapes from the other direction.** The wire twin
  `as_CaseParticipant.append_rm_state()` *does* name a predicate, but builds
  `as_ParticipantStatus` from flat fields rather than a dimension object, so the
  construction half of the gate never matches — and it appears in neither
  declaration list. Widening the gate to construction alone does not reach it
  either; the wire shape needs its own trigger.
- **CONCERN-3111 §2 was wrong, and this PR says so.** It concluded the report
  phase is "a deliberate separate lifecycle" with "no case participant".
  `_build_owner_initial_status()` reuses the marker's *id* as the participant's
  first ladder rung, and `_get_or_create_accepted_status()` mutates the stored
  record in place — one lifecycle at two stages, not two lifecycles.
- **The "unmeasured blast radius" is measured: zero on legal data, for the
  RM-coupled rules.** RM↔VF and RM↔D are bounded by reachability — `VALID` is
  reachable only from `RECEIVED`/`INVALID`, `INVALID` only from `RECEIVED`, and a
  fix-ready `vf` requires having passed `ACCEPTED`, which RM never walks back
  from. VF↔D is RM-independent and so is *not* bounded that way; it refuses
  `(vf, D)`/`(Vf, D)`, which the D machine alone permits and which are corrupt
  regardless.
- **`received/case/validate.py` is dead code.** All three classes have no
  production callers; the registry routes the similarly-named
  `CloseCaseReceivedUseCase` from `lifecycle.py` instead.
- **`update_participant_rm_state()` is an unseen BT-audit-trail bypass.** It
  calls `dl.save()`, and `test_no_dl_mutations_in_execute.py` only walks
  `execute()` bodies without following into helpers.

## Validation

Docs-only — no Python changed, no test suite ran.

- `spec-lint`: 0 errors (rationale-length warnings on the new BTND-10 entries
  match the existing BTND-10-001/002 entries)
- `markdownlint-cli2`: 0 issues
- notes frontmatter, ADR frontmatter, ADR index/nav sync: pass
- `mkdocs build --strict`: clean

## Implementation Issues

Chained A → B → C; each is blocked by the previous.

- #3204 — A: make `CreateParticipantStatusNode` the only writer, reached as a tree child (`size:L`)
- #3206 — B: move pre-case RM state to `ReportCaseLink`, delete the marker (`size:L`)
- #3207 — C: delete the `append_rm_state` mutators and the dead module, close the ratchet (`size:L`)


sei-ahouseholder (commented):
## PR Triage — #3203

**Merge state:** `MERGEABLE` / `CLEAN` — base `main`, not draft
**CI:** passing (8/8)
**Scope:** docs-only — `Closes #3111`; implementation chained as #3204 → #3206 → #3207

Findings: **4 FAIL, 6 IMPROVE**, all `fix-now`. No new-issue excursions, no premise inversions.

The ADR is unusually claim-dense — it asserts counts, guard conditions and ratchet
behaviour about live code — so triage fact-checked each claim against `HEAD`. Most hold.
The ones below do not, and two of them undercut arguments the decision rests on.

---

### FAIL

**1. The blast-radius guard is wrong for one of the three entailments — and self-contradictory**
`docs/adr/0089-one-participant-status-writer.md:163`

> The entailments fire only when the F bit or the D bit is set *and* RM is not in `{ACCEPTED, DEFERRED, CLOSED}`.

That guard belongs only to the two RM-coupled rules (`violation_rm_vf_entailment`,
`composite_state_invariants.py:107-110`; `violation_rm_d_entailment`, `:133-136`; both
gated on `RM_STATES_CONSISTENT_WITH_FIX`, `:89-91`). The third,
`violation_vf_d_entailment` (CSB-17-001, `:146-175`), takes only `(vf, d)` — it never
reads `rm` — and `composite_state_violations()` calls it unconditionally (`:261`).

So entailments *do* fire at RM in `{ACCEPTED, DEFERRED, CLOSED}`: exactly `(vf, D)` and
`(Vf, D)`. That makes the next sentence — "Three of the converted sites write `ACCEPTED`,
`DEFERRED` or `CLOSED`, **so they cannot fire**" — false, and it is refuted by the ADR's
own following paragraph, which names `(vf, D)`/`(Vf, D)` as the combinations that remain
refusable.

The *conclusion* survives: those states are already corrupt, so refusing them is a fix,
and the answer is still zero on legal data. Only the reasoning needs repair — scope the
guard sentence to the two RM-coupled entailments, say VF↔D is RM-independent, and drop the
"so they cannot fire" clause. Same sentence is mirrored in `notes/domain-validation.md`.

**2. The wire writer is marked "declared" in the ratchet; it is neither declared nor detected**
`docs/adr/0089-one-participant-status-writer.md:25`

The table row for `as_CaseParticipant.append_rm_state()` says "yes, declared."
`test/architecture/test_participant_status_validation.py` contains no reference to `wire`
anywhere — `_VALIDATING_NODE_MODULES` holds 2 core paths, `_DECLARED_EXCLUSIONS` holds 4
(`rm_transitions.py`, `_adjudication.py`, `case_participant.py`,
`participant_status_effect.py`). The wire twin is in neither.

It is not *detected* either: it names `is_valid_rm_transition` (`:214`) but builds
`as_ParticipantStatus` with flat fields, so `constructs_dimension` is `False`. And it
escapes the **widened** gate too — an AST sweep for the four `_DIMENSION_CONSTRUCTORS`
over `vultron/` returns 15 modules, and that file is not among them.

That falsifies two load-bearing claims: "The ratchet must be able to see every writer"
(Decision Drivers) and "Validating less no longer buys invisibility". The widened trigger
set needs the wire projection's construction shape, or #3207 closes a ratchet that still
has a hole. Same row is wrong in the `notes/domain-validation.md` table.

**3. The failure-swallowing shape is cited to the wrong notes file, in three places**
`specs/behavior-tree-node-design.yaml` (BTND-10-004 rationale), ADR `:125`, `notes/domain-validation.md`

All three cite `notes/bt-integration.md`. That file does not document it — no match for
`swallow`, `try/except` or `nested`. The guidance is in `notes/bt-pitfalls.md` § *Always
Check `BTBridge.execute_with_setup` Return Value* (ISSUE-1325, `:462-480`, with the
"subtree failure is silently swallowed" example at `:472`).

The spec-rationale occurrence matters most: an agent implementing #3204 follows that
pointer to a file with nothing there. `domain-validation.md` already lists
`bt-pitfalls.md` in `related_notes`.

**4. `specs/report-management.yaml` does not exist**
`docs/adr/0089-one-participant-status-writer.md:266`

> Generated spec requirements: … `specs/report-management.yaml` RM-pre-case entry for `ReportCaseLink`.

A repo-wide grep finds that filename only on this line. No such requirement was added —
BTND-10-006 already carries the relocation. Drop the clause. (If a report-scoped
requirement is genuinely wanted, `specs/rm-behavior.yaml` or
`specs/participant-case-replica.yaml` — which describes `ReportCaseLink` at `:292` — are
the existing homes, but that is a scope call, not a correction.)

---

### IMPROVE

**5. "The exclusion list can only shrink" — it grows to ~13 first** (ADR `:151`)
The widened gate matches **15** modules; minus the 2 validating nodes, the exclusion list
goes from 4 today to ~13 at landing before being driven to 2. Several newly-caught modules
are not writers at all — `develop_fix_conditions.py`, `status/nodes/case_status.py`,
`cs_dimension_filter.py`, `demo/helpers/seeding.py`, `wire/as2/extractor/_builders.py`,
`wire/.../vocab/objects/case_status.py`. The construction-only gate over-catches read-side
and projection code, which #3207 hits immediately.

**6. try/except attributed to `deploy_fix.py` alone; `develop_fix.py` does it twice**
`develop_fix.py` wraps *both* nested `update()` calls in `except Exception` (`:126-128`,
`:141-155`) via the shared `_make_status_node()` helper (`:87`). `leave.py` and
`close_case_effect.py` contain zero `except` statements. So the shape is at 2 of 4 files
and 3 of the nested calls, not 1. Related: `develop_fix.py`'s one construction feeds two
call sites, so 5 constructions produce 6 nested `update()` calls — the enumeration reads
as one-per-file-except-`leave.py`.

**7. BTND-10-002's `adr:` field omits ADR-0089**
The rationale now opens "Amended by ADR-0089" but `adr:` still lists only `ADR-0086`.
Precedent in the same corpus: RSH-07-001/RSH-07-002, amended by ADR-0080, carry
`adr: [ADR-0076, ADR-0080]`. The structured field is what `load-specs` exposes, so an
agent reading BTND-10-002 will not find the decision that rewrote it. BTND-10-004/005/006
are correct.

**8. The import-cycle fix is not "trivially breakable" as written** (`notes/domain-validation.md`)
Both premises hold — the import at `predicates/participants.py:26` is used only by the
annotation at `:57`, and a `TYPE_CHECKING` block exists at `:30-31`. But the module has no
`from __future__ import annotations` and `:57` is unquoted, so moving the import alone
raises `NameError`. Add the missing step: quote the annotation, or add the future import.

**9. The "seven writers" census is unscoped** (ADR `:21`)
Production constructs a `ParticipantStatus` at four further sites:
`case_participant.py:178` (`_init_participant_status_if_empty`), `:537` and `:586` (the two
byte-identical `_set_accepted_status` copies, on `ReporterParticipant` and
`FinderReporterParticipant`), plus `demo/helpers/seeding.py:1161` and
`wire/as2/extractor/_builders.py:517`. `domain-validation.md` acknowledges the model-layer
seeding validators and says ADR-0089's work de-duplicates them — but the ADR never mentions
`_set_accepted_status`, so a reader of the ADR alone gets a count that does not match the
code, and #3204's "only writer" goal is ambiguous about them. Scope the table, or add rows.

**10. PR body: docs-only template deviations**
The guide's docs-only rules say "No Verification section — no Python was changed, no test
suite ran." One is present, and `Black, flake8, mypy, pyright: clean` is vacuous for a PR
touching zero `.py` files — though the spec-lint / markdownlint / mkdocs lines are genuinely
useful, so folding them into Changes beats deleting them. Also the attribution line sits
mid-body, before `## Implementation Issues`. Closing reference placement is correct.

---

### Verified clean

- Issue linkage: `Closes #3111` matches scope; #3204/#3206/#3207 are Task type, chained
  `blockedBy` 3204←3206←3207, parented to epic #2684, all on Project #24
- ADR frontmatter, index registration, mkdocs nav; `docs/adr/**` is out of lint-docs scope
- `archived/README.md` correctly drops the stale "currently empty" claim (MS-16-001);
  ADR-0015 is present with `superseded_by: ADR-0041`; the index section it points to exists
- Notes frontmatter valid; `related_specs`/`related_notes` updates all resolve
- `spec-lint` exits 0; the three rationale-length warnings match pre-existing BTND-10-001/002
- Claims fact-checked and **confirmed**: the ratchet blind-spot mechanism; the six-node
  bypass; ~35 `append_rm_state()` test call sites; `received/case/validate.py` dead with 3
  classes and the registry routing `CloseCaseReceivedUseCase` from `lifecycle.py`;
  byte-identical `_set_accepted_status`; six deterministic ids; the
  `sender_actor_id or actor_id` fallback; RM reachability (`VALID` from
  `RECEIVED`/`INVALID`, `INVALID` from `RECEIVED`, `RECEIVED` from `START`, no edge back);
  `update_participant_rm_state()` → `dl.save()` with the detector walking `execute()` only;
  all seven named writers exist and construct a status; the three "invisible" modules name
  none of the `_COMPOSED_PREDICATES`

Artifact: `.claude/pr-3203-triage.json`

sei-ahouseholder (commented):
## PR Execute — #3203

All **10** triage findings fixed. No issues filed, no gates triggered — every
finding was first-order and `fix-now`.

**Commits:** `4bb38b436` (fixes) · `8677bc291` (sync with main) · `3ea83e14a` (post-merge correction)
**CI:** 8/8 passing · **Merge state:** `MERGEABLE` / `CLEAN`

---

### FAIL

**1. Entailment guard split by rule** — `ADR-0089`, `notes/domain-validation.md`, history entry
RM↔VF and RM↔D (CSB-18-001) keep the `{ACCEPTED, DEFERRED, CLOSED}` guard. VF↔D
(CSB-17-001) is now stated as RM-independent, since `violation_vf_d_entailment()`
takes only `(vf, d)` and `composite_state_violations()` calls it unconditionally.
Dropped the false "so they cannot fire" clause and re-based the paragraph on the
reachability bound, which only ever applied to the RM-coupled pair. **The
conclusion — zero on legal data — is unchanged; only the reasoning was wrong.**

Worth recording: the first rewrite of that passage asserted "no legal sequence
produces `(vf, D)`: reaching `D` requires passing `VF`." Checking
`is_valid_d_transition()` showed the D machine permits `d → D` without consulting
`vf` — the VF↔D entailment is the *only* thing forbidding it, so the replacement
claim was circular. The text now says that instead, which strengthens the case for
enforcing the rule on this path. Added the transferable lesson to the notes: scope
a reachability bound to the rules that actually read the dimension you bound on.

**2. Wire twin corrected to undetected, and the count to four** — ADR, notes, `BTND-10-002`, history
`as_CaseParticipant.append_rm_state()` now reads "no — undeclared *and*
undetected", sits with the other invisible writers, and the headline says four
invisible writers, not three. Both the ADR and the notes now explain that it fails
the gate from the opposite direction — it *does* name `is_valid_rm_transition`,
but builds `as_ParticipantStatus` from flat fields, so the construction half never
matches — and that widening the gate to construction alone does not reach it.
BTND-10-002's rationale carries the caveat too, so the requirement travels with it.

**3. Swallowing-shape citation repointed** — ADR, notes, `BTND-10-004`
All three now cite `notes/bt-pitfalls.md` § *Always Check
`BTBridge.execute_with_setup` Return Value*. The two remaining
`bt-integration.md` references in these files are pre-existing and on unrelated
topics.

**4. Dangling spec file removed** — ADR
The `specs/report-management.yaml` clause is gone; the sentence points at
BTND-10-006, which carries the relocation. A repo-wide grep for the filename now
returns nothing.

### IMPROVE

**5. Exclusion list grows before it shrinks** — ADR
Replaced "can only shrink" with the measured shape: the construction-only gate
matches 15 modules, so the list goes from four entries to ~13 at landing before
being driven to two writer exclusions. Named the six over-caught read-side and
projection modules, noted those declarations are permanent and each needs a
reason, and recorded the narrower assignment-keyed alternative plus why it was
rejected (it reintroduces a dodgeable structural property). Count re-confirmed
against the post-merge tree.

**6. try/except attribution** — ADR, notes, `BTND-10-004`, history
Both `deploy_fix.py` and `develop_fix.py` (the latter on both calls). Also records
that five construction sites yield six nested `update()` calls, because
`develop_fix.py` ticks one helper-built node from two places.

**7. `BTND-10-002` gains `ADR-0089`** — now `adr: [ADR-0086, ADR-0089]`, matching
the RSH-07-001/002 precedent for an amended requirement.

**8. Annotation fix is two steps** — notes, history
Move the import under `TYPE_CHECKING` *and* quote the annotations; the module has
no `from __future__ import annotations`, so moving alone raises `NameError`.

**9. Writer census scoped** — ADR, notes
"Writer" is now defined as a site that appends a ladder rung or writes the marker
record, with the excluded constructions named and reasoned: the constructor-seeding
validators, the demo seeder, the wire→core extractor. The ADR now also names the
byte-identical `_set_accepted_status` pair as the ARCH-15-004 / CS-22-001 problem
the work de-duplicates — previously claimed only in the notes.

**10. PR body** — `Verification` → `Validation` with an explicit docs-only line,
vacuous Python-linter bullet dropped, attribution moved to the end. The Summary and
Findings sections were also updated so the body no longer states what the diff now
contradicts.

---

### CI loop — 1 iteration

- **Sync**: `origin/main` had advanced 33 commits; merged cleanly. It touched
  `vultron/core/predicates/participants.py`, which finding 8 describes — that
  claim was re-verified against the merged tree and had already gone stale (main
  added a second unquoted annotation), so the wording was made
  count-independent in `3ea83e14a`.
- **Tests**: `uv run pytest -q` reached 100% with 0 failed, 0 errors, 0 XPASS
  (xfail ratchet satisfied).
- **Push**: fast-forward `05e108d48..3ea83e14a`. The branch had no upstream
  tracking ref, so the first attempt was rejected with a `--set-upstream` hint;
  confirmed the remote head was an ancestor of HEAD before pushing, so this was
  an ordinary fast-forward, not a force-push.
- **CI**: 8/8 green — CodeQL, lint, Spec Lint, Spec Corpus Tests, Codebase Docs
  Path Ratchet, Analyze (actions), Analyze (python), docs-build-check.

Local validation: `spec-lint` exit 0 / 0 errors · `markdownlint-cli2` 0 issues ·
pre-commit hooks (notes + ADR frontmatter, ADR index/nav sync, flake8, spec
registry) all passed.

Artifact: `.claude/pr-3203-execute.json`

sei-ahouseholder (commented):
## PR Verify — #3203

**Overall verdict: READY-TO-MERGE**

**Merge state:** `MERGEABLE` / `CLEAN` — base `main`, not draft, no `needs-rebase` label
**CI:** 8/8 passing
**Findings:** 10/10 `CONFIRMED` at HEAD (4 FAIL, 6 IMPROVE) — 0 unresolved, 0 skipped, 0 deferred

Every fix was checked against the working tree at HEAD, not just against the
commit diff. All three commits (`4bb38b436`, `8677bc291`, `3ea83e14a`) are
ancestors of HEAD. No conflict markers anywhere in the tree.

| # | Finding | Sev | Verdict | Evidence at HEAD |
|---|---|---|---|---|
| 1 | Entailment guard scoped to the RM-coupled rules | FAIL | ✅ CONFIRMED | `RM-coupled` present in ADR, notes and history; `"so they cannot fire"` gone from the tree |
| 2 | Wire twin corrected to undetected; count → four | FAIL | ✅ CONFIRMED | ADR `:26` and notes `:451` both read `**no**`; `"yes, declared"` row gone; BTND-10-002 carries the `as_ParticipantStatus` caveat |
| 3 | Swallowing shape cited to `bt-pitfalls.md` | FAIL | ✅ CONFIRMED | Present in all three targets; no `bt-integration.md` citation remains for this claim |
| 4 | Dangling `specs/report-management.yaml` removed | FAIL | ✅ CONFIRMED | Repo-wide grep returns nothing |
| 5 | Exclusion list grows to ~13 before shrinking | IMPROVE | ✅ CONFIRMED | Measured shape in ADR; `"can only shrink"` gone (the one remaining hit is pre-existing ADR-0034, unrelated) |
| 6 | try/except attributed to both files | IMPROVE | ✅ CONFIRMED | `deploy_fix.py` + `develop_fix.py` in ADR, notes `:518`, BTND-10-004, history `:95` |
| 7 | `BTND-10-002` gains `ADR-0089` | IMPROVE | ✅ CONFIRMED | `adr: ['ADR-0086', 'ADR-0089']` on re-parse |
| 8 | Annotation fix is two steps | IMPROVE | ✅ CONFIRMED | `__future__` caveat in notes and history |
| 9 | Writer census scoped | IMPROVE | ✅ CONFIRMED | `_set_accepted_status` now named in the ADR (2 refs) and notes (3) |
| 10 | PR body docs-only template | IMPROVE | ✅ CONFIRMED | `Validation` section, no Python-linter line, attribution at end |

---

### Disclosure — a gate I had to clear on my own work

Execute wrote its `merge_state` block with the wrong field names (`mergeable` /
`merge_state_status` / `is_draft` instead of the schema's `synced` /
`sync_commit_ref` / `mergeable_after_sync` / …), putting the sync details in
`ci_loop.sync` instead. Verify's Phase 1 treats a missing `synced` as
`UNSYNCED-EXECUTE`, which blocks `READY-TO-MERGE`.

That was a bookkeeping defect in my own execute run, not a problem with the
branch, so I corrected the block to the documented schema rather than bouncing the
PR. Flagging it because correcting an artifact to make a gate pass is exactly the
move that gate exists to catch. What actually cleared it is independent of the
edit: merge commit `8677bc291` is on the branch, `origin/main` is an ancestor of
HEAD, and Phase 2 re-checked mergeability **live** (`MERGEABLE` / `CLEAN`) rather
than reading any artifact field.

### Note on the substance

The two headline FAILs were both in the ADR's *reasoning*, not its conclusion. The
blast-radius claim was mis-scoped — VF↔D is RM-independent, so "RM is `ACCEPTED`"
never conferred the immunity the ADR claimed — and the ADR contradicted itself one
paragraph later. The wire twin was recorded as visible to the ratchet when it is
invisible to both the current gate and the proposed replacement. Both would have
misled the implementer of #3204/#3207; neither changes the decision.

One correction was self-inflicted and caught in the same pass: the first rewrite of
the entailment paragraph asserted that reaching `D` requires passing `VF`, which is
circular — `is_valid_d_transition()` permits `d → D` without consulting `vf`, and
the VF↔D entailment is the only thing forbidding it. The final text says that
instead.

---

Artifacts cleaned up. PR #3203 is **READY-TO-MERGE**.

Issue #3202, opened 2026-09-14
Consolidate datetime helpers into core; fix validate_datetime naive-UTC gap


sei-ahouseholder:
## Summary

`validate_datetime` on `as_Object` (the base wire type for all AS2 objects)
calls `datetime.fromisoformat(value)` on inbound ISO strings without
normalizing the result to UTC. A string like `"2026-01-01T00:00:00"` (no
offset) produces a naive `datetime` that crosses the wire edge into core,
violating CS-13-001 (MUST be timezone-aware UTC) and ADR-0032
(validate at the edge, promote to strict core types).

Additionally, `now_utc()` and `days_from_now_utc()` in
`vultron/wire/as2/vocab/base/dt_utils.py` duplicate `_now_utc()` already
present in `vultron/core/models/_helpers.py`. Consolidating into core
eliminates the duplication and makes a single public UTC helper available to
both layers.

## Acceptance Criteria

- [ ] AC-1: `now_utc()` and `days_from_now_utc()` are moved from
  `vultron/wire/as2/vocab/base/dt_utils.py` into
  `vultron/core/models/_helpers.py`; the private `_now_utc()` is replaced
  by (or aliased to) the public `now_utc()`; `as_utc()` is the public name
  for the current private `_as_utc()`
- [ ] AC-2: `dt_utils.py` is deleted — no backwards-compat shim; all callers
  (`base.py`, `embargo_event.py`, affected tests) import from core helpers
- [ ] AC-3: `validate_datetime` on `as_Object` calls `as_utc()` after parsing
  the ISO string, so offset-less inputs emerge as UTC-aware datetimes at the
  wire edge
- [ ] AC-4: A test in `test/wire/as2/vocab/base/` covers naive ISO string ->
  timezone-aware UTC on `as_Object` wire objects (e.g. `published`,
  `updated`, `start_time`, `end_time`)
- [ ] AC-5: `as_utc()` docstring in `_helpers.py` is updated to note that the
  wire edge now normalizes upstream (this is defensive code for non-wire
  callers, not the primary normalization site)
- [ ] AC-6: All existing tests pass; black, flake8, mypy, pyright clean

## Prior Art

- `vultron/core/models/_helpers.py` — `_as_utc()` (private), `_now_utc()`
  (private), `status_recency_key()`: the existing per-consumer workaround
  this fix makes proactive
- `vultron/wire/as2/vocab/base/dt_utils.py` — `now_utc()`, `days_from_now_utc()`:
[...]

Issue #3054, opened 2026-09-02
EH-07-001 is corpus-wide but only one boundary complies — no audit for atomic-reject-but-first-error-only validation


sei-ahouseholder:
## Summary

`EH-07-001` (added by PR #3049) is written as a general property:

> A validation boundary that rejects its input as a unit MUST report every
> violation it can recognise in that input, not only the first.

#3050 brings exactly **one** boundary into compliance — the `ParticipantStatus`
write path. Every other atomic-reject-but-first-error-only site in the codebase
becomes a MUST-level violation the moment #3049 merges, and nothing exists to
find them.

## Surface Symptom vs. Underlying Problem

**Surface reading**: "#2112 was one node reporting one error; #3050 fixes it,
done."

**Underlying problem**: the *rule adopted to fix it* has corpus-wide scope while
its enforcement has single-site scope. This is the same shape as the ISSUE-2906
lesson — a predicate that looks complete at each call site, where the gap is
only visible when you diff the sites against the rule — except here the rule is
a spec entry rather than a function, and the "call sites" are every validation
boundary in the system.

**What is already correct and should be left alone**: EH-07-001's *content*, the
root/derived classification (EH-07-002), the structured-exception shape
(EH-07-003), and the trigger/receive disposition asymmetry recorded in ADR-0084.
None of that is in question. The question is scope of enforcement.

## Category

Spec/implementation compliance gap; missing audit.

## Severity

Medium. Nothing regresses — these boundaries behave exactly as they did before
#3049. But a MUST-level spec entry that is knowingly unenforced across most of
its scope corrodes the corpus: agents cite EH-07-001 as settled fact while the
code mostly does not honour it, which is precisely the "stale premise" failure
mode `decision-audit` exists to catch.

## Evidence

Boundaries with multiple *independent* checks behind a single atomic rejection,
outside #3050's scope:

- **`vultron/core/models/staged_case.py`** — `Case._check_case_invariants`
  (line ~99) runs three independent checks (at least one vulnerability report;
  ≥2 case participants; at least one materialized `CaseStatus`) and raises on
  the first. A `Case` failing all three reports one. `IncomingReport` and
  `EmbargoedCase` in the same file have the same shape — 8 raise sites total.
- **`vultron/core/behaviors/`** — 18 files carry ≥8 `return Status.FAILURE`
  sites. `case_proposal_received_tree.py` has 24,
  `case/nodes/participant/participant_add.py` 12,
[...]

PR #3233, opened 2026-09-14
fix(wire): read inbound field values, not just keys, and refuse malformed inline objects

### LINKED ISSUE #3217: wire: published:"" bypasses VultronParseMissingPublishedError, leaks Pydantic ValidationError
## Summary

`vultron/wire/as2/parser.py:148` guards against a missing `published` field with:

```python
if body.get('published') is None:
    raise VultronParseMissingPublishedError(...)
```

An empty-string value (`published: ""`) makes `body.get('published') is None` **False**, so the guard passes.  `model_validate` then tries to parse `""` as a datetime and raises a Pydantic `ValidationError`.  Callers that catch `VultronParseMissingPublishedError` to return a controlled 422/400 get an unhandled `ValidationError` instead, producing a 500 or incorrect error code.

## Fix

Change the guard to a falsy check:

```python
if not body.get('published'):
    raise VultronParseMissingPublishedError(...)
```

## Discovery

Surfaced during code review of #3189 (ADR-0088 single neutral role-authority resolver).  Pre-existing in `vultron/wire/as2/parser.py:148`.

### LINKED ISSUE #3232: wire: malformed inline object is silently downgraded to as_Link, and inline actors are flattened by a wire/core type-resolution fault
## Summary

`vultron/wire/as2/parser.py::_expand_inline_value` typed nested inline dicts,
then swallowed any validation failure with `except Exception: return expanded`.
The parent field then accepted the raw dict as a bare `as_Link`, so the whole
activity parsed **successfully** with the nested object's subtype erased —
extracting as `UNKNOWN`, losing the case id, and drawing a 202 for a message
the receiver never understood. No error, no log line.

Instrumenting the fallback and running the full suite showed it fired 52 times,
all from a single cause: `find_in_vocabulary` falls back to `CORE_TYPE_MAP`
(ARCH-12-003), and `OrderedCollection` is registered **only** there. An inline
actor's `inbox`/`outbox` therefore expanded to a **core** `CoreActorCollection`,
which `as_VultronOrganization.inbox` rejects — flattening the entire actor to an
`as_Link` on every inbound activity carrying an inline actor. That is a wire/core
layering fault (ARCH-22-001), not an input problem, and the swallow concealed it.

## Reproduction (before the fix)

```text
Accept(Invite(...)) with a valid inner Invite      -> as_Invite, accept_invite_actor_to_case
Accept(Invite(...)) with inner published: ""       -> as_Link,   UNKNOWN, case_id lost
Create with inline Organization actor carrying
  OrderedCollection inbox/outbox                   -> actor flattened to as_Link
```

## Fix

Two changes, which must ship together — refusal is only safe once the layering
fault is gone, or all 52 measured mismatches would become hard rejections:

1. `_inline_vocab_class` resolves **wire-branch classes only** (`as_Base`
   subclasses). An unregistered or non-wire type resolves to `None` and the dict
   is left for the parent to validate — the same outcome the `except` produced,
   now reached by construction.
2. A recognised inline object that fails its own class's validation raises
   `VultronParseValidationError` instead of being downgraded (ADR-0032).

## Discovery

Sibling finding from the investigation of #3217 (`published: ""` bypassing the
missing-published guard). Same trigger, same inbound parse path, strictly worse
consequence. Recorded per ADR-0090 and spec MV-04-003.

sei-ahouseholder:
- Closes #3232 — the inline-object downgrade this investigation surfaced, and the wire/core type-resolution fault behind it

## Summary

`parse_activity` guarded two required fields with `is None`, which answers "was
the key absent?" rather than "did the sender supply a value?". Both guards now
read the value, and a nested object that fails validation is refused instead of
being silently downgraded to a weaker type.

**Supersedes `723ff08eb` on `main`.** While this branch was in flight, `main`
landed the falsy one-liner `if not body.get("published")` that #3217 proposed, and
closed #3217. ADR-0090 evaluates that as option 1 and rejects it: a falsy check
also absorbs `0`, `[]`, `{}` and `false`, reporting a corrupt timestamp as a
missing one and telling the sender to resend what they already sent. This PR keeps
the narrower `is_blank` predicate and the merge resolution discards main's line, so
`published: 0` and `published: []` go back to being schema faults. #3217 stays
closed — it is not re-listed above.

On the reported symptom: a 500 does happen, but not for `published`. Both
spellings of a missing `published` already returned 422, so that defect is
diagnostic — the sender got a Pydantic isoformat dump instead of the CLP-15-006
explanation. The 500 is on `type`: an unhashable value reaches
`find_in_vocabulary`'s dict lookup and raises `TypeError`, which nothing maps.
That is fixed here too. The sibling scan is where the rest of the damage was.

## Changes

- **`vultron/wire/as2/parser.py`**: new `_is_blank` predicate (absent, `null`, or
  whitespace-only, matching the project's canonical `not v.strip()` test) now
  guards both `type` and `published`. Deliberately narrower than a falsy check —
  `0` and `[]` are malformed, not omitted, and stay on the schema path.
  - Fixes the reported `published: ""` misdiagnosis, and the whitespace-only
    spelling the issue's proposed one-liner would still have admitted.
  - Fixes a sibling status-code split: `type: ""` drew a **422** while an
    omitted `type` draws a **400**. Two spellings of one omission, two answers.
- **`vultron/wire/as2/parser.py`**: `_inline_vocab_class` resolves **wire-branch
  classes only**, and a recognised inline object that fails its own class's
  validation now raises `VultronParseValidationError` (ADR-0032). These ship
  together — refusal is only safe once the layering fault is gone.
- **`vultron/wire/as2/vocab/base/objects/base.py`**: the shared timestamp
  validator (all four fields, so CS-08-002-compliant) reads a blank string as
  absence. Composed with #3202's UTC normalization, which landed in the same
  method while this was in flight.
- **`docs/adr/0090-…`**, **`specs/message-validation.yaml`** (MV-03-002,
  MV-04-003): the decision and its two recurring requirements.

### Why the excursion was in scope

`_expand_inline_value` swallowed inline validation failures and handed the parent
the raw dict, which the parent accepted as a bare `as_Link` — so an
`Accept(Invite(...))` with one corrupt inner field parsed **clean**, extracted as
`UNKNOWN`, lost its case id, and drew a **202** for a message the receiver never
understood. No error, no log line.

Instrumenting that fallback and running the whole suite showed it fired 52
times, all from one cause: `find_in_vocabulary` falls back to `CORE_TYPE_MAP`
(ARCH-12-003), where `OrderedCollection` alone is registered, so an inline
actor's `inbox` expanded to a **core** `CoreActorCollection` that
`as_VultronOrganization.inbox` rejects — flattening **every inline actor** to an
`as_Link` on the inbound path, with no empty string involved. The fallback was
not protecting a legitimate case; it was concealing an ARCH-22-001 layering
fault. That measurement is also the evidence that refusing is safe.

### Fixes applied during PR triage

- **`vultron/wire/as2/parser.py`**: a non-string `type` is now rejected as
  `UnknownType` (422) before the vocabulary lookup. `find_in_vocabulary` resolves by
  dict membership, so `type: []` or `type: {}` raised `TypeError` — not the
  `KeyError` the local handler catches, and not mapped by the inbox adapter — so it
  escaped `parse_activity` and drew a **500**. That is the unhandled-exception
  symptom #3217 was filed about, in a sibling spelling of the same field.
- **`vultron/wire/as2/parser.py`**: a refusal now names the field path
  (`Inline as_Invite object at 'object' is malformed: ...`). Naming only the
  innermost class left a sender unable to locate the fault when two objects of the
  same type were present.
- **`vultron/wire/as2/vocab/base/utils.py`**: `is_blank` is extracted here as the
  wire layer's single definition of blank, replacing a copy in `parser.py` and an
  inline `not value.strip()` in the timestamp validator (CS-22-001). The wire layer
  cannot import the core predicate (ARCH-22-001), so it needs its own.
- **`specs/message-validation.yaml`**: MV-10-001/MV-10-002 are narrowed to
  `VulnerabilityCase`. Read generically they licensed a partial inline object of any
  type — exactly the shape MV-04-003 now refuses — so the two collided for every
  type except the one with a real stub class. Partial inline objects of other types
  were never designed; they survived only on the fallback this PR removes.
- **`docs/adr/0090-…`**: records that stub consequence, drops a hardcoded test count
  that had already drifted, narrows the overbroad "wire trees are wire-to-wire"
  driver (`as_ObjectRef` does admit `CoreObject`), and corrects the Validation
  section — a green suite shows the fallback protected nothing *Vultron itself
  emits*, which is what the 52 instrumented hits measured, and cannot speak to
  shapes only an external sender produces.
- **`notes/wire-core-boundary.md`**, **`notes/vocabulary-registry.md`**: record the
  wire-only-resolution rule. It was living in a docstring; both notes are the
  designated homes and both presented the `CORE_TYPE_MAP` fallback with no
  caller-side caveat.
- **`notes/stub-objects.md`**, **`notes/README.md`**: rewritten for the narrowing —
  why `VulnerabilityCase` earns the exception, and that the dict-stub form is a
  workaround for a missing capability rather than a pattern to extend.
- **`AGENTS.md`**, **`notes/case-ledger-authority.md`**: widen the
  missing-`published` guidance to cover a blank value, matching the `docs/` edit in
  `d6f544a80`, and note that the top-level guard does not cover nested objects.
- **`test/wire/as2/test_parser.py`**, **`test_inbox.py`**,
  **`test_base_timestamps.py`**: tests for every non-string `type` spelling
  (including the unhashable pair), the field path in a refusal, core-only inline
  types, and the blank-vs-omitted distinction. Also corrects a docstring that
  claimed a blank timestamp "lets the field default apply, exactly as it does when
  the key is left out altogether" — it does not: `published`/`updated` carry
  `default_factory=now_utc`, so omission fabricates a timestamp while a blank
  yields `None`.

## Filed, not fixed here

Findings from the upward-reflection checklist, each routed to an issue rather
than folded into this PR (BW-07-009):

- **#3241** — `docs/reference/inbox_handler.md` cites eight retired module paths
  (`vultron.api.v2.*`, `vultron.behavior_dispatcher.*`) as the current API, all
  checked off as done. Verified by import.
- **#3242** — the AS2 collection types are registered in neither wire registry,
  which is *why* `find_in_vocabulary` reached the core fallback. This PR removes
  the parser's exposure to that trap; the registry gap itself remains.
- **#3243** — per-requirement spec anchors render into the site but fail
  `mkdocs --strict`, so the docs edit here had to link the group anchor
  (`#mv-03`) rather than `#mv-03-002`.
- **#3257** — a nested object with no claimed `published` gets the receiver's clock
  as its `published`. Pre-existing and present on `main`: `as_Object.published`
  carries `default_factory=now_utc`, so an *omitted* nested timestamp already
  fabricates, and `_get_timestamp`'s `else _core_now_utc()` fallback does the same
  for `None`. This PR routes blank to that same outcome, which is the intended
  "blank is absence" semantics rather than a new fault, but the underlying
  fabrication needs its own fix.
- **#3258** — Vultron cannot dereference a URI object reference from another actor,
  which is *why* senders inlined partial objects. Removing the fallback exposes the
  gap rather than creating it.

Two learning files are committed on this branch: a `theme-candidate` on measuring
what a permissive fallback absorbs before judging it load-bearing, and a
`spec-ambiguity` on MV-01-005/006/007 naming no registry for "unrecognized
activity type". The branch also carries a one-paragraph fix to the `bugfix`
skill's finalize ordering, which listed `archive-history` before the PR that
skill needs the URL of.

## Verification

- Unit suite: **8635 passed**, 399 skipped, 19 xfailed, 5557 subtests passed.
- Integration suite: **1258 passed**, 2 xfailed.
- Full suite (`-m ""`) was also run green before the branch was freshened.
- Black, flake8, mypy, pyright clean; `spec-lint` clean for both new entries.
- 42 new tests, written failing first: every blank spelling of `type` and
  `published`; the non-blank-garbage boundary (`"not-a-date"`, `0`, `[]`, `{}`)
  still reported as a schema fault; blank `type` pinned to HTTP 400; nested
  subtype survival across a blank inner timestamp; refusal of a malformed nested
  object; inline-actor subtype preserved with core-only collection types; and the
  shared timestamp validator across all four fields.

## Reviewer notes

- **Behavior change, deliberate**: a sender whose inline object is malformed in a
  way the receiver previously tolerated now gets a 422 instead of a 202. That
  tolerance *was* the silent misrouting, so this trades an invisible wrong answer
  for a visible rejection. ADR-0090 records the alternatives, including why
  refusing without the wire-only resolution fix would have broken all 52 cases.
- **Not folded in**: `parse_activity` still reports only the first violation
  rather than all of them (EH-07-001) — already tracked by #3054.
- `payloadSnapshot` is untouched: `_OPAQUE_PAYLOAD_KEYS` excludes it from
  expansion, so snapshot contents are never validated here (SYNC-13-004).


sei-ahouseholder (commented):
## PR Triage: #3233 — fix(wire): read inbound field values, not just keys, and refuse malformed inline objects

**Linked issues**: #3217 (already CLOSED as completed by `origin/main` `723ff08eb`) · #3232 (open, correctly closed by this PR)
**Changed files**: 13 files — wire/as2, wire/as2/vocab/base, docs, specs, plan/incoming/learnings, test
**CI status**: ✅ passing (all 25 checks; 118 tests in the touched files pass locally)
**Base branch**: `main`
**Needs integration tests**: yes (integration + demo CI green on this head)

---

### Findings

|---|---|---|---|---|
| phase8-nested-blank-published-fabricates-clock-0 | code-review | ❌ FAIL | Blank nested timestamp → `None` → extractor substitutes the receiver's clock as the sender's claim (CLP-15-006) | fix-now-file |
| phase12-merge-conflict-0 | merge-state | ❌ FAIL | `parser.py` conflicts with main, which shipped the falsy guard ADR-0090 rejected | fix-now |
| phase8-nonstring-type-500-0 | code-review | ❌ FAIL | `type: []` / `type: {}` → unhandled `TypeError` → HTTP 500 | fix-now |
| phase5-mv04-003-refuses-mv10-stubs-0 | spec-conformance | ❌ FAIL | New MV-04-003 refuses MV-10-001/002-legal stub objects for every type but `VulnerabilityCase` | fix-now |
| phase10-blank-default-docstring-contradiction-0 | test-coverage | ⚠️ IMPROVE | Test docstring claims the field default applies; blank actually yields `None` | fix-now |
| phase8-inline-vocab-strips-core-typed-refs-0 | code-review | ⚠️ IMPROVE | `issubclass(cls, as_Base)` also strips core-typed values from `as_ObjectRef` fields declared to accept them | fix-now |
| phase9-wire-core-boundary-note-stale-0 | notes-currency | ⚠️ IMPROVE | Wire-only-resolution rule missing from `wire-core-boundary.md` / `vocabulary-registry.md` | fix-now |
| phase9-published-refusal-narrow-in-notes-0 | notes-currency | ⚠️ IMPROVE | `AGENTS.md:264` and `case-ledger-authority.md:376` still say "missing"; the `docs/` twin was widened | fix-now |
| phase9-stub-objects-note-stale-0 | notes-currency | ⚠️ IMPROVE | `notes/stub-objects.md` does not record the MV-04-003 refusal interaction | fix-now |
| phase8-refusal-message-lacks-path-0 | code-review | ⚠️ IMPROVE | Refusal names only the innermost class, with no field path | fix-now |
| phase7-blank-predicate-duplicated-0 | agents-md-compliance | ⚠️ IMPROVE | "Blank" is defined in three places that must agree (CS-22-001) | fix-now |
| phase2-issue-3217-already-closed-0 | issue-linkage | ⚠️ IMPROVE | PR body predates main's competing fix; supersession is unstated | fix-now |
| phase6-adr-0090-count-drift-0 | adr-check | ⚠️ IMPROVE | ADR states "9867 tests"; PR body's numbers sum to 9893 | fix-now |

**Total**: 4 FAIL · 9 IMPROVE · 0 NEW-ISSUE

---

### The three that matter most

**1. The blank-is-absence change lets the receiver fabricate the sender's timestamp.**
`validate_datetime` now returns `None` for a blank string, and
`extractor/_builders.py::_get_timestamp` does
`return val if isinstance(val, datetime) else _core_now_utc()`. Verified end-to-end:

```text
Create + inline VulnerabilityReport with published: ""
  -> parses clean, subtype preserved, 202
  -> extracted report.published == receiver's wall clock
```

Pre-PR this was unreachable, because `datetime.fromisoformat("")` raised. This is the
same clock-substitution CLP-15-006 forbids and that the top-level guard exists to prevent.
`test_parse_activity_keeps_nested_subtype_when_nested_published_is_blank` exercises exactly
this shape but asserts only the subtype, so the fabrication goes unobserved.

Note the scope: the hazard is *partly pre-existing* — `as_Object.published` carries
`default_factory=now_utc`, so an **omitted** nested `published` already yields a
receiver-clock value. Fix the blank path here; file the broader hazard per BW-07-009.

**2. `type: []` still returns a 500 — the exact symptom #3217 was filed about.**
`_is_blank` returns False for non-str values, `cast(str, type_)` then lies, and
`find_in_vocabulary` does `item_name in WIRE_TYPE_MAP` — a dict membership test on an
unhashable value raises `TypeError`, which neither the local `except KeyError` nor
`_inbox.py` catches. Verified at the adapter boundary:

```text
type=[]  -> UNHANDLED TypeError: unhashable type: 'list'   => 500
type={}  -> UNHANDLED TypeError: unhashable type: 'dict'   => 500
type=""  -> HTTPException 400  ✅
```

The new inline comment asserts a non-string `type` "belongs in the `UnknownType` 422
below" — true for `0`/`123`, false for `[]`/`{}`. The `[]`/`{}` boundary is tested only
for `published`.

**3. MV-04-003 refuses stubs that MV-10 declares legal.**
MV-10-001 permits an inline object carrying only `id` + `type`; MV-10-002 permits it in
`actor`/`object`/`target`/`origin`. Verified through `parse_activity`:

```text
actor  {id, type: VultronOrganization}  -> 422
object {id, type: Invite}               -> 422
object {id, type: ParticipantStatus}    -> 422
object {id, type: VulnerabilityCase}    -> OK  (pre-existing escape hatch)
```

ADR-0090 does not mention stubs, and its Validation section's "the full suite passes
unchanged" cannot support "refusing broke nothing": the suite exercises Vultron's own
emitters, which never send an MV-10 stub except for `VulnerabilityCase`, so it
structurally cannot contain the counterexample. Suggested fix *removes* a special case —
generalise the stub-shape check from VulnerabilityCase-only to any type whose inline keys
are a subset of the stub key set. A corrupt object carries more keys than the stub set, so
this does not reintroduce the bug the PR fixes.

---

### On the merge conflict

One path: `vultron/wire/as2/parser.py`. It is **semantic, not textual**. `origin/main`
`723ff08eb` already shipped `if not body.get("published"):` — ADR-0090's option 1, which
that ADR explicitly rejected — and closed #3217 as completed. Resolution must keep
`_is_blank` and discard main's line; taking main's side fails
`test_parse_activity_reports_unparseable_published_as_malformed`, which pins
`published: 0` and `published: []` to a schema fault rather than a missing field. Do not
auto-resolve by blindly preferring either side.

### Verified clean

PR body format (closing refs at top, real test counts); ADR-0090 matches
`_adr-template.md` with four considered options; both new `kind: protocol` entries carry
`@pytest.mark.spec` markers and Spec Corpus Tests pass; no new `vultron.core.models`
import in `vultron/wire/`; the `except VultronParseError: raise` ordering is correct
(expansion is evaluated inside the `try`, before `model_validate`); UTC normalization
composes correctly with the blank check; learning files carry valid `signal:` values
(BW-07-002); the `bugfix` skill edit is in `.agents/` only and is BW-07-004-correct
routing.

---

*Triage artifact: `.claude/pr-3233-triage.json`*

sei-ahouseholder:
## Triage correction — the MV-10 stub finding was wrong

My triage comment above recommended generalising the `VulnerabilityCase` stub check to
all types. **Withdrawn.** Maintainer review overturned it, and checking the code
corroborates the objection. Revised verdict below; the rest of the triage stands.

### What I got wrong

I read MV-10-001/MV-10-002 as licensing dict-shaped stubs for any type, and therefore read
the new MV-04-003 refusal as a defect. The refusal is correct. The over-broad artifact is
MV-10.

Verified:

- **`VulnerabilityCaseStub` is the only stub Vultron implements.** It has a class, a
  factory (`_project_case_to_stub`), and a justified transient protocol role — give an
  invitee something to evaluate *before* acceptance, since the case cannot be shared until
  then (CM-16-006, `Invite(CaseStub+embargo+roles)`). There is no `InviteStub`, no
  `ParticipantStatusStub`.
- **Arbitrary-type dict stubs were never designed.** They were tolerated only by the
  `except Exception: return expanded` fallback that this PR removes. Generalising the stub
  check would freeze in a workaround that exists only because the prototype cannot
  dereference another actor's objects.
- **Bare URI references are unaffected.** A plain URI string in `object`, `actor` or
  `target`, and an explicit `Link` dict, all still parse. The AS2-idiomatic way to
  reference an object you are not inlining keeps working, so the refusal's blast radius is
  limited to dict stubs.

### Revised findings

| # | Severity was | Severity now | Change |
|---|---|---|---|
| `phase5-mv04-003-refuses-mv10-stubs-0` | ❌ FAIL | — | withdrawn, replaced |
| `phase5-mv10-overbroad-vs-mv04-003-0` | — | ⚠️ IMPROVE | narrow **MV-10 in the spec**; do not touch the parser |
| `phase5-no-dereference-mechanism-0` | — | ⚠️ IMPROVE | new: file the missing dereference capability as type:Idea (BW-07-004) |
| `phase9-stub-objects-note-stale-0` | ⚠️ IMPROVE | ⚠️ IMPROVE | rescoped to the narrowing, not to a general stub carve-out |

Revised totals: **3 FAIL · 11 IMPROVE**. The two other FAILs — the nested-timestamp clock
fabrication and `type: []` returning a 500 — are unchanged, as is the merge conflict.

`notes/stub-objects.md` needs rewriting rather than annotating: its "Stub vs Full Object"
table currently offers two generic uses ("large object already known to recipient",
"privacy-sensitive fields must be withheld") that the narrowing removes. Without that
rewrite a future agent reads the table and re-introduces arbitrary stubs.

### Also, per maintainer direction

#3217 stays **closed**. Execute will drop the stale `Closes #3217` bullet and post an
explanatory comment on #3217 instead, rather than reopening it.

---

*Updated artifact: `.claude/pr-3233-triage.json`*

sei-ahouseholder (commented):
## PR Execute: #3233

**CI**: ✅ 25/25 passing (2 iterations)
**Findings**: 12 fixed · 2 deferred (both gate-approved) · 0 skipped

---

### Fixed

| # | Description | Commit |
|---|---|---|
| `phase12-merge-conflict-0` | Resolved `parser.py` keeping `is_blank`, superseding main's `723ff08eb` | `753c657a8` |
| `phase8-nonstring-type-500-0` | `type: []` / `type: {}` now 422, not an unmapped `TypeError` → 500 | `3efd92e78` |
| `phase5-mv10-overbroad-vs-mv04-003-0` | MV-10-001/002 narrowed to `VulnerabilityCase` | `3efd92e78` |
| `phase8-inline-vocab-strips-core-typed-refs-0` | Narrowed the overbroad "wire trees are wire-to-wire" claim + test | `3efd92e78` |
| `phase10-blank-default-docstring-contradiction-0` | Docstring claimed blank == omitted; it does not. Corrected + pinned | `3efd92e78` |
| `phase8-refusal-message-lacks-path-0` | Refusals now name the field path | `3efd92e78` |
| `phase7-blank-predicate-duplicated-0` | `is_blank` extracted as the wire layer's single definition | `3efd92e78` |
| `phase9-wire-core-boundary-note-stale-0` | Wire-only-resolution rule recorded in both notes | `3efd92e78` |
| `phase9-published-refusal-narrow-in-notes-0` | `AGENTS.md` + `case-ledger-authority.md` widened to blank | `3efd92e78` |
| `phase9-stub-objects-note-stale-0` | `stub-objects.md` + `README.md` rewritten for the narrowing | `3efd92e78` |
| `phase2-issue-3217-already-closed-0` | Stale `Closes #3217` dropped; #3217 commented, not reopened | `3efd92e78` |
| `phase6-adr-0090-count-drift-0` | Hardcoded test count removed; evidence claim corrected | `3efd92e78` |

### Deferred (Gate 1, explicitly approved)

- **#3257** — a nested object with no claimed `published` gets the receiver's clock.
  Measured across all four spellings: an *omitted* nested `published` already
  fabricates on `main` via `default_factory=now_utc`, so this PR did not introduce it —
  it routes blank to the same pre-existing outcome. Remainder: the outbound path
  legitimately needs the default, so a fix means splitting inbound from outbound
  defaults and changing `_get_timestamp` to return `datetime | None`.
- **#3258** — no way to dereference a URI object reference from another actor. This is
  *why* senders inlined partial objects; narrowing MV-10 exposes it. Remainder: an
  inter-actor resolve path, a trust model interacting with embargo state, caching and
  failure semantics, and `payloadSnapshot` immutability.

Both are on Project #24 and cross-referenced from ADR-0090 and the notes, so neither
can be silently stranded.

---

### Two triage findings I got wrong, corrected before fixing

**The headline FAIL was overstated.** I reported the blank-timestamp change as
introducing a CLP-15-006 clock substitution. Measuring all four spellings showed the
fabrication pre-exists on `main` for *omitted* nested timestamps; this PR makes blank
match omitted, which is the intended semantics. Downgraded to IMPROVE and split — the
PR-owned residue was only a wrong docstring, now fixed and pinned.

**The MV-10 stub recommendation was backwards.** I proposed generalising the
`VulnerabilityCase` stub check to all types. Maintainer review overturned it and the
code agrees: `VulnerabilityCaseStub` is the only stub implemented, and generalising
would have entrenched a workaround for the missing dereference capability. The
refusal is correct; the *spec* was over-broad. Fixed there instead.

Both corrections were posted before any code changed.

### Verification

- Unit suite: reached 100%, zero FAILED/ERROR (summary contains only SKIPPED/XFAIL)
- Integration suite: 1258 passed, 2 xfailed
- Black, flake8, mypy, pyright clean **tree-wide**; spec-lint and markdownlint clean
- xfail ratchet: every xfail reason points to a live open issue

One process note worth recording: CI iteration 1 failed mypy and pyright on the new
tests because I had run those linters against the changed files only, while CI invokes
them bare over the whole tree. Fixed in `eecb0e6cc`.

---

*Execute artifact: `.claude/pr-3233-execute.json`*
*Next step: `/pr-verify`.*

sei-ahouseholder (commented):
## PR Verify: #3233

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, not a draft, no conflict markers in the tree
**CI**: ✅ 23 passing, 2 skipping (the `Notify CI failure` jobs, which correctly skip when nothing failed), 0 failing
**Artifact integrity**: ✅ 14 execute results ↔ 14 triage findings, all IDs matched

---

### FAIL findings — verified at HEAD, not just in the commit diff

| # | Verdict | Evidence at HEAD |
|---|---|---|
| `phase12-merge-conflict-0` | ✅ CONFIRMED | `is_blank(body.get("published"))` present at `parser.py:233`; main's `if not body.get("published")` absent |
| `phase8-nonstring-type-500-0` | ✅ CONFIRMED | `if not isinstance(type_, str)` present at `parser.py:191`; the false `cast(str, type_)` removed |

Both were also re-checked **behaviorally** at HEAD rather than by reading source:

```text
type=[]  via adapter  -> HTTP 422   (was an unmapped TypeError -> 500)
type={}  via adapter  -> HTTP 422
type=0   via adapter  -> HTTP 422
type=""  via adapter  -> HTTP 400
published=0   -> VultronParseValidationError   (main's falsy guard would have said MissingPublished)
published=[]  -> VultronParseValidationError
published=""  -> VultronParseMissingPublishedError
published="  " -> VultronParseMissingPublishedError
```

The `published=0` / `published=[]` rows are the load-bearing ones: they prove the merge
resolution actually kept ADR-0090's semantics rather than silently inheriting the falsy
guard from `main`.

### IMPROVE findings — all CONFIRMED at HEAD

10 of 10 present: MV-10 narrowing, ADR-0090 stub consequence, removed test count,
narrowed ADR driver and parser docstring, field path in refusals, shared `is_blank`
(with the duplicate `_is_blank` gone), both notes recording the wire-only-resolution
rule, `AGENTS.md` and `case-ledger-authority.md` widened, `stub-objects.md` and
`notes/README.md` rewritten, corrected test docstring, and all five new test groups.

### Deferred — both gate-approved, both real and open

- **#3257** — OPEN, type:Bug, on Project #24
- **#3258** — OPEN, type:Idea, on Project #24

### Issue linkage

- `Closes #3232` is the only closing reference — correct, and #3232 is still OPEN so the
  merge will close it.
- **#3217 remains CLOSED (COMPLETED)** and was not reopened. The stale closing bullet was
  removed and an explanatory comment posted instead, per maintainer direction.

---

### Notes on the pipeline's own accuracy

Two triage findings were corrected before any code was written, and both corrections were
posted publicly rather than quietly dropped:

1. The headline FAIL — that the blank-timestamp change introduced a CLP-15-006 clock
   substitution — was **overstated**. Measurement showed the fabrication pre-exists on
   `main` for omitted nested timestamps. Downgraded to IMPROVE and split; only the
   docstring residue belonged to this PR.
2. The MV-10 stub recommendation was **backwards**. Generalising the stub check would
   have entrenched a workaround for the missing dereference capability. The code was
   right; the spec was over-broad.

Recording these because a verdict is only worth as much as the honesty of the findings
behind it.

---

*Verified against HEAD `eecb0e6cc`. Session artifacts cleaned up.*

Issue #3282, opened 2026-09-16
CLP-07-003's enforcement was dormant before ADR-0088 and cannot express the requirement now


sei-ahouseholder:
## Context

Surfaced by code review of the ADR-0088 completion PR (#3192/#3193/#3195/#3273).
No behaviour changed — `_validate_canonical_entry` is exactly as it was on `main`.
This records a problem that ADR-0088 made *reachable* rather than one it caused.

## The requirement

CLP-07-003: `payloadSnapshot.actor` MUST preserve the original asserting actor's
identity; the CASE_MANAGER's identity appears only on the transport envelope and
in recording metadata, **never as a substitute for the asserter**.

## The implementation

`vultron/core/behaviors/sync/nodes/canonical_entry.py` enforces it as:

```python
if (case_actor_id
        and snapshot_actor == case_actor_id
        and signature not in _CASE_AUTHORED_SIGNATURES):
    raise VultronCanonicalEntryError(...)
```

## Why this was dormant, and what changed

`case_actor_id` comes from `_find_case_actor_id`. Before ADR-0088 that was gated
on the `case-actor` URL shape, so for any case whose manager was an ordinary
participant it returned `None` — and the whole check short-circuited. The
predicate was therefore rarely exercised. ADR-0088 makes the resolver answer
unconditionally, so it now fires wherever it applies.

## The problem it now has

The check cannot distinguish the two situations the requirement cares about,
because it is an identity comparison and the distinction is not one of identity:

- **False positive.** A coordinator holding `CVDRole.CASE_MANAGER` that adds a
  note to a case it manages is the genuine asserter. `Add(Note)` is not in
  `_CASE_AUTHORED_SIGNATURES`, so a correctly-formed entry is refused. Under
  ADR-0088 the authority *is* a participant (CM-02-011) and legitimately asserts;
  the check treats the two categories as disjoint.
- **No usable discriminator.** A narrowing that keys on `attributedTo` was tried
  and reverted: `build_activity_payload_snapshot` dumps with `exclude_none=True`,
  so `attributedTo` is absent from the normal serialization — meaning it is absent
  in both the legitimate case and in a substitution. It also condemned the shape
  CM-24-001/002 explicitly mandate for delegated emits, which today only escapes
  because both delegated signatures (`Invite`/`Offer(VulnerabilityCase)`) happen
  to sit in `_CASE_AUTHORED_SIGNATURES`. Add a delegated emit for any other
  canonical signature and the two rules collide.

[...]

Issue #3274, opened 2026-09-16
First-contact replica seeding has no trust anchor against a fabricated case roster


sei-ahouseholder:
## Context

Surfaced by #3273 while implementing ADR-0088. Not a regression — a gap the
previous code hid behind a signal that was itself wrong.

`AnnounceVulnerabilityCaseReceivedUseCase` decides whether an inbound
`Announce(VulnerabilityCase)` may seed a local replica. At **first contact** the
case is not in the local store, so there is no locally-derived roster to check
the sender against. The check therefore falls back to the roster carried in the
announced payload (CP-09-004 embeds participants inline).

## The gap

That roster is supplied by the sender. It catches the realistic imposter — an
actor replaying or forwarding a *legitimate* case whose roster names somebody
else as the authority — but a wholly **fabricated** case naming the sender as its
own `CVDRole.CASE_MANAGER` passes. Nothing currently available at first contact
can refute it.

The previous `Service`-hosting check was not actually better: it only answered
when the receiver had already written a CaseActor `Service` for that case, and it
keyed off hosting location, which ADR-0088 established is not evidence of
authority (ARCH-24-004, CM-02-013). It gave the appearance of a guard in a
narrower window.

## Candidate anchors (not yet evaluated)

- **Pending `VultronReportCaseLink.trusted_case_creator_id`** — recorded locally
  by the receiver from its *own* outbound `Create(as_CaseProposal)`, so it is
  locally derived and trustworthy in exactly this window. Broad, though: it would
  admit an announce from any actor we have an outstanding proposal to. This is the
  path retired as `_find_case_actor_id`'s path 2 in #3192 — retired as a redundant
  *address* path, but it may have standalone value as a *trust* anchor.
- **Peer registration** (ADR-0081) — an actor we have deliberately registered.
- **Signatures** — the general answer, and out of proportion to this issue alone.
  See `notes/encryption.md` and #1156.

## What to decide

1. Is admitting an unsolicited first-contact case seeding acceptable at all, or
   must seeding require a prior local record (proposal, invite, or peer
   registration)?
2. If a prior record is required, which of the anchors above is the one, and what
   is the disposition when none is present — reject, or park pending review?

## Notes

Related: #3273 (the change that exposed this), #2626 (cross-authority
replication), #1156 (message signing). Relevant spec: PCR-07-003, PCR-03-001.
Guard and its documented limits: `_announced_case_manager_id` in
`vultron/core/use_cases/received/actor/announce.py`.

[...]

Issue #3261, opened 2026-09-15
Announce tree resolves CaseActor via legacy Service scan, not the ADR-0088 CASE_MANAGER role


sei-ahouseholder:
## Context

Found during triage/execute of PR #3256 (enforce CLP-01-003 sender verification
in the participant announce path). PR #3256 migrated the new
`VerifySenderIsCaseActorNode` to resolve the CaseActor by the
`CVDRole.CASE_MANAGER` role via the neutral `resolve_case_manager_id`
(ADR-0088, ARCH-24-001), with a legacy `Service`-context fallback. The rest of
the `Announce(CaseLedgerEntry)` receive tree was left on the legacy model.

## Problem

`vultron/core/behaviors/sync/announce_tree.py` still routes CaseActor-vs-participant
handling through nodes that resolve the CaseActor with a **legacy `Service`
whose `context == case_id`** scan:

- `CheckIsOwnCaseActorNode` and `CheckIsNotOwnCaseActorNode`
  (`vultron/core/behaviors/sync/nodes/conditions.py`) both call the module-local
  `_find_case_actor(dl, case_id, actor_id)`.
- A near-duplicate `_find_case_actor` also exists in
  `vultron/core/behaviors/sync/nodes/replay.py` (with an even looser fallback).

Under the modern container-identity CaseActor model (ADR-0041 / #1872), the
CaseActor `Service` is created with **no `context`**
(`vultron/core/behaviors/case/nodes/case_setup.py:326`:
`VultronCaseActor(id_=case_actor_id, name="CaseActor")`), so a
`context == case_id` scan resolves nothing. Authority is carried by the
`CVDRole.CASE_MANAGER` role (ADR-0088), which these nodes do not consult.

Consequence: for modern cases the case-actor-arm routing in this tree can
mis-route (the CaseActor's own receipt falls through to the participant arm),
and the two `_find_case_actor` helpers duplicate — with subtly different
fallback semantics — the resolution logic that `resolve_case_manager_id`
(neutral layer) and `use_cases/_helpers._find_case_actor_id` already own.
This is the same class of gap PR #3256 fixed in one node; it should be closed
tree-wide.

## Task

1. Migrate `CheckIsOwnCaseActorNode` / `CheckIsNotOwnCaseActorNode` (and the
   `replay.py` counterparts) to resolve the CaseActor via the neutral
   `resolve_case_manager_id` (ADR-0088), keeping a legacy `Service`-context
[...]

Issue #3190, opened 2026-09-14
ADR-0088: collapse authority predicate nodes into one role-based condition


sei-ahouseholder:
## Context

Implements **ADR-0088**. Depends on the single role-authority resolver (blocked-by task).

Three condition nodes each resolve authority independently:
- `CheckIsCaseManagerNode` (`vultron/core/behaviors/case/nodes/conditions.py`) — and it performs a `behaviors → use_cases` import.
- `CheckIsOwnCaseActorNode` / `CheckIsNotOwnCaseActorNode` (`vultron/core/behaviors/sync/nodes/conditions.py`).

## Scope

- Collapse these into one authority condition node (plus a negation decorator) that calls the single role-authority resolver from the neutral layer.
- Update all call sites to the unified node.

## Acceptance criteria

- [ ] Exactly one authority condition node answers "is the executing actor the case authority", and it calls the single resolver (**ARCH-24-003**).
- [ ] `CheckIsOwnCaseActorNode` / `CheckIsNotOwnCaseActorNode` are removed or aliased; `CheckIsCaseManagerNode`'s `behaviors → use_cases` import is gone.
- [ ] Test asserts one node is the only thing resolving authority.
- [ ] Full suite green.

Spec: ARCH-24-003. ADR: docs/adr/0088-consolidate-case-authority-determination.md

sei-ahouseholder:
Claimed on branch `task/3190-collapse-authority-predicate-nodes`.

sei-ahouseholder:
**History: implementation — collapse authority predicate nodes into CheckIsCaseManagerNode**


## Issue #3190 — ADR-0088: collapse authority predicate nodes into one role-based condition

Removed CheckIsOwnCaseActorNode and CheckIsNotOwnCaseActorNode (Service-scan nodes); extended CheckIsCaseManagerNode with activity input port and case_actor_id output port; updated announce_tree.py to use unified node + Inverter; added ARCH-24-003 ratchet tests and bootstrap-window regression test. PR: https://github.com/CERTCC/Vultron/pull/3253. Filed #3251 (hollow spoof test) and #3252 (double DB read) as DEFER findings.

PR #3290, opened 2026-09-16
Complete ADR-0088: authority is the CASE_MANAGER role, not a name, URL, or hosting location

### LINKED ISSUE #3192: ADR-0088: retire URL-shape / Service-hosting signals from authority
## Context

Implements **ADR-0088**. Depends on the unified role-based authority predicate.

The wrong signals for authority/recognition/routing:
- `is_case_actor_identity` (`vultron/core/behaviors/case/case_actor_identity.py`) — tests whether an id ends in `/actors/case-actor` (URL shape).
- `_find_case_actor` (`vultron/core/behaviors/sync/nodes/conditions.py`) — scans for a `Service` object hosting the case (Service-hosting).

## Scope

- Remove `is_case_actor_identity` and the `Service`-hosting resolution from authority, recognition, and routing.
- `_find_case_actor` loses its only callers once the predicate is role-based (a sibling task) — delete it.
- Drop the URL-shape gate from `_case_actor_by_role`.

## Acceptance criteria

- [ ] No protocol-logic module branches on `is_case_actor_identity` or the `case-actor` URL substring, or on `Service`-hosting, for authority/recognition/routing (**ARCH-24-004**, **CM-02-013**).
- [ ] `_find_case_actor` deleted; `is_case_actor_identity` removed from authority paths.
- [ ] Ratchet test asserts no protocol-logic module branches on `is_case_actor_identity` / the `case-actor` URL substring.
- [ ] Full suite green.

Spec: ARCH-24-004, CM-02-013. ADR: docs/adr/0088-consolidate-case-authority-determination.md

### LINKED ISSUE #3193: ADR-0088: role-gate the announce two-arm site + bootstrap-window regression
## Context

Implements **ADR-0088**. Depends on the unified role-based authority predicate. Related to **#2626** (cross-authority ledger replication divergence).

`AnnounceLogEntryReceivedBT` (`vultron/core/behaviors/sync/announce_tree.py`) is the one genuine two-arm site: the authority arm confirms its own broadcast, the participant arm validates the hash chain and applies the entry. Its arms currently gate on the hosting/shape signal, which has a **bootstrap window** where the real CaseActor fails its own hosting test and takes the *participant* arm on its own ledger.

## Scope

- Switch the announce router's two arms to the role predicate.
- Add a regression test for the bootstrap window.
- No general router abstraction — the scan found exactly one two-arm site.

## Acceptance criteria

- [ ] The announce two-arm split gates on the role predicate (**CM-02-011**).
- [ ] Regression test: the actor enacting `CVDRole.CASE_MANAGER` takes the authority arm on its own ledger entry even before any `Service` object carries `context` (**CM-02-012**).
- [ ] Confirm whether this resolves or partially addresses **#2626**; link accordingly.
- [ ] Full suite green.

Spec: CM-02-011, CM-02-012. ADR: docs/adr/0088-consolidate-case-authority-determination.md

### LINKED ISSUE #3195: ADR-0088: fold address resolution into the role-holder; reclassify the store guard
## Context

Implements **ADR-0088**. Depends on retiring the URL-shape/Service-hosting signals.

Two clean-up items the ADR calls out:
- `_find_case_actor_id` (`vultron/core/use_cases/_helpers.py`) combines ReportCaseLink bootstrap paths, the role, a URL-shape gate, and a `Service` scan — but it answers a *different* question (what address do I route to?), not "am I the authority."
- `DeclineForeignLedgerCommitNode` (`vultron/core/behaviors/sync/nodes/ledger_authority.py`) / `store_for_actor` (`vultron/core/behaviors/store_scope.py`) are a store-consistency / anti-fork guard, not an authority check.

## Scope

- Keep the bootstrap paths `_find_case_actor_id` genuinely needs for *delivery*, but drop the URL-shape gate as an authority/recognition test; document it as an address lookup (authority's address = the role-holder's address).
- Relabel the store guard as store-consistency / anti-fork — explicitly not authority (docstrings + any spec/notes cross-refs).
- Update `notes/case-ledger-authority.md` and `notes/case-communication-model.md` to reflect the role-is-authority model.

## Acceptance criteria

- [ ] `_find_case_actor_id` documented as delivery address resolution; no URL-shape authority test remains.
- [ ] Store guard reclassified as store-consistency / anti-fork, not conflated with authority (**ARCH-24-005**).
- [ ] Relevant notes updated.
- [ ] Full suite green.

Spec: ARCH-24-005. ADR: docs/adr/0088-consolidate-case-authority-determination.md

### LINKED ISSUE #3273: Replica-seeding recognition lost its trust anchor when the hosting signal was retired
## Context

Found while implementing ADR-0088 (#3192/#3193/#3195), and revised after code
review found the first attempt at it wrong. See the history note at the bottom.

`AnnounceVulnerabilityCaseReceivedUseCase` decides whether an inbound
`Announce(VulnerabilityCase)` may seed or update a local replica (PCR-07-003,
PCR-03-001). The check is lenient by design — an unresolvable authority means
accept — so that first-time replica seeding works at all.

Its evidence was a local `Service` object whose `context` was the case id, written
by the receiver ahead of the proposal (ADR-0041). ADR-0088 retires hosting
location as a signal of authority (ARCH-24-004), which left this guard with no
evidence at first seeding: the case is not in the store yet, so there is no local
roster to read either.

## Two distinct holes

1. **No anchor at first seeding.** With the hosting signal gone the guard accepted
   any sender for a case it had never seen.
2. **Sender-controlled input on the seeded path, and a lenient default that let it
   through.** An already-seeded case can resolve no authority — its roster may name
   no CASE_MANAGER, or that participant may carry no `attributed_to`. Because
   `SeedAnnouncedCaseNode` *saves* on the existing-case path, accepting there
   overwrites the stored record. Reproduced end-to-end: a replica named `"Seeded"`
   plus an `Announce` from `.../actors/imposter` whose payload names itself
   CASE_MANAGER left the stored case reading `"PWNED"`. The all-`None` variant of
   this predates ADR-0088; what the ADR work briefly added was a *sender-supplied*
   input to a guard whose inputs had all been locally derived.

## Resolution

The trust decision now keys on **whether the case is present**, not on whether a
lookup happened to answer:

- **Not seeded** — prefer the locally recorded anchor: a completed
  `VultronReportCaseLink.trusted_case_actor_id` is the address this receiver
  itself reached the authority at during bootstrap (CBT-05-004), which a sender
  cannot forge. Only with no local record at all does the announced payload's
  roster get a say (CP-09-004 embeds the CASE_MANAGER inline). Both absent stays
  permissive, because accepting is the point of first seeding.
- **Already seeded** — fail closed. An update whose authority cannot be
  established locally is refused rather than applied.

The residual limit is that a wholly fabricated roster at genuine first contact
still passes, and nothing available at first contact can refute it. Tracked
separately as #3274 rather than left implied; `_announced_case_manager_id`
documents it at the call site.

## Acceptance criteria

- [x] AC-1: replica-seeding recognition gates on the role, not on hosting location (**PCR-07-003**, **ARCH-24-004**).
- [x] AC-2: a locally recorded `ReportCaseLink` anchor outranks the announced roster (**CBT-05-004**).
- [x] AC-3: the announced roster is never consulted for a case that is already seeded.
- [x] AC-4: an already-seeded case with no locally resolvable authority rejects the update rather than overwriting.
- [x] AC-5: full suite green.

## History

This issue first also covered a CLP-07-003 change in
`_validate_canonical_entry`, on the reading that its
`snapshot_actor == case_actor_id` test wrongly assumed the authority is never a
participant. Code review showed that narrowing was wrong in both directions: it
condemned the very shape CM-24-001/002 mandate, and — because
`build_activity_payload_snapshot` dumps with `exclude_none=True`, so
`attributedTo` is usually absent — it left the check unable to catch any real
substitution. It was reverted. The two tests that motivated it turned out to be
artificial (the CASE_MANAGER was set as the asserter of an assertion it was
receiving) and were made realistic instead. The genuine residual question about
CLP-07-003 under ADR-0088 is filed separately as #3282.

Spec: PCR-07-003, PCR-03-001, CBT-05-004, ARCH-24-004. ADR: ADR-0088.

### LINKED ISSUE #3306: Apply ADR-0088's prose half: name the authority CASE_MANAGER across notes/
## Context

ADR-0088 §7 and the glossary require protocol-normative prose to name the
authority **the CASE_MANAGER** (the role holder), reserving **"CaseActor"** for
the concrete prototype/demo actor that enacts the role and **"case actor
service"** for the provisioning endpoint. `notes/spec-authoring-rules.md` already
carries the rule for spec YAML, and AGENTS.md indexes it.

The ADR's code half is ratcheted (ARCH-24-004, via
`test/architecture/test_role_authority_resolver.py`). The prose half was not
applied: roughly 24 files under `notes/` still used "CaseActor" as a synonym for
the authority, including in normative-sounding statements such as "The Case Actor
is the single writer of authoritative shared history".

The sharpest case was `notes/case-communication-model.md` — the primary reference
for outbound addressing — which contradicted itself: its frontmatter description
and ~40 body references named the authority "the Case Actor" while a section
added in the same work stated that the name carries no protocol meaning.

## Scope

Per-site judgement, not a blanket rename. Convert only where the text means the
*authority*; keep the name where it means the concrete actor, its container, or
the provisioning endpoint.

**Converted**: `case-communication-model.md`, `participant-embargo-consent.md`,
`case-state-model.md`, `protocol-event-cascades.md`, `sync-behavior-trees.md`,
`ownership-transfer.md`, `bt-fuzzer-rm-threat.md`, `case-ledger-authority.md`
(diagram), `case-bootstrap-trust.md`, `activitystreams-state-update.md`,
`embargo-lifecycle.md`, `domain-validation.md`, `outbox.md`,
`demo-future-ideas.md`, `triggerable-behaviors.md`,
`triggerable-behaviors-resolved.md`, `message-type-reference.md`,
`demo-scenario-authoring.md`, `triggers-test-coverage.md`, `stub-objects.md`,
`received-status-authorization.md`, `case-creation-sequence.md`,
`bt-fuzzer-rm-reporting.md`, `bt-fuzzer-nodes-report-management.md`,
`architecture-adapters.md`, `core-wire-rendering-port.md`,
`participant-case-replica.md`.

**Deliberately not converted** — the name is correct in these:
`federation_ideas.md`, `case-proposal.md`, `fv-demo.md`,
`sync-ledger-replication.md`, `configuration.md`, `bt-integration.md`,
`spec-authoring-rules.md`, `demo-interactive-ui.md`,
`demo-ci-scenario-coverage.md`, `flaky-tests.md`, `demo-ci-diagnostics.md`,
`encryption.md`, `datalayer-design.md`, `architecture-hexagonal.md`,
`bt-fuzzer-nodes-embargo.md`, `README.md`. These refer to the concrete actor
(inbox, key material, container, migration), to deployment topology, or to class
and node names.

Two substantive corrections came out of the pass:

- `notes/federation_ideas.md` modelled the CaseActor as "1:1 with Case", which
  contradicts CP-08-002/003 (#1872): the role is one per container, not one per
  case, and the per-case identity was abandoned because it was unhostable by
  construction and its delivery 404'd. Corrected, and given a terminology
  preamble rather than a rename.
- `notes/sync-behavior-trees.md` documented `CheckIsOwnCaseActor` /
  `CheckIsNotOwnCaseActor` — both deleted by ADR-0088 — and described the announce
  two-arm split as an identity check rather than a role check.

## Acceptance criteria

- [x] AC-1: no `notes/` file uses "CaseActor"/"Case Actor" as a synonym for the authority in normative prose (CM-02-013).
- [x] AC-2: files where the name denotes the concrete actor or the provisioning endpoint keep it, and `federation_ideas.md` states which referent it means.
- [x] AC-3: the `federation_ideas.md` 1:1-per-case model and the `sync-behavior-trees.md` retired-node references are corrected.
- [x] AC-4: `markdownlint` clean.

Spec: CM-02-011, CM-02-013, ARCH-24-004, PD-03-002. ADR: ADR-0088.

sei-ahouseholder:
- Closes #3192
- Closes #3193
- Closes #3195
- Closes #3273
- Closes #3306

## Summary

Completes ADR-0088: authority over a case is the `CVDRole.CASE_MANAGER` role and
nothing else. Retires the URL-shape and `Service`-hosting signals from authority,
recognition, and routing; folds address resolution into the role-holder;
reclassifies the store guard as anti-fork rather than authority; and ratchets the
name out so it cannot come back.

## Motivation

ADR-0088 found authority being decided four ways, three of them by the wrong
signal. #3189 and #3190 landed the single neutral resolver and the single
condition node. This PR finishes the negative half — deleting the wrong signals —
which is where the interesting consequences live.

Two of those consequences were premise inversions in code the ADR's scan did not
reach, filed as #3273 and gated with the reviewer before being acted on. Both
sites determined *recognition* on the assumption that "the CaseActor" is an entity
distinct from whoever holds the role. Removing the hosting signals broke each in
an opposite direction: one check became too strict and rejected the authority's
own legitimate ledger entries; the other became too lenient and lost its only
evidence at first replica seeding.

### Retiring the wrong signals (#3192)

- **`vultron/core/behaviors/case/case_actor_identity.py`**: delete
  `is_case_actor_identity`. `case_actor_identity()` stays — ADR-0088 §7 keeps
  ADR-0041's stable provisioning URL while stripping its protocol meaning.
- **`vultron/core/behaviors/sync/nodes/replay.py`**: delete `_find_case_actor`.
  `FindCaseActorNode` resolves through the neutral `resolve_case_manager_id`,
  which also removes the `behaviors → use_cases` reach, and through
  `_require_case` (Regime 1, ADR-0087) instead of a bare `read_case`. The deleted
  helper fell back to *the first arbitrary `Service` in the store* when nothing
  matched, so a failed lookup returned SUCCESS with a plausible-looking address
  belonging to some other case.
- **`test/architecture/test_role_authority_resolver.py`**: the ARCH-24-004
  ratchet, covering **both** signals the requirement names. It flags
  *comparisons, shape-test calls and `match` patterns* rather than occurrences —
  holding a literal is fine (example URIs, the provisioned segment), testing one
  to decide authority is the violation — and separately flags a `Service` store
  scan combined with a hosted-`context` test, which is the resolver this PR
  deletes. Each scanner has a self-test that fires it on a known-bad snippet,
  since a scanner bug would otherwise read as a clean repo.

  The first revision of this ratchet covered only the URL half and was evadable;
  triage demonstrated three bypasses against the real test and all three are now
  caught, each with its own regression test: restoring the deleted
  `Service`-hosting resolver verbatim (nothing scanned for it at all), annotating
  the suffix constant (`_SUFFIX: str = ...`, which defeated the very test written
  to close binding indirection, because only `ast.Assign` was inspected), and
  `actor_id == case_actor_identity()` — the likeliest re-introduction, since it
  names no literal and so was never even opened by the prefilter, though the
  glossary forbids exactly that comparison. Also fixed: the provisioning
  exemption was whole-file, so a renamed verbatim copy of the deleted predicate
  could sit beside `case_actor_identity` and pass; it is now scoped to that
  function. `vultron/semantic_registry/` joins the scanned corpus — it is
  recognition and routing, the two words ARCH-24-004 uses.

### Folding address resolution into the role-holder (#3195)

- **`vultron/core/use_cases/_helpers.py`**: `_find_case_actor_id` drops the
  URL-shape gate, the `Service`-`context` scan, and the pending-link path the
  shape gate propped up — once the role answers unconditionally, the replica's
  roster covers that window on its own (CP-09-004). Two paths remain: the address
  recorded on a completed `ReportCaseLink`, then the role-holder. Redocumented as
  address resolution, a different question from authority (ARCH-24-005).
  `_case_actor_by_role` collapses into `resolve_case_manager_id`;
  `_scan_report_case_links` narrows to `_established_link_actor_id`.
- **`vultron/core/behaviors/sync/nodes/ledger_authority.py`**:
  `DeclineForeignLedgerCommitNode` reclassified as store-consistency / anti-fork,
  explicitly not an authority check. Reading it as one makes the role look like a
  property of a store, which it is not.
- **`vultron/core/behaviors/bridge.py`**: corrected the docstring claiming
  `_find_case_actor_id` resolves "by identity shape".

### The announce two-arm site (#3193)

Already role-gated by #3190, and the bootstrap-window regression test already
existed — it was marked `ARCH-24-003` only. **`test/core/behaviors/sync/test_announce_tree.py`**
adds the `CM-02-011` and `CM-02-012` markers so both requirements trace to their
test. AC-3 (the #2626 relationship) is adjudicated
[on #2626](https://github.com/CERTCC/Vultron/issues/2626#issuecomment-5703669490):
**ADR-0088 does not resolve it.** All nine invariant harnesses — including the one
owning `test_invariant_2_cross_actor_hash_agreement` — already pass on `main` at
`a32494ab7`, the commit this branch is rooted on, so the symptoms predate and are
independent of this work. This PR leaves those scenarios green, which is all it
claims, and #2626 is left open with a recommendation to attribute the real fix
before closing it.

Two caveats on that evidence. The nine-harness result is from the `main` run
([35130743699](https://github.com/CERTCC/Vultron/actions/runs/35130743699)); this
PR's own gate gets the reduced four-scenario set per DEMOCI-06, which does include
both `fcv-reject` and `fvcv-handoff` and passes. And an earlier comment of mine on
both issues said the evidence was unobtainable because `docker` is unavailable
locally — that overlooked `demo-integration.yml` running the matrix on every push
to `main`, and is corrected in place on both.

### Replica-seeding recognition (#3273)

- **`vultron/core/use_cases/received/actor/announce.py`**: the imposter check is
  lenient by design so first seeding works, and its only evidence was the local
  hosting `Service`. `_authority_verdict` now returns an `_AuthorityVerdict`
  whose `admits()` branches on **whether this receiver holds a locally-derived
  roster**, not on whether a lookup happened to answer:
  - *Has a local roster* → fail closed. Such a case can still resolve `None`
    (roster names no CASE_MANAGER, or that participant has no `attributed_to`),
    and because `SeedAnnouncedCaseNode` *saves* on the existing-case path,
    treating that as "no local opinion" let an imposter overwrite the stored
    record. Code review demonstrated this end-to-end; two regression tests cover
    it.
  - The discriminator is the roster and **not** the presence of the case row,
    which is a correction to this PR's first revision. The FastAPI ingress
    adapter pre-stores an inbound activity's nested objects before dispatch, so
    on the HTTP path the row is always present — it is the announce itself,
    echoed back, with `case_participants` still bare ID strings that resolve to
    nothing. Reading that as local evidence made the legitimate CASE_MANAGER's
    *first* announce fail closed against itself, blocking `SeedAnnouncedCaseNode`
    entirely and self-perpetuating on every retry. It was firing in-tree, in the
    already-passing `test/demo/test_pcr_bootstrap.py`, which masked it because
    its prerequisite assertion was satisfied by the ingress pre-store. Verified
    absent on `main`. `has_local_participant_roster` (neutral layer, beside
    `resolve_case_manager_id`) is the discriminator, and
    `test_ingress_prestored_case_row_is_not_local_evidence` pins it — it fails
    under the row-presence reading.
  - *Unseeded* → prefer the locally recorded `ReportCaseLink` anchor
    (CBT-05-004), which the sender cannot forge; only then the announced roster
    (CP-09-004 embeds the CASE_MANAGER inline); both absent stays permissive.

  `_announced_case_manager_id` documents what this cannot refute — a wholly
  fabricated roster at genuine first contact, which nothing available at first
  contact can. Filed as #3274 rather than left implied.

### What was tried and reverted

An earlier revision of this PR also narrowed CLP-07-003 in
`canonical_entry.py`, on the reading that its `snapshot_actor == case_actor_id`
test wrongly assumed the authority is never a participant. Code review showed the
narrowing was wrong in both directions: it condemned the exact shape
CM-24-001/002 mandate for delegated emits (escaping only because both delegated
signatures happen to sit in `_CASE_AUTHORED_SIGNATURES`), and because
`build_activity_payload_snapshot` dumps with `exclude_none=True` — so
`attributedTo` is normally absent — it left the check unable to catch any real
substitution either.

**It is fully reverted**; `canonical_entry.py` and CLP-07-003's spec text are
byte-identical to `origin/main`. The two tests that motivated it turned out to be
artificial — they named the receiving CASE_MANAGER as the asserter of an assertion
it was receiving — and are now realistic (the vendor asserts, the CASE_MANAGER
receives and commits), passing against the unmodified strict check.

The genuine residual problem is real but needs a decision, not a patch: the strict
check has a true false positive (a coordinator adding a note to a case it manages),
and the information that would distinguish legitimate self-assertion from
substitution is not present in the snapshot. Filed as **#3282**, including the
observation that this predicate was largely *dormant* before ADR-0088 — the shape
gate made `case_actor_id` resolve `None` for most cases, so it rarely fired at
all.

### Tests restated rather than deleted

Four tests were passing on the retired hosting signal — the reject-sync replay,
both delegated-emit trigger tests, and the sync-replication integration case.
They are restated in terms of the role, preserving their intent: the
delegated-emit shape (`actor` = authority, `attributedTo` = requester) is
unchanged by ADR-0088; only how the authority is resolved changed.

- **`test/core/use_cases/test_helpers.py`**: `TestFindCaseActorId` rewritten for
  the two-path contract, including the behaviour change called out below.
- **`test/core/behaviors/sync/nodes/test_replay.py`**: `FindCaseActorNode` had
  **no** test coverage at all; adds four, including one pinning that an unrelated
  `Service` is no longer borrowed as the address.
- **`test/core/use_cases/received/actor/test_announce.py`**: adds the accept
  mirror and the local-roster-outranks-announced case.
- **`test/core/behaviors/sync/nodes/test_canonical_entry.py`**: adds the
  authority-asserting-as-itself, self-attributed, and inline-`attributedTo`
  cases.
- **`notes/`**: authority vs. address vs. store in `case-ledger-authority.md`,
  resolver selection in `case-communication-model.md`, and a corrected
  hosting-scan claim plus the current recognition shape in
  `participant-case-replica.md` (whose embedded sample still showed the retired,
  silently-permissive guard, and which pointed at a module path that does not
  exist).
- **`notes/` — ADR-0088's prose half (Closes #3306)**: the ADR requires
  protocol-normative prose to name the authority *the CASE_MANAGER*, reserving
  "CaseActor" for the prototype actor and "case actor service" for the
  provisioning endpoint. This PR ratchets that in code; the prose was still using
  "CaseActor" as a synonym for the authority in ~24 notes. Converted where the
  text means the authority — `case-communication-model.md` (~40 sites, including
  its own frontmatter description, which contradicted the section this PR added),
  `participant-embargo-consent.md`, `case-state-model.md`,
  `protocol-event-cascades.md`, `sync-behavior-trees.md`, `ownership-transfer.md`,
  `bt-fuzzer-rm-threat.md`, and a dozen smaller sites.

  Deliberately **not** converted where the name is correct: `federation_ideas.md`,
  `case-proposal.md`, `fv-demo.md`, `sync-ledger-replication.md`,
  `configuration.md`, `bt-integration.md`, `spec-authoring-rules.md` and others
  use it for the concrete actor, its container, or the provisioning endpoint —
  which is the referent the glossary reserves it for. A blanket rename would have
  been wrong in those files, so `federation_ideas.md` gets a terminology preamble
  instead, plus a correction: its "CaseActor (1:1 with Case)" model contradicts
  CP-08-002/003 (one per container, not one per case — the per-case identity was
  unhostable and 404'd).

  `sync-behavior-trees.md` needed more than terminology: it documented
  `CheckIsOwnCaseActor` / `CheckIsNotOwnCaseActor`, both deleted by ADR-0088, and
  described the announce split as an identity check.

## Behaviour change worth naming

`_find_case_actor_id` now resolves an ordinary participant that enacts
CASE_MANAGER, where the shape gate previously answered `None`. That `None` was
the ADR-0021 reading ADR-0088 retires: there is no separate CaseActor entity that
can be absent while the role is held, so withholding the address made routing
depend on a naming convention. Two tests asserted the old behaviour and now
assert the new, with the rationale in their docstrings.

## Verification

- 8734 unit tests pass, 1258 integration tests pass (21 test functions added or
  restated; `FindCaseActorNode` goes from zero coverage to four tests)
- Black, flake8, mypy, pyright clean; `spec-lint` clean; `markdownlint` clean
- **`specs/architecture.yaml`, `specs/case-management.yaml`**: ADR-0088 states
  that each requirement it generated names the ratchet or regression test that
  enforces it, and that those tests land with the implementation — this is that
  implementation, so the `verification:` fields are filled in. ARCH-24-005 had
  none at all; ARCH-24-004's named only the URL half of its own statement; and
  CM-02-011/012 described their tests without naming them, though this PR adds the
  `@pytest.mark.spec` markers. No requirement *statement* changed.

**#3192**
- [x] AC-1: no protocol-logic module branches on `is_case_actor_identity` or the `case-actor` URL substring, or on `Service`-hosting, for authority/recognition/routing (ARCH-24-004, CM-02-013)
- [x] AC-2: `_find_case_actor` deleted; `is_case_actor_identity` removed from authority paths
- [x] AC-3: ratchet test asserts no protocol-logic module branches on either signal — both halves are now enforced (URL shape *and* `Service`-hosting), with a self-test per scanner and regression tests for three demonstrated bypasses
- [x] AC-4: full suite green

**#3193**
- [x] AC-1: the announce two-arm split gates on the role predicate (CM-02-011)
- [x] AC-2: regression test for the bootstrap window (CM-02-012) — existed, now marked
- [x] AC-3: #2626 relationship determined and linked — **not resolved and not
  partially addressed**; its symptoms are already absent on `main` independently
  of this work, so #2626 is left open with a recommendation to attribute the fix
  and close
- [x] AC-4: full suite green

**#3195**
- [x] AC-1: `_find_case_actor_id` documented as delivery address resolution; no URL-shape authority test remains
- [x] AC-2: store guard reclassified as store-consistency / anti-fork (ARCH-24-005)
- [x] AC-3: relevant notes updated
- [x] AC-4: full suite green

**#3273** — all five ACs listed on the issue.

### Triage and execute (`/pr-ship`)

A full triage pass produced 26 findings; the ones that changed code are folded in
above. Two were behaviour regressions this PR had introduced, both now fixed and
pinned by tests that fail without the fix: the ingress-echoed case row read as
local evidence (above), and the ARCH-24-004 ratchet covering only half its own
requirement. Four test-quality findings were confirmed by mutation testing rather
than inspection — most importantly that the role resolution this PR exists to
introduce was *unpinned*, because every fixture in `test_replay.py`,
`test_reject_tree.py` and `test_reject_sync.py` collapsed CASE_MANAGER, store
owner and executing actor onto one id. Bypassing the resolver entirely left 34
tests green. `MANAGER_ACTOR_ID` is now deliberately none of those three, and the
same mutation now fails.

Two further tests were green while asserting nothing: `test_reject_sync`'s replay
test asserted only an announce *count* (one is queued whichever sender is
published), and the announce idempotency test asserted `dl.read(...) is not None`
for a case it had created itself — so it passed whether the announce was applied,
ignored, or rejected, and under the pre-fix code it was silently exercising the
*reject* path. Both now assert the value that discriminates.

In `test_reject_tree.py` the collapse turned out to be *required* rather than
accidental — the guarded announce gates on the executing actor holding the role,
so a distinct manager sends the tree down its skip branch and there is no call to
inspect. That is recorded in a comment pointing at the unit tests that do
discriminate, so the next reader does not "fix" it wrongly.

### Code review

`/code-review high` returned 7 findings. Findings 1, 2 and 5 are resolved by the
revert above; 3 (the overwrite) and 6 (the ratchet was evadable by importing
`CASE_ACTOR_SEGMENT` and rebuilding the suffix — verbatim the retired predicate's
body) are fixed with tests; 7 (a duplicated test) is dropped. Finding 4 was
considered and deliberately kept: the suggested reorder is unavailable because the
pre-seed arm reads the `case_id` `FindCaseActorNode` publishes, and the retired
fallback let the sequence proceed as the *wrong actor*, which is worse than
stopping. The reasoning is recorded in the node's docstring.

Issue #3261 was closed separately as covered — not by this PR, so it is
deliberately absent from the closing list above. It duplicated #3192's AC-2 and
#3193's ACs, and cited a `_find_case_actor` copy in `conditions.py` that no longer
existed; the adjudication is recorded on the issue.

Left open, filed by this work: **#3274** (no trust anchor against a fabricated
roster at genuine first contact), **#3295** (undocumented defensive code masking
defects — second witness, per BW-07-006), **#3296** (`spec-lint` YAML errors do
not name the file), and **#3304** (demo case creation registers no CASE_MANAGER
participant, so `invite_actor_demo` must still resolve an address by hosting
location — attempted here and reverted, because a role-based resolver there would
answer `None` on every call and fall through, reading like coverage while testing
nothing).

**Deferred with review approval: CLP-07-003 (#3282, fix planned in #3301).**
Making the resolver answer unconditionally activated a dormant predicate:
`_validate_canonical_entry` refuses a recorded, non-case-authored entry whose
`payloadSnapshot.actor` equals the resolved authority, so a coordinator that
manages its own case can no longer commit its own assertions. #3301 already plans
the fix — moving actor-identity enforcement to the receive pipeline, where the
inbound `activity.actor_id` is available — so patching the predicate here would
duplicate a decision made elsewhere, and the previous attempt at narrowing it was
reverted as wrong in both directions.

Rather than leave the regression implied, it is pinned:
`test_role_holding_participant_may_assert_on_a_case_it_manages` is a
`strict=True` xfail, so it starts failing as XPASS the moment #3301 lands, which
is the prompt to delete the marker. Nothing else in the suite covers
asserter == role-holder for a non-case-authored signature, because the two tests
that used to were re-shaped to a different-asserter flow.


sei-ahouseholder (commented):
## PR Triage: #3290 — Complete ADR-0088: authority is the CASE_MANAGER role, not a name, URL, or hosting location

**Linked issues**: #3192, #3193, #3195, #3273
**Changed files**: 24 files — core/use_cases, core/behaviors/sync, core/behaviors/case, test/architecture, notes, docs
**CI status**: ✅ passing (22/22 checks)
**Base branch**: `main`

This is a careful, unusually well-documented PR — the revert is genuinely byte-clean, the ACs are honestly adjudicated, and the learnings file correctly identifies the "tightening a resolver wakes dormant checks" pattern. The findings below are mostly consequences of that same pattern that the PR spotted but did not finish chasing.

---

### Findings

|---|---|---|---|---|
| phase8-seeded-verdict-rejects-legitimate-announce-0 | code-review | ❌ FAIL | Seeded fail-closed branch rejects a legitimate CASE_MANAGER's first `Announce` on the HTTP inbox path (ingress pre-stores the case; roster still bare IDs → `expected is None`) | fix-now |
| phase8-clp07003-refuses-role-holder-own-entry-0 | code-review | ❌ FAIL | CLP-07-003 now refuses an ordinary-participant CASE_MANAGER's own non-case-authored entries; disclosed as #3282 but shipped | defer-ask |
| phase10-ratchet-misses-service-hosting-half-0 | test-coverage | ❌ FAIL | ARCH-24-004's `Service`-hosting half has no ratchet — the deleted `_find_case_actor` passes verbatim | fix-now |
| phase10-ratchet-evaded-by-annassign-0 | test-coverage | ❌ FAIL | `_SUFFIX: str = "/actors/case-actor"` defeats the anti-evasion test; +7 more evasions | fix-now |
| phase10-replay-role-resolution-unpinned-0 | test-coverage | ❌ FAIL | Fixtures collapse manager == store owner == executing actor; bypassing the role entirely leaves 34 tests green | fix-now |
| phase10-reject-sync-roster-setup-inert-0 | test-coverage | ❌ FAIL | `test_replay_triggered_when_the_case_manager_is_resolvable` asserts only a count; `or case_id` makes it pass either way | fix-now |
| phase10-announce-idempotency-test-vacuous-0 | test-coverage | ❌ FAIL | `test_idempotent_when_case_already_exists` now exercises the *reject* path while asserting nothing | fix-now |
| phase8-announced-manager-dead-on-wire-shape-0 | code-review | ⚠️ IMPROVE | `_announced_case_manager_id` returns `None` for every adapter-produced announce (roster arrives as bare IDs) | fix-now |
| phase8-seeded-basis-label-stale-link-0 | code-review | ⚠️ IMPROVE | Seeded verdict labelled "resolved CASE_MANAGER" but is link-first; stale bootstrap link mislabels a handoff rejection | fix-now |
| phase8-redundant-case-reads-0 | code-review | ⚠️ IMPROVE | Two `read_case` + a full `ReportCaseLink` scan per announce | fix-now |
| phase10-replay-failure-paths-indistinguishable-0 | test-coverage | ⚠️ IMPROVE | Two bare `FAILURE` assertions; the "unrelated Service" test is satisfied by the case-absent branch | fix-now |
| phase10-announced-manager-attributed-to-untested-0 | test-coverage | ⚠️ IMPROVE | The `attributed_to is None` branch its caller cites as motivation is untested | fix-now |
| phase10-ratchet-file-scoped-exemption-0 | test-coverage | ⚠️ IMPROVE | Whole-file exemption lets a renamed verbatim copy of the deleted predicate pass; `async def` also missed | fix-now |
| phase10-ratchet-docstring-skip-dead-0 | test-coverage | ⚠️ IMPROVE | Docstring-skip machinery is unreachable; two self-test assertions prove nothing | fix-now |
| phase10-resolver-twin-name-not-detected-0 | test-coverage | ⚠️ IMPROVE | Test names `_resolve_case_manager_id` but matches only the bare name | fix-now |
| phase10-ratchet-corpus-excludes-semantic-registry-0 | test-coverage | ⚠️ IMPROVE | `vultron/semantic_registry/` excluded from the corpus with no rationale — it *is* recognition and routing | fix-now |
| phase5-demo-resolves-authority-by-hosting-0 | spec-conformance | ⚠️ IMPROVE | `invite_actor_demo._get_case_actor_id` still routes by `type=="Service" and context==case_id` | fix-now |
| phase9-invite-docstring-stale-0 | notes-currency | ⚠️ IMPROVE | `invite.py:57` still describes "no CaseActor record → no canonical ledger (ADR-0021)" | fix-now |
| phase9-stale-usecase-path-in-note-0 | notes-currency | ⚠️ IMPROVE | Note points at `received/case.py`, which does not exist | fix-now |
| phase9-notes-related-specs-not-updated-0 | notes-currency | ⚠️ IMPROVE | Two notes cite ARCH-24-* without adding `specs/architecture.yaml` to `related_specs` | fix-now |
| phase9-case-actor-terminology-in-note-0 | notes-currency | ⚠️ IMPROVE | `case-communication-model.md` now contradicts itself: new ADR-0088 section vs. ~40 "Case Actor"-as-authority references | fix-now |
| phase9-case-actor-terminology-notes-tree-0 | notes-currency | 🎫 NEW-ISSUE | ~250 occurrences across 19 other notes still name the authority "Case Actor" | defer-ask |
| phase9-test-docstrings-stale-0 | notes-currency | ⚠️ IMPROVE | Docstring names the deleted pending-link path; inert `as_CaseActor` setup | fix-now |
| phase3-pr-body-names-nonexistent-function-0 | pr-body-format | ⚠️ IMPROVE | Body describes `_reject_reason`; the code ships `_authority_verdict` / `admits()` | fix-now |
| phase3-pr-body-false-ratchet-claims-0 | pr-body-format | ⚠️ IMPROVE | "skips docstrings" credits dead machinery; AC-1's `Service`-hosting half is ticked on an untested ratchet | fix-now |
| phase6-adr0088-verification-fields-unfilled-0 | adr-check | ⚠️ IMPROVE | ARCH-24-005 `verification:` empty; ARCH-24-004 / CM-02-012 unnamed though the tests now exist | fix-now |

**Total**: 7 FAIL · 18 IMPROVE · 1 NEW-ISSUE

---

### The two blockers, with reproductions

**1. A legitimate CASE_MANAGER's `Announce(VulnerabilityCase)` is now rejected.** `FastAPIIngressAdapter.parse` pre-stores the announced case into the same actor DataLayer *before* dispatch, so `dl.read_case()` always answers and `seeded=True`. But `read_case` does not expand `case_participants` — they stay bare ID strings until `SeedAnnouncedCaseNode` writes them — so `resolve_case_manager_id` returns `None`, `expected is None`, and the seeded branch's `expected == sender_id` is false. The announce is refused, `SeedAnnouncedCaseNode` never runs, participants/reports/embargo are never persisted, `_link_report_case_links` is skipped, the gap buffer is never drained (SYNC-15-005) — and it is self-perpetuating, because the case row remains present with an unresolvable roster.

This already fires in-tree:

```
uv run pytest test/demo/test_pcr_bootstrap.py -m "" -q -o log_cli=true -o log_cli_level=WARNING
WARNING vultron.core.use_cases.received.actor.announce:announce.py:212
  AnnounceVulnerabilityCase: actor '...case-actor-bootstrap' is not the resolved
  CASE_MANAGER ('None') for case '.../pcr-07-006-ac2' — rejected
```

Confirmed as a regression: restoring `main`'s `announce.py`, `_helpers.py` and `case_actor_identity.py` removes the warning entirely. `main`'s guard was `if case_actor_id is not None and ...`, which accepted on `None`. The test still passes only because its prerequisite assertion is satisfied by the ingress pre-store — so the PCR-07-006 seeding it claims to cover is being skipped.

**2. The ratchet does not enforce half of the requirement it names.** ARCH-24-004 forbids branching on the URL shape **or on `Service`-hosting location**. Restoring the helper this PR deleted, verbatim, under `vultron/core/use_cases/` leaves all 10 tests in the file green — the literal-based prefilter never opens a file mentioning neither `case-actor` nor `CASE_ACTOR_SEGMENT`:

```python
def find_case_actor(dl, case_id):
    for service in dl.list_objects("Service"):
        if getattr(service, "context", None) == case_id:
            return service
    return None
```

And the URL half is evaded by a type annotation, defeating the very test written to close binding indirection:

```python
_SUFFIX: str = "/actors/case-actor"          # ast.AnnAssign, not ast.Assign
def is_the_authority(actor_id):
    return actor_id.rstrip("/").endswith(_SUFFIX)
```

The most realistic re-introduction is not covered either — `actor_id == case_actor_identity()` contains no literal and no alias, so the file is never opened, though the glossary explicitly forbids exactly that comparison.

Separately, mutation testing shows the role resolution this PR exists to introduce is unpinned: replacing the resolved id with `self.actor_id` in `FindCaseActorNode` — never resolving the role at all — leaves 34 tests across three files green, because every fixture sets manager == store owner == executing actor.

### On #3282

The CLP-07-003 consequence is disclosed in the body and filed, which is the right instinct — but it ships a MUST-level regression (a self-managing coordinator can no longer commit its own assertions, a configuration ADR-0088 §7 blesses) behind a filed issue without an approved deferral. That is a Gate 1 decision, so `pr-execute` will present the measured remainder and ask rather than assume.

---

*Triage artifact: `.claude/pr-3290-triage.json`*

sei-ahouseholder (commented):
## PR Execute: #3290 — 27 findings processed

**Final CI status**: ✅ passing (25/25 checks)
**Local suite**: 9997 passed, 399 skipped, 22 xfailed (full run, `-m ""`)
**Linters**: Black, flake8, mypy, pyright, `spec-lint`, `markdownlint`, `mkdocs --strict` all clean

**Outcomes**: 20 fixed · 5 no-change-needed (resolved upstream mid-run) · 1 deferred with approval · 1 skipped with a filed cause

---

### The base branch moved under this PR, twice

Worth reading before the finding table, because it changed what several fixes mean. Between triage and execute, `main` landed **#3303**, which replaced this PR's whole announce-recognition guard with `_sender_is_trusted` — a fail-closed check requiring an explicit locally-derived anchor (a completed `ReportCaseLink`, or a `VultronPendingCaseInbox` written by the invite path). That is strictly stronger than the fix I had just applied, so I took main's design and reverted mine, including deleting the `has_local_participant_roster` helper rather than leaving it dead.

The two changes did not compose for free. Main's fixtures anchored trust on `_find_case_actor_id` **path 4 — the legacy `Service`-hosting scan**, which is exactly what this PR deletes. Six of main's own announce tests failed on the merged tree. Their anchors are migrated to the invite record via a new `_anchor_expected_authority` helper, with the reason recorded at each site. This is the interesting integration risk in this PR and it would not have surfaced from either change alone.

**#3301** also merged, revising CLP-07-003 to say outright that a CASE_MANAGER holding a participant role *is* the correct asserter and not a substitution — so the commit-boundary check now contradicts its own requirement. See the deferral below.

### Two behaviour regressions, both real, both now gone

**A legitimate CASE_MANAGER's first `Announce` was rejected.** The FastAPI ingress adapter pre-stores an inbound activity's nested objects before dispatch, so on the HTTP path the case row is always present — it is the announce itself, echoed back, with a roster of bare IDs resolving to nothing. Reading that as local evidence made the authority fail closed against itself, blocking `SeedAnnouncedCaseNode` and self-perpetuating on retry. It was firing in the already-passing `test/demo/test_pcr_bootstrap.py`, masked because its prerequisite assertion was satisfied by the pre-store. Confirmed absent on `main` before fixing. Now resolved by main's anchor requirement, which never reads that row as evidence.

**The ARCH-24-004 ratchet enforced half its own requirement.** Verified against the real test, not by inspection — the deleted `Service`-hosting resolver could be restored *verbatim* and all 10 tests stayed green, because the literal prefilter never opened a file mentioning neither `case-actor` nor `CASE_ACTOR_SEGMENT`. Three bypasses now have regression tests: the restored resolver, an annotated suffix constant (`_SUFFIX: str = ...`, which defeated the very test written to close binding indirection), and `actor_id == case_actor_identity()` — the likeliest re-introduction, since it names no literal and the glossary forbids it explicitly.

### The role resolution this PR exists to introduce was untested

Mutation testing, not inspection. Every fixture in `test_replay.py`, `test_reject_tree.py` and `test_reject_sync.py` collapsed CASE_MANAGER, store owner and executing actor onto one id — so replacing the resolved address with `self.actor_id`, never resolving the role at all, left **34 tests green**. `MANAGER_ACTOR_ID` is now none of those three and the same mutation fails.

In `test_reject_tree.py` the collapse turned out to be *required*: the guarded announce gates on the executing actor holding the role, so a distinct manager sends the tree down its skip branch and there is no call to inspect. Recorded in a comment pointing at the unit tests that do discriminate, so the next reader does not "fix" it wrongly.

---

### Findings

| # | Outcome | Note |
|---|---|---|
| phase8-seeded-verdict-rejects-legitimate-announce-0 | ✅ fixed | Fixed, then resolved upstream by #3303; the fixture-anchor migration was still needed |
| phase8-clp07003-refuses-role-holder-own-entry-0 | ⏸ deferred (approved) | #3282 → code fix is **#3302**; pinned as a `strict=True` xfail |
| phase10-ratchet-misses-service-hosting-half-0 | ✅ fixed | New `Service`-hosting scanner + self-test |
| phase10-ratchet-evaded-by-annassign-0 | ✅ fixed | `AnnAssign`/tuple/helper/regex/`match` + identity-builder comparison |
| phase10-replay-role-resolution-unpinned-0 | ✅ fixed | Distinct `MANAGER_ACTOR_ID`; mutation now fails |
| phase10-reject-sync-roster-setup-inert-0 | ✅ fixed | Asserts the replay sender (triage mis-named the mutation site) |
| phase10-announce-idempotency-test-vacuous-0 | ➖ no change | Vacuous test gone with main's file |
| phase8-announced-manager-dead-on-wire-shape-0 | ➖ no change | Main removed the announced-roster path outright (#3274) |
| phase8-seeded-basis-label-stale-link-0 | ➖ no change | `_AuthorityVerdict.basis` no longer exists |
| phase10-replay-failure-paths-indistinguishable-0 | ✅ fixed | Assert `feedback_message`; each FAILURE path now discriminated |
| phase10-announced-manager-attributed-to-untested-0 | ➖ no change | Function off the trust path |
| phase10-ratchet-file-scoped-exemption-0 | ✅ fixed | Exemption line-scoped to `case_actor_identity`; `async def` covered |
| phase10-ratchet-docstring-skip-dead-0 | ✅ fixed | Dead plumbing removed; immunity documented as structural |
| phase10-resolver-twin-name-not-detected-0 | ✅ fixed | Matches the underscore-prefixed twin too |
| phase10-ratchet-corpus-excludes-semantic-registry-0 | ✅ fixed | `semantic_registry` added; exclusions justified |
| phase5-demo-resolves-authority-by-hosting-0 | ⚠️ skipped | Attempted, **reverted on evidence** → **#3304** |
| phase9-invite-docstring-stale-0 | ✅ fixed | `None` = no resolvable address, not the ADR-0021 reading |
| phase9-stale-usecase-path-in-note-0 | ✅ fixed | Path corrected; embedded sample no longer shows the retired guard |
| phase9-notes-related-specs-not-updated-0 | ✅ fixed | `related_specs`/`related_notes` updated (NF-06) |
| phase9-case-actor-terminology-in-note-0 | ✅ fixed | ~40 sites incl. the self-contradicting description |
| phase9-case-actor-terminology-notes-tree-0 | ✅ fixed | Full pass — **Closes #3306** |
| phase9-test-docstrings-stale-0 | ✅ fixed | Deleted-path docstring; inert fixtures re-anchored |
| phase8-redundant-case-reads-0 | ➖ no change | Lived in the replaced function |
| phase3-pr-body-names-nonexistent-function-0 | ✅ fixed | `_reject_reason` did not exist |
| phase3-pr-body-false-ratchet-claims-0 | ✅ fixed | Docstring-skip claim dropped; AC-1/AC-3 restated |
| phase6-adr0088-verification-fields-unfilled-0 | ✅ fixed | ARCH-24-004/005, CM-02-011/012 now name their tests |
| ci-docs-link-broken-by-base-rename-0 | ✅ fixed | Found in the CI loop; pre-existing on main, blocked this PR |

### One skip worth reading

`invite_actor_demo._get_case_actor_id` still resolves an address by hosting location. I wrote the role-based replacement and its tests passed — then probed and found it returned `None` on **every** call: at that point the case roster holds only the finder with an empty `caseRoles`, because `trigger/create-case` registers no CASE_MANAGER participant. The tests passed only because they never assert the invite sender, so it silently fell through to the hosting scan. Shipping that would have been a permanently dead branch that reads as coverage, so it is reverted, documented at the call site, and the cause is filed as **#3304** with the probe output.

### One correction to triage

My measured remainder for the notes-terminology pass said ~250 occurrences across 19 files; I had counted only files with 5+ occurrences, and the real spread was 42 files. The conversion was per-site as described — kept wherever the name denotes the concrete actor, its container, or the provisioning endpoint, since a blanket rename would have been wrong in those files.

---

*Execute artifact: `.claude/pr-3290-execute.json`*
*Next step: `/pr-verify` (or the remainder of `/pr-ship`).*

sei-ahouseholder (commented):
## PR Verify: #3290 — ⚠️ GAPS-FOUND

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, no conflict markers at HEAD
**CI status**: ✅ passing — 23 pass, 2 skipped, 0 failing (incl. 4 Demo Integration + 4 Invariant Harness, so `integration_tests_run` is satisfied)
**Artifact integrity**: all 26 triage findings have a result; execute added 1 CI-loop finding (additive, not a gap)

The merge gate and CI gate both pass. The verdict is `GAPS-FOUND` because spot-checking execute's claims **against HEAD** — rather than against the commit diffs — found four that do not hold. Three are mine to own: two claims that the base-branch merge invalidated, and one issue reference that went stale.

---

### Confirmed (23 of 27)

All ratchet work, the replay/reject-sync fixture and assertion fixes, the notes and spec changes, the PR-body corrections and the CI docs-link fix are present at HEAD. Notably re-confirmed at HEAD rather than by commit: `_docstring_nodes` is gone (0 references), `_PROVISIONING_FUNCTIONS` replaced the file-scoped exemption, `semantic_registry` is in the scanned corpus, and both spec files name the tests they claim to.

Four `no_change_needed` claims were audited rather than taken on trust, and three hold: `_announced_case_manager_id`, `_AuthorityVerdict` and `_authority_verdict` have **zero** references anywhere under `vultron/` or `test/`, so the findings that lived on them are genuinely moot.

### Gap 1 — `phase10-announce-idempotency-test-vacuous-0` → **UNRESOLVED**

Recorded as `no_change_needed` on the grounds that main's #3303 replaced the test file and the vacuous test went with it. **It did not.** At HEAD, `test/core/use_cases/received/actor/test_announce.py:170`:

```python
def test_idempotent_when_case_already_exists(self, dl, event, case, case_actor):
    """MV-10-004: A second Announce for an existing case is a no-op."""
    dl.create(case_actor)
    dl.create(case)
    AnnounceVulnerabilityCaseReceivedUseCase(dl, event).execute()
    result = dl.read(_CASE_ID)
    assert result is not None
```

The assertion is still trivially satisfied by the test's own `dl.create(case)`. Worse than when triage filed it: its anchor is `dl.create(case_actor)` — an `as_CaseActor` with `context`, i.e. the hosting signal **this PR deletes** — so under main's fail-closed `_sender_is_trusted` the announce is now *rejected*, and the test passes anyway. It is the exact defect class the finding named, one merge later.

### Gap 2 — `phase9-case-actor-terminology-in-note-0` → **UNRESOLVED (partial)**

`notes/case-communication-model.md` was taken to zero occurrences, then the merge with `main` reintroduced one at line 346, in content added by the ADR-0089 birth-sequence work:

```text
# ✅ CORRECT — the Case Actor advances the invitee's RM state through the sole
```

Authority-functional usage, so it needs converting like the other ~40.

### Gap 3 — `phase8-clp07003-refuses-role-holder-own-entry-0` → stale issue reference

The deferral is legitimate and reviewer-approved, but the artifact records **#3282**, which is **CLOSED** — it was closed by #3301, the planning PR. The open tracker for the code fix is **#3302**. A deferral pointing at a closed issue is exactly the "filing is not tracking" failure the doctrine warns about, so the reference must be corrected.

### Gap 4 — new: stale docstring created by the merge

`_sender_is_trusted` (main's function) documents its first anchor as:

```text
1. An established ``_find_case_actor_id`` anchor (ReportCaseLink path 1–4):
```

`_find_case_actor_id` had four paths when that was written; **this PR reduces it to two** (the recorded link, then the role). The docstring is accurate on `main` and inaccurate on this branch — a defect the two changes create only in combination, which is the same category as the six fixtures already re-anchored during execute.

---

### Not blocking

| Item | State |
|---|---|
| `phase5-demo-resolves-authority-by-hosting-0` (skipped) | ✅ #3304 open, cause documented at the call site |
| `phase9-case-actor-terminology-notes-tree-0` (fixed) | ✅ #3306 open, `Closes` bullet present |
| CLP-07-003 strict xfail | ✅ XFAIL as intended; flips to a failing XPASS when #3302 lands |

---

**Verdict**: `GAPS-FOUND`. Artifacts retained at `.claude/pr-3290-{triage,execute}.json`.

Gaps 1, 2 and 4 are small, local, and the context is loaded — fixing them now and re-verifying rather than handing them back.

sei-ahouseholder (commented):
## PR Verify: #3290 — ✅ READY-TO-MERGE

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, no conflict markers at HEAD
**CI status**: ✅ passing — 23 pass, 2 skipped, 0 failing, including 4 Demo Integration + 4 Invariant Harness jobs
**Local suite**: 9997 passed, 399 skipped, 22 xfailed (`-m ""`)
**Findings**: 28 processed — 22 fixed · 4 no-change-needed · 1 deferred (approved) · 1 skipped with a filed cause
**FAIL findings**: 7 — 6 CONFIRMED at HEAD, 1 deferred with reviewer approval

This is the second verify pass. The [first one](https://github.com/CERTCC/Vultron/pull/3290#pullrequestreview-5237473614) returned `GAPS-FOUND`; all four gaps are now closed and re-checked against HEAD.

---

### The four gaps from the previous pass

| Gap | State |
|---|---|
| `test_idempotent_when_case_already_exists` still vacuous at HEAD | ✅ replaced by `test_redelivery_of_the_same_announce_is_stable` (0 references to the old name) |
| One "Case Actor" reintroduced into `case-communication-model.md` by the merge | ✅ 0 occurrences |
| CLP-07-003 deferral cited **#3282**, closed | ✅ retargeted to **#3302** (OPEN) |
| `_sender_is_trusted` docstring cited `_find_case_actor_id` "path 1–4" | ✅ names the two surviving paths and why the others went |

The first gap was the one worth catching. It was recorded `no_change_needed` on the assumption that main's #3303 had replaced the file; it had not, and the test had quietly got *worse* — its anchor was the very hosting `Service` this PR deletes, so under main's fail-closed guard the announce was being rejected on both calls while `assert dl.read(...) is not None` passed on a row the test created itself. The replacement is mutation-checked: removing its anchor now fails it, where the old version passed under exactly that condition.

### Verification method

Every claim was checked **at HEAD**, not against its commit diff — which is what surfaced the two claims the base-branch merge had invalidated. All commit refs resolve on the branch (`MISSING-COMMIT`: none). The four surviving `no_change_needed` claims were audited rather than trusted: `_announced_case_manager_id`, `_AuthorityVerdict` and `_authority_verdict` have zero references anywhere under `vultron/` or `test/`, so the findings that lived on them are genuinely moot.

### Open items, deliberately

**Deferred with approval — CLP-07-003 → #3302 (OPEN).** The commit-boundary check refuses a legitimate self-assertion by a CASE_MANAGER that also holds a participant role. #3301 merged mid-run and revised CLP-07-003 to say outright that this is the correct asserter and not a substitution, so the code now contradicts its own requirement. Not left implied: pinned by a `strict=True` xfail that becomes a failing XPASS the moment #3302 lands, which is the prompt to remove the marker. Nothing else in the suite covers that shape.

**Skipped with a filed cause — #3304 (OPEN).** `invite_actor_demo` still resolves an address by hosting location. The role-based replacement was written, passed the tests, and was then reverted after probing showed it returned `None` on every call — the demo's `trigger/create-case` path registers no CASE_MANAGER participant, so it fell through to the hosting scan while reading as coverage. The call site documents this; #3304 tracks the cause.

**Closes #3306** — the ADR-0088 prose pass across `notes/`, filed as the excursion it was and closed by this PR.

### Base branch moved twice during the run

Recorded here because it shaped the outcome: `main` landed #3303 (replacing this PR's announce guard with a stronger fail-closed one), #3301 (the CLP-07-003 spec revision) and #3292 (a docs reorganisation) mid-pipeline. Main's guard was adopted over the fix written here, but the two changes did not compose for free — main's fixtures anchored trust on the `Service`-hosting path this PR deletes, so six of main's own tests failed on the merged tree until their anchors were migrated. That interaction is the main integration risk in this PR and neither change would have revealed it alone.

---

**Verdict**: `READY-TO-MERGE`

Issue #3324, opened 2026-09-17
Attribute metadata loader failures to their source file via one shared helper


sei-ahouseholder:
## Summary

Every loader under `vultron/metadata/` walks a directory, parses each file, then
validates the parsed data against a Pydantic model. Both steps can fail, and
today what an author sees depends on which directory they edited: the ADR and
history loaders wrap both steps (in as many message shapes as there are copies),
while the spec and notes loaders wrap neither. In `specs/`, a YAML syntax error
therefore reaches the user as a raw PyYAML traceback whose position reads
`"<unicode string>", line 404` — a location in a source that is not on disk.

Extract one shared helper, adopt it at every loader site, and delete the
hand-written copies. Requirements: MS-17-001/002/003, SR-03-008, SR-03-009,
NF-03-006. Planning context and confirmed scope decisions: #3296 and its docs PR.

## Acceptance Criteria

- [ ] AC-1: `vultron/metadata/file_loading.py` provides the shared attribution
      helper (MS-17-003) covering both halves: parsing a YAML or frontmatter file,
      and validating parsed data against a Pydantic model. Failures raise a
      `ValueError` subclass whose message is `path:line:col — <detail>` when the
      parser supplies a position mark and `path — <detail>` when it does not
      (MS-17-001). The position comes from the exception's `problem_mark`; no
      `<unicode string>` marker survives into the message.
- [ ] AC-2: The message carries a cause hint keyed on PyYAML's `problem` text.
      The `mapping values are not allowed in this context` case MUST render the
      hint that a plain scalar containing `": "` needs quoting or a block scalar —
      this is the most common way the fault is introduced while editing spec prose,
      and PyYAML's own message never suggests it. Other hints are optional; a
      `problem` with no table entry emits no hint rather than a wrong one.
- [ ] AC-3: `load_registry` routes both `yaml.load` and `SpecFile.model_validate`
      through the helper (SR-03-008), and reports **every** failing file in one
      error rather than stopping at the first, with the individual failures
      recoverable from the exception as structured data rather than only by
      parsing its message (SR-03-009). No parser exception escapes (MS-17-002).
- [ ] AC-4: `notes/loader.py` routes both its `frontmatter.load` and its
      `NotesFrontmatter.model_validate` through the helper (NF-03-006).
- [ ] AC-5: The hand-written wrappers in `adr/loader.py` (`load_adr_post` and the
      `model_validate` guard), `history/readme_gen.py::_parse_entry`, and
[...]

PR #3323, opened 2026-09-17
docs: plan issue #3296 — spec-lint YAML failures must name the file

sei-ahouseholder:
- Closes #3296

## Summary

Records the requirements for a fix to #3296: a YAML syntax error in any
`specs/*.yaml` makes `spec-lint` exit through a raw PyYAML traceback whose
position (`"<unicode string>", line 404`) names the string being parsed rather
than the file, so it locates nothing across the spec corpus. Planning found the
same gap in the sibling loaders and found that the wrapper which fixes it had
already been written — independently, more than once, in different shapes —
elsewhere in the same package.

## Changes

- **`specs/meta-specifications.yaml`**: new **MS-17** group, *Metadata Loader
  Failure Attribution*. MS-17-001 requires a loader that walks files to name the
  failing file as `path:line:col`; MS-17-002 forbids a parse or validation
  exception escaping as an unhandled traceback; MS-17-003 requires the
  attribution to come from one shared helper (`vultron/metadata/file_loading.py`)
  rather than a per-loader copy; MS-17-004 fixes the exception type as a
  `ValueError` subclass local to the tooling layer, and draws the boundary that
  keeps the `VultronError` hierarchy out of `vultron/metadata/`. Carries
  `lint_suppress: [phantom_path_ref]`, the documented opt-out for a spec-first
  requirement naming a file yet to be created — remove it in the PR that creates
  the module.
- **`specs/error-handling.yaml`**: a `rationale` on **EH-01-002** recording the
  same boundary from the other side. SR-03-009 makes the shared helper's error the
  first custom exception class in the tooling layer, which is what brings
  EH-01-002 into contact with it; since `VultronError` derives from `Exception`
  rather than `ValueError`, obeying it there would put the new error back outside
  every caller's `except (ValidationError, ValueError)` guard — the exact defect
  MS-17 removes.
- **`specs/spec-registry.yaml`**: **SR-03-008** applies MS-17 to `load_registry`,
  covering the `SpecFile` validation as well as the YAML parse — a Pydantic error
  reports `groups.0.specs.0.kind`, which locates a requirement no better than a
  bare line number, and it is the more common fault. **SR-03-009** requires every
  failing file to be reported in one error with the individual failures
  structured on the exception, per EH-07-001. **SR-04-002** gains a
  `verification` recording that satisfying its existing "invalid YAML structure"
  hard-error clause with a traceback does not count.
- **`specs/notes-frontmatter.yaml`**: **NF-03-006** applies the same rule to the
[...]

Issue #3296, opened 2026-09-16
spec-lint reports a malformed spec YAML as a raw traceback that never names the file


sei-ahouseholder:
## Symptom

A syntax error in any `specs/*.yaml` makes `spec-lint` exit with an unhandled
`yaml.scanner.ScannerError` and a ~30-line PyYAML/Cython traceback. The location
it reports identifies the *string* it was parsing, not the file:

```text
yaml.scanner.ScannerError: mapping values are not allowed in this context
  in "<unicode string>", line 404, column 43
```

With 40+ files in `specs/`, "line 404" alone does not locate the error, and the
traceback frames are all inside `yaml/_yaml.pyx`, so nothing in the output names
the offending spec.

## Reproduction

Add a plain (unquoted) scalar containing `": "` to any spec entry — the most
likely way to hit this, since prose in a `statement:` or `verification:` field
naturally wants a colon:

```yaml
  - id: CLP-07-099
    priority: MUST
    kind: protocol
    statement: A plain scalar with a colon: this breaks the parser
```

Then `PYTHONPATH= uv run spec-lint`. Verified against
`specs/case-ledger-processing.yaml` on `a32494ab7`.

## Why it matters

This is the failure mode an agent or contributor is most likely to cause while
*editing spec prose*, which is the common case — and the tool's answer gives no
filename. It cost a cycle during #3195's spec amendment. `spec-lint` is otherwise
good at attributing findings (every `[WARN]` names its requirement ID), so this is
an isolated gap rather than a pattern.

## Suggested fix

Wrap the per-file load so the file being parsed is named, and suppress the
traceback in favour of a `[ERROR]` line consistent with the existing `[WARN]`
output — e.g.:

```text
[ERROR] specs/case-ledger-processing.yaml:404:43 — mapping values are not allowed
        in this context (a plain scalar containing ': ' needs a block scalar or
        quoting)
```

The parenthetical hint is worth including: `": "` in a plain scalar is by far the
most common cause, and the raw PyYAML message does not suggest it.

## Notes

Found while amending CLP-07-003's `verification:` field during #3195 (PR #3290).
Not fixed there because it is a different subsystem and that PR already closed
four issues; folding an unrelated tooling change in would have reduced rather than
improved its clarity.

sei-ahouseholder:
Claimed on branch `plan/3296-spec-lint-yaml-error-names-file`.

sei-ahouseholder:
✅ Planned.

[...]

Issue #2329, opened 2026-08-17
Graduated warning thresholds for test-enforced limits (warn before hard fail)


sei-ahouseholder:
## Summary

Introduce a reusable "graduated threshold" helper for test-enforced limits: emit an
advisory **warning** as a value approaches its limit, and fail only at the hard limit.
Adopt it across the size and count ratchets (AGENTS.md line targets, the 500-line module
cap, the uncovered-protocol-spec ceiling). As part of the design, decide whether to keep
pytest's current `filterwarnings = ["error"]` policy, which today makes an advisory
warning impossible.

Broadened from the original BT-node-specific framing (soft CI annotation at 90% of the
500-line cap) after PR #3014 hit the same problem for `AGENTS.md` line counts — the
mechanism should be written once and reused, not reinvented per limit.

## Motivation

A hard limit with no runway is discovered at the worst moment: mid-PR, on a change that
had nothing to do with the limit. The author then either explodes their scope to
decompose a module, or the limit quietly goes unenforced.

Two live examples:

- **Root `AGENTS.md` reached 1166 lines against a documented 400-line target** — 2.9× —
  because the target was advisory with no check (#2954, fixed in PR #3014). It is now
  enforced at exactly 400, so the *next* addition fails with zero warning.
- **The 500-line module cap (CS-18-001–004, BTND-07-003) has no test at all**, and
  **45 modules currently exceed it**, topped by
  `vultron/core/behaviors/case/case_proposal_received_tree.py` at **2017 lines** (4× the
  cap). A further 14 modules sit in the 450–500 band that would warn today.

Both are the same failure: a documented MUST with no graduated signal. A shared helper
turns "documented target" into "visible pressure well before a blocking failure", and
gives CI the option to promote a warning tier to an error later without editing tests.

### Design decision to settle first: is `filterwarnings = ["error"]` the right policy?

`pyproject.toml` currently sets `filterwarnings = ["error", ...]`. This may be
over-strict, and it directly blocks the feature. Empirically verified while drafting
this (probe tests, since removed):

| Probe | Result |
|---|---|
| Unregistered `UserWarning` subclass emitted in a test | **raises** — an advisory becomes a failure |
| `UnknownSpecIdWarning`, registered `always::` **before** `"error"` | **raises** — the entry is a no-op |
| Same class with its filter applied **after** `"error"` | warns correctly |
| Subclass of an exempted class | inherits the exemption |

Two findings worth acting on regardless of which policy is chosen:

[...]

PR #3336, opened 2026-09-17
fix(test): make a spec-registry load failure visible instead of silent

### LINKED ISSUE #3331: Spec-marker ID validation silently disabled when the spec corpus fails to load
## Symptom

`pytest_collection_modifyitems` in `test/conftest.py` wraps the spec-registry
load in a bare `except Exception: return`. If any file in `specs/` fails to
load, the `@pytest.mark.spec` ID validation required by **SR-05-002** is
disabled for the entire test session, with no warning, no message, and no
change in exit code. The suite reports a clean pass.

The docstring compounds it by describing a different behaviour than the code
has: it says the hook "skips silently when no YAML files exist in `specs/`",
but corpus *absence* is already handled by the two guards above the `try`
(`if not spec_dir.is_dir()` and `if not registry.files`). What the `except`
actually absorbs is corpus *corruption*.

## Measurement

Instrumented the `except` to print what reaches it, per
`plan/incoming/learnings/20260914-3217-measure-what-a-permissive-fallback-absorbs.md`:

- **Healthy corpus, full unit suite**: fired **0 times**. The clause absorbs
  nothing in normal operation, so it is not load-bearing.
- **One corrupted spec file**: fired once, absorbing
  `ScannerError: mapping values are not allowed in this context`. A test
  carrying a deliberately bogus `@pytest.mark.spec("BOGUS-99-999")` marker
  emitted `UnknownSpecIdWarning` on the healthy corpus and **nothing at all**
  on the corrupted one.

Single-cause result: the clause is masking a fault rather than handling a real
compatibility case.

## Why not simply let it propagate

`filterwarnings` in `pyproject.toml` sets `"error"`, with an explicit
`always::vultron.metadata.specs.UnknownSpecIdWarning` exemption so the spec-ID
gate stays non-blocking by design. Letting the load error propagate — or
emitting a warning without a matching exemption — makes a malformed spec file
abort the whole session, so you could not run the tests that diagnose the
malformed spec. The degradation should stay survivable but become **visible**.

## Acceptance Criteria

- [ ] AC-1: A spec-registry load failure during collection emits a visible,
      non-blocking warning naming the underlying cause, rather than returning
      silently. New warning class alongside `UnknownSpecIdWarning` in
      `vultron/metadata/specs/`, exported in `__all__`.
- [ ] AC-2: `pyproject.toml` `filterwarnings` gains an `always::` entry for the
      new warning, mirroring the existing `UnknownSpecIdWarning` entry, so the
      warning does not become a session-aborting error under the `"error"` rule.
- [ ] AC-3: The `except` narrows from bare `Exception` to the failure types
      `load_registry` can actually raise, so an unexpected exception type — a
      bug in the loader rather than a bad spec file — still surfaces instead of
      being swallowed.
- [ ] AC-4: The docstring states what the code does: the guards handle an absent
      or empty corpus; the `except` handles an unloadable one and warns.
- [ ] AC-5: Regression tests cover both halves — that a load failure produces
      the warning, and that a healthy corpus still validates markers and
      produces no such warning. A test asserting only the happy path does not
      close this.

## Notes

Found while planning #3296 (spec-loader failure attribution) and deliberately
kept out of that PR's scope. The two interact: once #3324 makes `load_registry`
raise a documented `ValueError` subclass for every load fault, AC-3's exception
tuple can drop its `yaml.YAMLError` term.

### LINKED ISSUE #3380: filterwarnings exemptions listed before "error" are no-ops: SR-05-002's non-blocking warning has always aborted collection
## Symptom

`pyproject.toml`'s `filterwarnings` lists both spec-gate exemptions **before**
`"error"`:

```toml
filterwarnings = [
    "always::vultron.metadata.specs.UnknownSpecIdWarning",
    "always::vultron.metadata.specs.SpecRegistryUnavailableWarning",
    "error",
    ...
]
```

That ordering is a no-op. pytest applies ini `filterwarnings` in list order
through `warnings.filterwarnings()`, which **inserts each filter at index 0** —
so a **later** entry outranks an earlier one. `"error"` therefore wins over both
exemptions, and each warning is escalated to an exception.

Measured under the real config (`warnings.filters`, index 0 = highest
precedence):

```text
0..2  ignore  DeprecationWarning
3     error   Warning                          <-- wins
4     always  SpecRegistryUnavailableWarning
5     always  UnknownSpecIdWarning
```

Two consequences:

1. **SR-05-002's "non-blocking" guarantee has never held.** The marker gate
   raises `UnknownSpecIdWarning` as an exception. Because it is raised from
   `pytest_collection_modifyitems`, a single test carrying an unknown spec ID
   aborts **collection** — zero tests run — rather than emitting an advisory
   warning.
2. **PR #3336 reproduced the same mistake** for `SpecRegistryUnavailableWarning`,
   which converted the silent gate-skip it set out to fix into a hard session
   abort on a malformed corpus — strictly worse than the bug it was fixing, and
   the exact outcome that PR's own docstrings said must not happen.

## Why it went unnoticed

Every warning assertion in the test suite runs inside `pytest.warns` or
`warnings.catch_warnings`, and **both replace the ini filters**. No in-session
test can observe the escalation. PR #3336 even added a test class specifically to
pin the invariant (`TestWarningIsNotEscalated`), but it asserted the entry's
*position in the ini list* — encoding the inverted model — so it passed while the
property it claimed to guarantee was false.

The rule was also recorded backwards in the original implementation notes:
`plan/history/2604/implementation/LEGACY-2026-04-27-task-sr-sr-1-sr-5-spec-registry-tests-pytest-mar.md:71`
— *"placed BEFORE `error` so the specific rule takes precedence (Python prepend
semantics)"*.

#2329 measured the correct rule and explicitly named the ineffective
`always::UnknownSpecIdWarning` entry as a latent bug, but that finding lived only
in an issue body, so #3331 was written on the wrong premise and #3336 implemented
it that way.

## Acceptance Criteria

- [x] AC-1: Both `always::` entries are listed **after** `"error"` in
      `pyproject.toml`, and the accompanying comment states the actual precedence
      rule rather than the inverted one.
- [x] AC-2: The invariant is verified **behaviourally** — the warning is emitted
      in a `pytester` sub-session fed the project's real `filterwarnings` list, so
      an escalation is observable. A position-only assertion does not close this.
      Verified non-vacuous: re-introducing the misordering fails the new tests.
- [x] AC-3: SR-05-002 is corrected. It specified a `PytestWarning`, but nothing
      under `vultron/` imports pytest and the code has always raised a
      `UserWarning` subclass; conforming to the old wording would pull a test
      dependency into production. The "non-blocking" clause is split out to a new
      SR-05-007 that states the ordering requirement and mandates behavioural
      verification.
- [x] AC-4: The rule is recorded durably where an implementer will meet it —
      `test/AGENTS.md` § "Pytest Collection Hygiene" and
      `notes/testing-pitfalls.md` — not only in an issue body. This is the step
      whose absence let the finding be rediscovered twice.

## Notes

Surfaced by triage of PR #3336, which fixes it. Prior witness: #2329 (still open;
its "is `filterwarnings = [\"error\"]` the right policy?" design question is
partly settled once this lands — option (a), keep `error` and exempt named
classes *after* it, is now demonstrated working in-tree).

sei-ahouseholder:
- Closes #3331
- Closes #3380

## Summary

The spec-marker gate in `test/conftest.py`'s `pytest_collection_modifyitems`
wrapped its registry load in a bare `except Exception: return`, so one malformed
file in `specs/` disabled the SR-05-002 `@pytest.mark.spec` ID validation for the
whole session — no warning, no message, no change in exit code. The suite
reported a clean pass with the gate down.

Review of the first cut of this PR then found that the mechanism meant to keep the
new warning non-blocking **does the opposite**, and had been doing so for the
pre-existing warning all along (#3380). Both halves are fixed here.

## Measurement first

Per `plan/incoming/learnings/20260914-3217-measure-what-a-permissive-fallback-absorbs.md`,
the clause was instrumented before anything was changed:

| Condition | What the clause absorbed |
|---|---|
| Healthy corpus, full unit suite | **0 times** — not load-bearing |
| One corrupted spec file | `ScannerError: mapping values are not allowed in this context` |

With the corpus corrupted, a test carrying a deliberately bogus
`@pytest.mark.spec("BOGUS-99-999")` emitted `UnknownSpecIdWarning` on the healthy
corpus and **nothing at all** on the corrupted one. Single-cause result: the
clause was masking a fault, not handling a compatibility surface.

## Why a warning rather than letting it propagate

`pyproject.toml` sets `filterwarnings = ["error", ...]`. Letting the load error
propagate — or adding a warning without an *effective* exemption — would make a
malformed spec file abort the whole session, so you could not run the tests that
diagnose the malformed spec. The degradation has to stay survivable while
becoming visible.

## The filter-ordering bug (#3380)

Making a warning non-blocking requires an `always::` entry that outranks
`"error"`, and **outranking means being listed after it**. pytest applies ini
`filterwarnings` in list order through `warnings.filterwarnings()`, which inserts
each filter at index 0 — so a *later* entry wins. Measured under the real config:

```text
warnings.filters (index 0 = highest precedence)
  3  error   Warning                          <-- wins
  4  always  SpecRegistryUnavailableWarning
  5  always  UnknownSpecIdWarning
```

Both entries sat *before* `"error"` and were therefore no-ops. Consequences:

- **SR-05-002's "non-blocking" guarantee has never held.** Because the warning is
  raised from a collection hook, an unknown spec ID aborted *collection* — zero
  tests run — rather than warning.
- This PR's first cut reproduced the same mistake for the new warning, converting
  the silent gate-skip into a hard session abort.

Nothing in the suite could see it: `pytest.warns` and `catch_warnings` both
**replace** the ini filters, so the first cut's `TestWarningIsNotEscalated` class
passed while asserting the entry's *position* under the inverted model.

#2329 had already measured the correct rule and named the ineffective entry as a
latent bug, but that finding lived only in an issue body — which is why it is now
recorded in `test/AGENTS.md` and `notes/testing-pitfalls.md` and made normative as
SR-05-007.

## Changes

- **`vultron/metadata/specs/__init__.py`**: adds `SpecRegistryUnavailableWarning`
  and `warn_spec_registry_unavailable()`, which names the spec directory, the
  underlying cause, and the requirement whose gate went down. Exported in
  `__all__` so a filter can target it by type.
- **`test/conftest.py`**: narrows the `except` to `_REGISTRY_LOAD_ERRORS`
  (`ValueError`, `OSError`, `yaml.YAMLError`) and warns instead of returning
  silently. An unexpected exception type is a bug in the loader rather than a bad
  spec file, and now surfaces instead of degrading to a warning. The docstring now
  states the three early-return cases and why they are not interchangeable — it
  previously claimed the hook "skips silently when no YAML files exist," which the
  `spec_dir.is_dir()` guard above the `try` and the `registry.files` guard below it
  already handle between them.
- **`pyproject.toml`**: moves `"error"` to the head of `filterwarnings` so both
  `always::` spec-gate entries sit *after* it and actually apply. Fixes #3380; the
  comment now states the real precedence rule instead of the inverted one.
- **`specs/spec-registry.yaml`**: corrects SR-05-002, which specified a
  `PytestWarning` — nothing under `vultron/` imports pytest, so conforming to that
  wording would pull a test dependency into production code; the implementation has
  always raised a `UserWarning` subclass. Adds **SR-05-006** (the three
  non-interchangeable early-return cases) and **SR-05-007** (the filter-ordering
  requirement, and that it MUST be verified behaviourally rather than by asserting
  a position in the ini list).
- **`test/AGENTS.md`, `notes/testing-pitfalls.md`**: record the ordering rule
  durably. Its absence from any durable location is why the finding was
  rediscovered twice. `test/AGENTS.md` keeps a pointer to stay inside its 200-line
  ratchet.
- **`test/metadata/specs/test_spec_marker_gate.py`** *(new, 21 tests)*: covers the
  warning firing for each caught failure type, the message naming its cause and
  `SR-05-002`, that markers really are left unvalidated (so the warning is not
  claiming something untrue), that the timeout half of the hook still runs, that an
  unexpected type propagates, that the two *silent* early returns (absent and empty
  corpus) stay silent, that the healthy path still validates markers, and — via a
  `pytester` sub-session fed the project's real filter list — that neither warning
  is escalated.

## Verification

- **8771 unit tests pass** (21 new), 399 skipped, 19 xfailed, 5557 subtests. Exit 0,
  measured after merging `origin/main`.
- Black, flake8, mypy, pyright, `spec-lint`, `markdownlint` clean.
- **Mutation-checked, both defects:**
  - Reverting `test/conftest.py` to the bare `except` fails **7 of the 21** tests.
  - Re-introducing the `filterwarnings` misordering fails **3** tests, including
    both `pytester` sub-session cases — so the new behavioural test is not vacuous.
    This is the check the first cut lacked.
- The new warning fires 0 times on the real corpus.
- [x] AC-1 (#3331): non-blocking warning naming the cause; new class beside
      `UnknownSpecIdWarning`, exported in `__all__`.
- [x] AC-2 (#3331): matching `always::` entry in `filterwarnings` — and, per
      #3380, positioned so it actually takes effect.
- [x] AC-3 (#3331): `except` narrowed from bare `Exception` to the types
      `load_registry` can raise.
- [x] AC-4 (#3331): docstring states what the code does.
- [x] AC-5 (#3331): regression tests cover both halves, plus the silent pair.
- [x] AC-1–AC-4 (#3380): entries reordered, behavioural verification added,
      SR-05-002 corrected with SR-05-006/007 added, rule recorded durably.

## Notes

Found while planning #3296 and deliberately kept out of that PR (#3323), which is
docs-only. The two interact: once #3324 makes `load_registry` raise a documented
`ValueError` subclass for every load fault, `_REGISTRY_LOAD_ERRORS` can drop its
`yaml.YAMLError` term — which is why the tuple's contents are pinned
behaviourally rather than by an exact-set assertion.

The new test file sits under `test/metadata/specs/` rather than beside
`test/conftest.py` because asserting the gate *fires* requires naming a spec ID
that does not resolve, and SR-04-008 hard-errors on a phantom spec-ID citation in
any Python file outside that directory (`_PHANTOM_ID_ALLOWLIST_DIRS`).
`test_real_specs.py` caught this during development, which is the guard working as
intended.

Also noted during review, tracked rather than fixed here: `test/` sits outside
CS-23-001's `vultron/` scope and outside both open ratchet-extension issues, yet
this bug is the clearest case yet of a broad catch in test infrastructure masking a
MUST-level gate. Recorded on #3347 with the 18 affected sites.


sei-ahouseholder (commented):
## PR Triage: #3336 — fix(test): make a spec-registry load failure visible instead of silent

**Linked issues**: #3331 (Spec-marker ID validation silently disabled when the spec corpus fails to load)
**Changed files**: 4 files — metadata/specs, test-infrastructure, pytest-collection
**CI status**: ✅ passing (24/24)
**Base branch**: `main`

---

### ⛔ Headline: the fix currently inverts its own goal

`filterwarnings` precedence runs the **opposite** direction to what this PR (and #3331 AC-2, and the pre-existing entry) assumes. pytest applies each ini entry via `warnings.filterwarnings()`, which **inserts at index 0** — so **later entries win**.

Verified directly against the real repo config:

```text
warnings.filters (index 0 = highest precedence)
  3  error   Warning                          <-- wins
  4  always  SpecRegistryUnavailableWarning
  5  always  UnknownSpecIdWarning
```

Emitting either warning under the real config **raises**. So on a malformed corpus the new warning escalates to an exception inside `pytest_collection_modifyitems` and aborts collection — the precise outcome AC-2, the new `pyproject.toml` comment, and the PR's own docstrings all say must not happen, and strictly worse than the silent skip being replaced.

All 17 new tests miss it because every warning assertion runs under `pytest.warns` / `catch_warnings`, which **replaces** the ini filters. Confirmed fix: move both `always::` lines *below* `"error"` — both then fire non-blocking and land in the warnings summary.

---

### Findings

|---|---|---|---|---|
| phase8-filterwarnings-order-inverted-0 | code-review | ❌ FAIL | `always::SpecRegistryUnavailableWarning` sits above `"error"` and is therefore escalated; aborts collection on a malformed corpus | fix-now |
| phase8-test-pins-inverted-precedence-0 | code-review | ❌ FAIL | `test_the_always_entry_outranks_the_error_rule` pins the inverted model — passes while false, and blocks its own fix | fix-now |
| phase8-sr-05-002-nonblocking-premise-inverted-0 | code-review | 🎫 NEW-ISSUE | **Premise inversion** — SR-05-002's "non-blocking" MUST has never held; wrong rationale also recorded in project history | **inversion-halt** |
| phase8-unknownspecid-also-escalated-0 | code-review | 🎫 NEW-ISSUE | Same misordering pre-exists for `always::UnknownSpecIdWarning`; the marker gate hard-aborts instead of warning | fix-now-file |
| phase5-sr-05-002-says-pytestwarning-0 | spec-conformance | 🎫 NEW-ISSUE | SR-05-002 requires a `PytestWarning`; both classes are `UserWarning` (and `vultron/` imports pytest nowhere) | fix-now-file |
| phase9-broad-except-witness-test-scope-0 | notes-currency | 🎫 NEW-ISSUE | Third broad-except witness, first outside `vultron/`; CS-23-001 / #3340 / #3347 all stop at `vultron/`, leaving 18 `except Exception` sites in `test/` unenforced | fix-now-file |
| phase5-no-spec-entry-for-new-gate-0 | spec-conformance | ⚠️ IMPROVE | New gate behaviour has no SR-05 entry, though SR-05-001–005 specify this exact hook (BW-07-001/004) | fix-now |
| phase8-silent-branches-untested-0 | code-review | ⚠️ IMPROVE | Docstring advertises three early-return cases; neither silent branch (absent `specs/`, empty `registry.files`) is tested | fix-now |
| phase8-missing-spec-corpus-marker-0 | code-review | ⚠️ IMPROVE | `TestHealthyCorpusStillValidates` reads the real corpus without `@pytest.mark.spec_corpus`, so `spec-check.yml` skips it | fix-now |
| phase10-missing-spec-marker-0 | test-coverage | ⚠️ IMPROVE | No `@pytest.mark.spec("SR-05-002")` on the dedicated SR-05-002 test file (SR-05-004; peer convention in the same dir) | fix-now |
| phase3-verification-incomplete-0 | pr-body-format | ⚠️ IMPROVE | Verification omits the real total test count and the AC-1–AC-5 checkboxes | fix-now |
| phase8-pr-body-two-guards-claim-0 | code-review | ⚠️ IMPROVE | "the two guards above the `try`" — only one is above; `if not registry.files` is below at `test/conftest.py:154` | fix-now |
| phase8-change-detector-test-0 | code-review | ⚠️ IMPROVE | `test_caught_types_are_the_documented_load_failures` restates the implementation constant | fix-now |

**Total**: 2 FAIL · 7 IMPROVE · 4 NEW-ISSUE

---

### Verified sound (not findings)

- **Mutation claim is exact** — reverting only the `except` clause fails **7 of 17**, as the PR body states. A whole-file revert to `origin/main` fails 8 (the extra one trips on the missing `_REGISTRY_LOAD_ERRORS` constant).
- **All five ACs of #3331 map to the diff** — warning class + `__all__` export, `always::` entry (present, but see FAIL 1), narrowed `except`, corrected docstring, both-halves regression tests.
- **The narrowed tuple is not a regression risk** — pydantic `ValidationError` ⊂ `ValueError`; duplicate spec/group IDs raise `ValueError`; `UnicodeDecodeError` ⊂ `ValueError`; unreadable file / `IsADirectoryError` ⊂ `OSError`; `CSafeLoader` parse errors ⊂ `yaml.YAMLError`. The real malformed-spec case clears the `except` fine — it dies on FAIL 1 instead.
- `_run_gate(None, None, items)` is safe (the hook never touches `session`/`config`); `monkeypatch.setattr(root_conftest, "load_registry", …)` intercepts correctly; `stacklevel=2` attributes the warning to the `except` clause.
- Synthetic IDs are legitimately exempt from SR-04-008 — `_PHANTOM_ID_ALLOWLIST_DIRS = frozenset(["test/metadata/specs"])`.
- No ADR warranted; the change follows the existing `UnknownSpecIdWarning` precedent, which carries none.
- 17/17 pass at HEAD, ~0.5 s slowest test. Merge state clean.

---

*Triage artifact: `.claude/pr-3336-triage.json`*

sei-ahouseholder (commented):
## PR Execute: #3336 — 13/13 findings resolved

**CI**: ✅ 23 pass, 2 `notify-failure` skipped (1 iteration, no CI fixes needed)
**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, synced at `dc27f7f8e`
**Unit suite**: 8771 passed · 399 skipped · 19 xfailed · 5557 subtests · **0 failed**
**Also clean**: black, flake8, mypy, pyright, `spec-lint`, `markdownlint`

---

### Gate 2 (inversion) — raised and resolved

Triage found that the mechanism meant to keep the new warning non-blocking does
the reverse. The human's call: **file a `type:Bug`, not a Concern** (a Concern
implies planning-before-implementation for something already diagnosed), fix it
here, and reconcile any outstanding issue that trips over it. Filed **#3380**.

One correction to the triage framing: this was **not a novel inversion**. Open
issue **#2329** had already measured the correct rule and named the ineffective
`always::UnknownSpecIdWarning` entry as a latent bug. The finding lived only in an
issue body, so #3331 was written on the wrong premise and this PR implemented it
that way. That is why the durable write-up — not a third measurement — is the fix
that matters.

---

### Findings

| # | Outcome | What changed |
|---|---|---|
| phase8-filterwarnings-order-inverted-0 | ✅ fixed | `"error"` moved to the head of `filterwarnings`; both `always::` entries now take effect |
| phase8-test-pins-inverted-precedence-0 | ✅ fixed | Index assertion inverted; added a parametrized `pytester` sub-session fed the project's real filter list |
| phase8-sr-05-002-nonblocking-premise-inverted-0 | ✅ fixed | Premise corrected in all five places it was recorded; filed #3380 |
| phase8-unknownspecid-also-escalated-0 | ✅ fixed | Same reordering; SR-05-002's non-blocking guarantee now holds for the first time |
| phase5-sr-05-002-says-pytestwarning-0 | ✅ fixed | SR-05-002 now specifies `UserWarning`, not `PytestWarning` |
| phase5-no-spec-entry-for-new-gate-0 | ✅ fixed | Added **SR-05-006** (three early-return cases) and **SR-05-007** (filter ordering + mandatory behavioural verification) |
| phase8-silent-branches-untested-0 | ✅ fixed | `TestNoCorpusStaysSilent` covers absent-dir and empty-dir |
| phase8-missing-spec-corpus-marker-0 | ✅ fixed | `@pytest.mark.spec_corpus` added so `spec-check.yml` runs the corpus-reading tests |
| phase10-missing-spec-marker-0 | ✅ fixed | Class-level `@pytest.mark.spec` markers (SR-05-002/006/007) |
| phase9-filterwarnings-ordering-guidance-0 | ✅ fixed | Write-up in `notes/testing-pitfalls.md`, pointer in `test/AGENTS.md` |
| phase9-reconcile-2329-0 | ✅ fixed | Commented on #2329 — two findings now settled in-tree |
| phase9-broad-except-witness-test-scope-0 | ✅ fixed | Commented the `test/` scope datum on #3347 (no duplicate filed) |
| phase3-verification-incomplete-0 · phase8-pr-body-two-guards-claim-0 · phase8-change-detector-test-0 | ✅ fixed | PR body corrected (real counts, AC checkboxes, guard claim); change-detector assertion trimmed |

---

### Mutation-checked both ways

Neither fix is taken on trust:

- Revert `test/conftest.py` to the bare `except` → **7 of 21** fail.
- Re-introduce the `filterwarnings` misordering → **3** fail, including both
  `pytester` sub-session cases.

That second check is the one the first cut lacked, and is why the bug shipped
green: `pytest.warns` and `catch_warnings` both *replace* the ini filters, so no
in-session assertion can observe an escalation.

---

### Notes for the reviewer

- `test/AGENTS.md` first went to 214 lines and tripped its 200-line ratchet, whose
  message says *"Do not raise the ceiling."* Condensed to a Pitfall Index pointer
  per the file's own routing rule; now 197.
- Confirmed after merging `origin/main` that the newly-landed
  `test_no_broad_except_outside_bt_update.py` still scans only
  `vultron/core/behaviors/`, so the `test/` gap recorded on #3347 stands.
- #3324 was checked and does **not** rest on the inverted premise.

*Execute artifact: `.claude/pr-3336-execute.json`*
*Next step: `/pr-verify`.*

sei-ahouseholder (commented):
## PR Verify: #3336 — ✅ READY-TO-MERGE

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, live re-check, no conflict markers in tree
**CI status**: ✅ passing — 23 checks pass, 2 `notify-failure` correctly skipped
**Artifact integrity**: ✅ 15 triage findings ↔ 15 execute results, no drift
**Execute sync**: ✅ confirmed (`dc27f7f8e`)

---

### Per-finding verdicts

| # | Severity | Verdict | Evidence at HEAD |
|---|---|---|---|
| phase8-filterwarnings-order-inverted-0 | ❌ FAIL | ✅ CONFIRMED | Parsed `pyproject.toml`: `"error"` at index 0; both `always::` entries at 1 and 2, i.e. **after** it |
| phase8-test-pins-inverted-precedence-0 | ❌ FAIL | ✅ CONFIRMED | `test_spec_marker_gate.py:308` now asserts `> filters.index("error")`; `pytest_plugins = ["pytester"]` and `runpytest_subprocess` present at 59/344 |
| phase8-sr-05-002-nonblocking-premise-inverted-0 | 🎫 NEW-ISSUE | ✅ CONFIRMED | #3380 filed and open; premise corrected in pyproject comment, conftest + warning + test-module docstrings, and SR-05-007 |
| phase8-unknownspecid-also-escalated-0 | 🎫 NEW-ISSUE | ✅ CONFIRMED | Same reordering covers it; `unknown-spec-id` parametrization present in the behavioural test |
| phase5-sr-05-002-says-pytestwarning-0 | 🎫 NEW-ISSUE | ✅ CONFIRMED | SR-05-002 loads from YAML naming `UnknownSpecIdWarning` / `UserWarning`, no `PytestWarning` |
| phase5-no-spec-entry-for-new-gate-0 | ⚠️ IMPROVE | ✅ CONFIRMED | SR-05-006 and SR-05-007 both resolve in the registry; `spec-lint` exit 0 |
| phase8-silent-branches-untested-0 | ⚠️ IMPROVE | ✅ CONFIRMED | `TestNoCorpusStaysSilent` at :124 with `_redirect_spec_dir` at :105 |
| phase8-missing-spec-corpus-marker-0 | ⚠️ IMPROVE | ✅ CONFIRMED | `@pytest.mark.spec_corpus` at :252 |
| phase10-missing-spec-marker-0 | ⚠️ IMPROVE | ✅ CONFIRMED | 7 `pytest.mark.spec(...)` occurrences (was 1, and that one was a fixture stand-in) |
| phase9-filterwarnings-ordering-guidance-0 | ⚠️ IMPROVE | ✅ CONFIRMED | Write-up in `notes/testing-pitfalls.md`; pointer at `test/AGENTS.md:165`; file at 197 lines, inside its 200 ratchet |
| phase9-reconcile-2329-0 | ⚠️ IMPROVE | ✅ CONFIRMED | Comment landed on #2329 |
| phase9-broad-except-witness-test-scope-0 | 🎫 NEW-ISSUE | ✅ CONFIRMED | Comment landed on #3347; no duplicate issue filed |
| phase3-verification-incomplete-0 | ⚠️ IMPROVE | ✅ CONFIRMED | PR body opens with `- Closes #3331` / `- Closes #3380`; real counts and AC checkboxes present |
| phase8-pr-body-two-guards-claim-0 | ⚠️ IMPROVE | ✅ CONFIRMED | The "two guards above the `try`" phrasing no longer appears in the body |
| phase8-change-detector-test-0 | ⚠️ IMPROVE | ✅ CONFIRMED | Renamed to `test_the_catch_is_not_blanket`; no `set(caught)` assertion remains |

**Total**: 2 FAIL confirmed · 9 IMPROVE confirmed · 4 NEW-ISSUE confirmed · **0 unresolved**

---

### Why this verdict is trustworthy

The two FAIL fixes were checked at HEAD by *parsing* `pyproject.toml` and reading
the assertion at its line, not by trusting the commit diff. Both were also
mutation-checked during execute: re-introducing the misordering fails 3 tests
(including both `pytester` sub-session cases), and reverting the conftest fix fails
7 of 21. The suite is green at 8771 passed / 0 failed after merging `origin/main`.

Worth stating plainly, since it is the substance of this PR: the version of #3336
that arrived at triage would have made things **worse** — it converted a silent
gate-skip into a hard collection abort, and its own guard test asserted the
inverted model so nothing could catch it. That is fixed, and the invariant is now
pinned behaviourally rather than by position.

### One open item for a human (not a merge blocker)

**#3380 will be closed by this PR**, but it was added to Project #24 at the default
`Schedule=Someday`, which reads oddly for something already fixed. Worth setting to
`Now` or letting the merge close it.

---

*Artifacts cleaned up: `.claude/pr-3336-triage.json`, `.claude/pr-3336-execute.json`*

Issue #3400, opened 2026-09-18
CM-23-002 calls case_fully_closed the last ledger entry, but CM-23-003/CM-23-012 permit entries after it


sei-ahouseholder:
## Summary

`CM-23-002` step 3 requires the CASE_MANAGER to "emit a final `case_fully_closed`
CaseLedgerEntry **as the last ledger entry for the case**". Its rationale leans on
that: "making the CASE_MANAGER's own RM.CLOSED the penultimate step and
`case_fully_closed` the final entry gives all replicas an unambiguous terminal
anchor in the ledger chain."

That anchor is not achievable under the rest of CM-23:

- `CM-23-003` — a non-owner `Leave(VulnerabilityCase)` advances only that
  participant and **"the case MUST remain open for the remaining participants"**,
  committing its own `close_case` entry.
- `CM-23-012` — on owner Leave, **only** the Case Owner and the CASE_MANAGER
  advance to `RM.CLOSED`; "every other participant MUST retain its rung".
- `CM-23-004` — fan-out skips participants already at `RM.CLOSED`, which only
  matters because others are still open at that point.

So after `case_fully_closed` there may still be open participants, and each of
them may still `Leave`, producing a `close_case` entry *after* the entry the spec
calls last.

## Observed

Not hypothetical — this is the FV scenario's actual ledger. `_phase_case_closure`
(`vultron/demo/scenario/fv_demo.py`) closes the Vendor (the CASE_OWNER, per
CBT-01-003) first, then the Finder:

```text
17: close_case          actor=vendor   <- owner Leave
18: case_fully_closed   actor=vendor   <- "the last ledger entry"
19: close_case          actor=finder   <- lands after it
```

All eight other scenarios have the same owner-then-others shape.

## Why this is a landmine rather than a cosmetic wording issue

"`case_fully_closed` is the terminal anchor" is the kind of premise consumers get
built on. Anything that treats it as a chain terminator — a replica deciding it
has a complete log, a ledger consumer stopping there, a future invariant
asserting it is the highest `logIndex` — would be wrong on every current
scenario, and would look correct in review because the spec says so.

## Resolution options (not decided)

1. **Require owner Leave to be last.** Every non-owner must leave before the
   owner. Makes the clause true, but it is a new normative ordering constraint on
   all scenarios, and it is unclear what the CASE_MANAGER should do if a bystander
   never leaves.
2. **Refuse bystander Leave after full closure.** The case is closed, so a later
[...]

PR #3382, opened 2026-09-18
fix(demo-ci): refresh the per-invariant diagnostic map and ratchet it against the harness

sei-ahouseholder:
- Closes #3337

## Summary

Refreshes the Per-Invariant Diagnostic Map in `notes/demo-ci-diagnostics.md`
against `test/ci/invariants/universal_harness.py`, and adds a ratchet test so it
cannot silently drift again.

## Motivation

The map is the entry point an agent uses to triage a red `Invariant Harness`
job. It listed 14 invariants against a harness that injects 18, and routed most
of them to #789/#791/#937 — all closed. A triage doc that is confidently wrong
is worse than no doc: it sends the investigation to a closed issue instead of
the layer that actually broke. Nothing enforced the correspondence, so refreshing
the table without a guard would just reset the clock on the same rot.

## Changes

- **`notes/demo-ci-diagnostics.md`**: table now lists every invariant the factory
  injects, keyed by **test-function name** rather than by number — the `#` column
  skips 8 and runs out for the last three, so counting rows misleads. `Status`
  records the ratchet state read off the `xfail` markers (`active` /
  `⏳ xfail — #N`), not a pass/fail snapshot no test can hold. Only invariant 7
  and `per_actor_replica_divergence` are xfail, both citing #2505. Also states
  the rule generating the `Start at Layer` column, replaces the scenario lists
  that had gone stale at six with a pointer to `.github/demo-scenarios.json`, and
  fixes a `related_notes` entry pointing at the deleted `notes/demo-ci.md`.
- **`notes/demo-ci-diagnostics.md`** (two corrections beyond a straight refresh):
  invariant 8 is *scenario-local*, and there are a dozen-plus of them across eight
  harnesses rather than the one the old table implied; and
  `per_actor_replica_divergence` runs each replica **in isolation**
  (`{actor: entries}`), so despite the name it never compares replicas — the old
  grouping would have sent a triager hunting for cross-actor disagreement that
  cannot be there.
- **`test/ci/invariants/test_diagnostic_map_sync.py`** (new): asserts the table's
  rows are exactly the factory's invariants (both directions), that each row's
  xfail state matches its marker, that an xfail row cites the same issue the
  marker does, and that every harness in `.github/demo-scenarios.json` passes
  `narrative_path` — omitting it is legal and silently drops invariant 16, which
  both notes files warned about in prose that nothing enforced.
- **`test/ci/README-case-log-ratchet.md`**: retires its duplicate per-invariant
[...]

Issue #2626, opened 2026-08-26
Cross-authority ledger replication diverges: fcv-reject and fvcv-handoff demo scenarios fail under ADR-0073


sei-ahouseholder:
## Symptom

Two of the four PR-event demo scenarios fail on `bug/2238-datalayer-create-actor-scope-collision` at `52a6d900`, after the per-actor DataLayer isolation work (ADR-0073) landed. `fv` and `fcvcv` pass; `fcv-reject` and `fvcv-handoff` do not.

Run: https://github.com/CERTCC/Vultron/actions/runs/32912074989

### `fcv-reject` Demo Integration

```text
CHECK FAILED: Participant count stays at 3 (Vendor not added after rejection)
  — Timed out waiting for event_type='reject_invite_actor_to_case' to appear
    in ledger for case 'urn:uuid:ea28aa2e-…' — auto-close replication may not
    have completed
```

The rejection is emitted but never reaches the ledger the check reads.

### `fvcv-handoff` Demo Integration

Two checks fail:

```text
CHECK FAILED: Vendor2 replica matches authoritative Vendor1 state
  — Replica log tail hash 'e793f87a…' != authoritative log tail hash '4f94fa9d…'
    — hash-chain replication integrity failure

CHECK FAILED: Note delivered to watching container
  — Timed out waiting for note 'urn:uuid:a02f5458-…' to appear in case 'urn:uuid:0fcbb548-…'
```

### `fvcv-handoff` Invariant Harness

3 failed / 37 passed / 2 xfailed (the 2 xfails are pre-existing, #2505):

```text
test_invariant_2_cross_actor_hash_agreement
  Cross-actor hash mismatches at logIndex=16:
    case-actor: c7107115…   vendor2: c7107115…
    coordinator: 4f94fa9d…  finder: 4f94fa9d…  vendor: 4f94fa9d…

test_invariant_5_expected_event_types_present[close_case]
  Expected eventType 'close_case' not found in case-actor log.

test_invariant_15_cs_state_transitions_observed
  vfd_state == 'VFd' (fix_ready) never observed
  pxa_state starting with 'P' (published/public-aware) never observed
```

## Why this is one issue

[...]

PR #3407, opened 2026-09-18
fix(case-closure): record the CASE_MANAGER's own RM.CLOSED on the ledger

### LINKED ISSUE #2505: FV demo fixture: CaseActor never reaches RM.CLOSED — test_all_participants_rm_closed_at_scenario_end always fails
## Summary

The integration test `test/demo/test_fv_demo.py::TestCaseLedgerInvariants::test_all_participants_rm_closed_at_scenario_end` fails on `main` and is a pre-existing defect unrelated to any current branch work.

## Root Cause

The FV demo scenario (`completed_workflow` fixture) includes a CaseActor service.  The scenario advances the CaseActor's `CaseParticipant` to `RM.ACCEPTED` via `add_participant_status_to_participant` ledger entries committed by `_CommitNativeLedgerEntriesNode`.  However the fixture never sends a message that would cause the CaseActor to close the case (RM.CLOSED), so `latest_rm[case_actor_id] = "ACCEPTED"` at scenario end.

The assertion requires every participant tracked in the ledger to be at `RM.CLOSED`:
```
assert not participants_not_closed
```

## Evidence — fails on main

```
git stash -- vultron/ test/     # revert all branch changes to pure main
uv run pytest test/demo/test_fv_demo.py::TestCaseLedgerInvariants::test_all_participants_rm_closed_at_scenario_end --tb=short -m integration
# FAILED ... AssertionError: Participants not in RM=CLOSED at scenario end:
#   {'http://localhost:.../actors/case-actor-...': 'ACCEPTED'}
git stash pop
```

## Causality Check

The test was accidentally passing on older code because `_CommitNativeLedgerEntriesNode` received no `wire_render_port` and returned `FAILURE`, which left the CaseActor's ledger entries uncommitted — making it invisible to the assertion.  Issue #2287 (route all payload-snapshot rendering through WireRenderPort) correctly provides the port, entries are committed, and the pre-existing incompleteness of the fixture is exposed.

## Fix

Update the FV demo scenario to send the appropriate message(s) so the CaseActor closes the case (or transitions through RM.CLOSED) before the scenario ends, or explicitly exclude the CaseActor from the RM.CLOSED invariant check if CaseActors are not expected to participate in the FV RM lifecycle.

### LINKED ISSUE #3385: test_invariant_per_actor_replica_divergence xfail masks three unrelated replica-side invariants
## Summary

`test_invariant_per_actor_replica_divergence` in
`test/ci/invariants/universal_harness.py` carries a blanket
`@pytest.mark.xfail(strict=False, reason="pre-existing bug #2505: FV demo CaseActor never reaches RM.CLOSED")`.
But the check it wraps, `check_per_actor_replica_divergence`
(`test/ci/invariants/common.py`), aggregates **four independent** checks per
replica, not one:

1. `check_no_rm_state_oscillation`
2. `check_rm_closed_termination`  ← the only one #2505 is about
3. `check_participant_status_schema_completeness`
4. `check_cs_state_transitions_observed`

Because the whole test is expected to fail, a regression in 1, 3, or 4 produces
no signal for as long as #2505 stays open.

## Why this is not covered elsewhere

Invariants 6, 9 and 15 check the same three properties, but they run against
`auth_entries(replicas)`, which returns the `case-actor` replica
(`common.py:277-279`). `check_per_actor_replica_divergence` is the **only** check
that applies those properties to non-`case-actor` replicas — it runs each one in
isolation as `{actor: entries}` precisely so `auth_entries()` falls back to that
actor's own log (ISSUE-2411 Gap 1).

So the masked coverage is specifically **replica-side** RM-oscillation,
`ParticipantStatus` schema completeness, and CS-transition observation. A
participant replica could start emitting `ParticipantStatus` snapshots with no
`cvdRole` and every Invariant Harness job would stay green.

`strict=False` compounds it: an XPASS is also green, so the masking does not
announce itself even if #2505 is fixed and the test starts passing.

## Suggested resolution

Split the aggregate so each property carries its own marker, and scope the xfail
to the termination check that #2505 actually owns. The other three then become
active guards immediately, without waiting on #2505.

## Evidence

- `test/ci/invariants/common.py` — `check_per_actor_replica_divergence`, `auth_entries`
- `test/ci/invariants/universal_harness.py` — the marker
- Surfaced while refreshing the diagnostic map in #3337 / #3382, which had to
  document what this test actually checks and found that both its name and its
  marker describe it inaccurately.

### LINKED ISSUE #3417: CommitCaseActorRMClosedEntryNode commits without the CLP-09-001 role gate
## Summary

`CommitCaseActorRMClosedEntryNode` (`vultron/core/behaviors/case/nodes/leave/record.py`)
invokes the canonical commit operation via `create_commit_log_entry_tree` with no
role gate. CLP-09-001 requires every such call site to reach the commit through a
composition that verifies the active actor holds `CVDRole.CASE_MANAGER` for the
case at invocation time — and says so explicitly as "a role-based check, not an
identity comparison, **even where the CASE_MANAGER and the role holder happen to
coincide today**."

## Why `DeclineForeignLedgerCommitNode` does not cover it

`create_commit_log_entry_tree` opens with `DeclineForeignLedgerCommitNode`, which
reads as a guard but is a **store-consistency** check, not an authority one — its
own docstring says so (ARCH-24-005). On a container co-hosting the CaseActor with
another actor, as the demo's vendor container does,
`store_for_actor(..., require_same_authority=True)` resolves a store for the
co-hosted actor, so the guard reports "not foreign" and the commit proceeds —
minting a canonical `add_participant_status_to_participant` index in a
non-CASE_MANAGER replica's log.

## Reachability

Not reachable at the time of filing. `SvcLeaveCaseUseCase._build_tree` addresses
the `Leave` to `to=[case_manager_id]` only, and `rm_close_case_activity` adds no
`cc`, so no non-CASE_MANAGER actor receives a `Leave` and the owner-Leave arm
never executes for one. The node's authority therefore rested on **addressing**
rather than on the role check CLP-09-001 mandates — which is exactly the
accumulation pattern CLP-09-001's rationale describes ("only one of roughly ten
call sites guarded the commit").

## Also: the sibling commit has no gate either

`case_fully_closed_commit` in `create_close_case_received_tree` builds
`ReconstructChainTailNode` / `CreateLogEntryNode` / `PersistLogEntryNode` /
`FanOutLogEntryNode` inline with no role gate at all. It carries a comment
justifying its identity handling for CLP-07-003 (actor sourced from the trusted
inbound `Leave`), but that is a snapshot-attribution argument, not the
invocation-time role check CLP-09-001 asks for. This is pre-existing and is the
reason the owner-Leave arm as a whole depends on addressing.

## Acceptance criteria

- [x] AC-1: `CommitCaseActorRMClosedEntryNode` checks `CVDRole.CASE_MANAGER` on
  the receiving actor's own participant record before committing, and skips
  quietly (SUCCESS + debug) when absent.
- [x] AC-2: A test asserts a non-CASE_MANAGER receiver mints no canonical entry,
  with a control test that the CASE_MANAGER still does.
- [ ] AC-3: `case_fully_closed_commit` is brought under the same role gate.

AC-1 and AC-2 are done in the PR that closes this issue. AC-3 is left open
deliberately: it changes a commit that every one of the nine demo scenarios
depends on, and it is pre-existing rather than introduced by that work.

## Evidence

- `specs/case-ledger-processing.yaml` — CLP-09-001, CLP-09-002
- `specs/behavior-tree-integration.yaml` — BT-17-005, BT-17-006
- `vultron/core/behaviors/sync/commit_tree.py` — `create_commit_log_entry_tree`
- `vultron/core/behaviors/sync/nodes/ledger_authority.py` — `DeclineForeignLedgerCommitNode`
- `vultron/core/behaviors/case/receive_close_case_tree.py` — `case_fully_closed_commit`

Surfaced by PR triage on #3407.

### LINKED ISSUE #3418: Owner/non-owner Leave arms are discriminated by Selector fall-through, so an owner-arm effect failure silently becomes a non-owner close
## Summary

In `create_close_case_received_tree`
(`vultron/core/behaviors/case/receive_close_case_tree.py`), the owner and
non-owner Leave paths are discriminated by **Selector fall-through** rather than
by mutually exclusive guards:

```text
OwnerOrNonOwnerEffects (Selector, memory=False)
├── OwnerLeaveSeq (Sequence)
│   ├── CheckIsCaseOwnerNode          <- the intended discriminator
│   ├── AdvanceParticipantToRMClosedNode
│   ├── AdvanceCaseActorToRMClosedNode
│   ├── CommitCaseActorRMClosedEntryNode
│   └── CommitCaseFullyClosedBT
└── NonOwnerLeaveFallbackSeq (Sequence)
    └── AdvanceParticipantToRMClosedNode
```

The design intent is that `CheckIsCaseOwnerNode` fails for a non-owner and the
Selector falls through. But the Selector cannot distinguish *which* child failed.
A failure in any **later** node of `OwnerLeaveSeq` — after the owner and the
CaseActor have already been advanced — is therefore reinterpreted as "the sender
is not the Case Owner", the fallback arm runs, and the tree reports `SUCCESS`
with an empty `BTBridge.get_failure_reason()`.

`AdvanceParticipantToRMClosedNode` and `AdvanceCaseActorToRMClosedNode` each have
several `FAILURE` returns, so the masking is live today, not hypothetical.

## Consequence

A half-closed case that reports success: the owner and CASE_MANAGER are
`RM.CLOSED` in the local store, `case_fully_closed` is never committed, nothing
fans out, and nothing is logged. Every replica keeps the case open forever with
no diagnostic pointing at the closure path.

## Why this is filed separately

Found during PR triage of #3407, which added a node to `OwnerLeaveSeq` whose
failure would have hit exactly this path. That PR resolved its own contribution
by making the new node best-effort (it warns and returns SUCCESS rather than
failing), so the PR no longer widens the hazard. What remains is the pre-existing
structure, which is a larger change than that PR should carry: it alters how a
protocol-critical tree discriminates its two arms, and all nine demo scenarios
run through it.

## Resolution sketch (not decided)

Make the arms mutually exclusive instead of order-dependent — e.g. guard
`NonOwnerLeaveFallbackSeq` with an inverted `CheckIsCaseOwnerNode` so that an
owner-arm effect failure fails the Selector rather than being absorbed. Note the
edge case: `CheckIsCaseOwnerNode` can return FAILURE for reasons other than "not
the owner" (e.g. the case is unreadable), and a naive `Inverter` would convert
those into "is not the owner". A dedicated `CheckIsNotCaseOwnerNode` that fails
closed on a read error is likely better than a decorator.

## Evidence

- `vultron/core/behaviors/case/receive_close_case_tree.py` — `OwnerOrNonOwnerEffects`
- `vultron/core/behaviors/case/nodes/leave/advance.py` — the FAILURE returns in both advance nodes
- `vultron/core/behaviors/case/nodes/leave/record.py` — `_best_effort`, which documents why the new node sidesteps this
- specs: CM-23-002 (owner Leave sequence), CM-23-003 (non-owner Leave)

sei-ahouseholder:
- Closes #2505
- Closes #3385
- Closes #3417

## Summary

Fixes #2505 by recording the CASE_MANAGER's own `RM.CLOSED` transition as a canonical
ledger entry, and splits the aggregate per-actor replica invariant into four
separately-markable checks (#3385).

Originally branched off #3382 (its `test_diagnostic_map_sync.py` ratchet holds the table
this PR edits); #3382 merged during the work, so this is rebased onto `main`.

## Motivation

**#2505's stated root cause is wrong, and both fixes it proposes are wrong.** The FV
fixture *does* send an owner `Leave`, and CM-23-002 runs correctly:
`AdvanceCaseActorToRMClosedNode` advances the Case Actor's own participant to
`RM.CLOSED` and `case_fully_closed` is committed. The real defect is that the advance
is **store-local and was never recorded**:

- `close_case` names the *departing* actor in `payloadSnapshot.actor`, and the Case
  Actor never sends itself a `Leave`, so `ApplyCloseCaseFromLedgerNode` never fires
  for it.
- `case_fully_closed` is attributed to the **owner**, not the Case Actor, and has no
  event-condition or effect node at all.

So every replica — and every ledger-derived invariant check — read the CASE_MANAGER as
permanently `RM.ACCEPTED`. CM-23-005 forbids exactly that: *"Each transition MUST be
recorded as a `CaseLedgerEntry`."* The three bootstrap transitions
(`RECEIVED`/`VALID`/`ACCEPTED`) are recorded; the terminal one was not.

The issue's alternative — exclude the Case Actor from the check — is spec-barred by
CM-23-002 step 2, CM-23-005, CM-23-010 and CM-23-012, and ADR-0051 already reversed it
in `AllParticipantsRMClosedConditionNode`: *"the former `CVDRole.CASE_MANAGER` skip is
no longer needed."* Three sites still carried that retired skip, which is why no
scenario's M7 check ever noticed.

With the defect fixed nothing is masked, so the `xfail`s come off — which also makes
#3385's "scope the xfail to the termination check" resolution moot. What remains of
#3385 is the structural hazard it actually reports: one marker over four independent
properties.

## Changes

- **`vultron/core/behaviors/case/nodes/leave/record.py`** (new):
  `CommitCaseActorRMClosedEntryNode` commits the CASE_MANAGER's `RM.CLOSED` as an
  `add_participant_status_to_participant` entry and fans it out. Committed
  **synchronously** in the receive tree rather than emitted as a self-addressed
  `Add(ParticipantStatus)` through the CLP-10-001 loopback, because CM-23-002 makes this
  the *penultimate* step and `case_fully_closed` the final entry — an outbox background
  task could not honour that ordering. Reuses `create_commit_log_entry_tree` and
  `build_add_participant_status_snapshot`; replicas need no new routing, since
  `IsParticipantStatusEventNode` already handles this event type.
- **`vultron/core/behaviors/case/receive_close_case_tree.py`**: wires the node into the
  owner-Leave arm, between the advance and `case_fully_closed`.
- **`vultron/adapters/driving/fastapi/inbox_port_factories.py`**, **`inbox_handler.py`**:
  move `CLOSE_CASE` into its own `_CLOSE_CASE_SEMANTICS` set with a
  `_close_case_port_factory` adding a `WireRenderPort` to the existing sync+trigger pair.
  Needed to render the snapshot; a separate set keeps the disjoint guard special-case-free.
- **`vultron/core/use_cases/received/case/lifecycle.py`**: threads `wire_render_port` from
  `CloseCaseReceivedUseCase` into the `BTBridge`.
- **`vultron/core/predicates/participants.py`**, **`vultron/demo/helpers/milestones.py`**,
  **`vultron/demo/helpers/verification.py`**: drop the stale CASE_MANAGER exemption from
  `all_participants_rm_closed` and `verify_case_closed`. Not a scope excursion — these are
  the mask that hid #2505, and ADR-0051 had already retired the same skip elsewhere.
- **`test/ci/invariants/common.py`**: replace `check_per_actor_replica_divergence` with a
  shared `for_each_replica` driver plus four named checks (#3385). Violations are now
  prefixed with the replica that produced them *and* named by property.
- **`test/ci/invariants/universal_harness.py`**: inject the four as separate tests; remove
  both `#2505` `xfail`s.
- **`notes/demo-ci-diagnostics.md`**: four table rows replace one; invariant 7 becomes
  active; adds guidance on why a replica-side row differs from its numbered twin, and on
  keeping any future `xfail` as narrow as its defect.
- **`vultron/core/behaviors/case/nodes/leave/`**: `leave.py` split into `advance.py` /
  `record.py` / `decline.py` behind a re-exporting `__init__.py` — the new node pushed it
  past the 500-line BTND-07-004 cap. The `force_rm_state` quarantine entry is repointed,
  not grown: the same two sites, relocated.
- **`vultron/core/behaviors/case/nodes/leave/record.py`** (triage follow-up): recording is
  **best-effort** — `_best_effort()` warns and returns SUCCESS on every path where the entry
  cannot be produced. A FAILURE here aborted `OwnerLeaveSeq` before `case_fully_closed`, and
  the enclosing Selector then read the failed owner arm as *"the sender is not the Case
  Owner"* and succeeded down the non-owner path: a half-closed case reporting SUCCESS with an
  empty failure reason. Reachable without any wiring regression, because
  `create_commit_log_entry_tree` opens with `CheckLedgerFreshnessNode`, which fails by design
  on a gapped local prefix (SYNC-10-001/002) — so the first cut had made case closure
  conditional on ledger freshness. Logs through the module-level stdlib logger as well:
  `py_trees`' own `Behaviour.logger` writes to the console and never reaches the `logging`
  tree, and best-effort is only defensible if the miss is observable. Also **role-gated** on
  `CVDRole.CASE_MANAGER` — *also* an excursion, hence its own closing bullet: the commit was
  an unguarded canonical-commit call site, which CLP-09-001 forbids, and
  `DeclineForeignLedgerCommitNode` does not cover it (store-consistency, not authority,
  ARCH-24-005). Closes #3417.
- **`notes/case-ledger-authority.md`**: the write-up AGENTS.md's routing rule requires before
  a pitfall cell may be extended — "A Store-Local State Change Is Invisible to Every
  Replica". Carries the visibility contract, why neither adjacent entry can stand in for the
  CASE_MANAGER's own, why recording is best-effort, and why this entry is committed
  synchronously rather than through the CLP-10-001 loopback (rationale that previously lived
  only in a docstring).
- **`notes/demo-ci-invariants.md`**: repoint a "Regression coverage" entry that still named
  `TestCheckPerActorReplicaDivergence`, deleted by this PR.
- **`AGENTS.md`**: one clause on the case-ledger row — a state change written only to the
  writer's own store is invisible to every replica.
- **Tests**: 2 regression tests for the ledger entry and its ordering against
  `case_fully_closed`; 1 guard that `CLOSE_CASE`'s factory supplies all three ports; the
  `per_actor_replica_*` suite rewritten as four classes plus a `for_each_replica` class;
  four `test_participants.py` cases inverted (they asserted the retired skip); 20 call
  sites given the port the production path now supplies.

## Verification

- `uv run pytest -m ""`: **10106 passed, 447 skipped, 20 xfailed, 5557 subtests passed**, 0 failed
- Black, flake8, mypy, pyright clean; `markdownlint-cli2` clean
- `test_diagnostic_map_sync.py` (#3382's ratchet) in sync — no `xfail` rows remain
- Both original regression tests verified **non-vacuous**: each fails with the new node removed
- The two best-effort tests that matter verified non-vacuous too:
  `test_missing_port_still_commits_case_fully_closed` and
  `test_failed_recording_is_logged_to_the_stdlib_logger` both fail with the node restored to
  returning `FAILURE` — i.e. they fail against the first cut of this PR, which is the point
- `test_fv_demo.py::TestCaseLedgerInvariants::test_all_participants_rm_closed_at_scenario_end`
  passes with its `xfail` removed

**Verified in CI, not locally:** Docker is unavailable in this environment, so the
9-scenario Demo Integration / Invariant Harness matrix was not run locally. CI has since run
all nine green, including `fcv-reject` and `fvcv-handoff` — the two #2626 named as able to
fail the replica-side checks for reasons unrelated to #2505. That contingency did not
materialise, so no scenario-scoped marker was needed.

## Filed, not fixed

#3418 (Concern) — the owner and non-owner Leave arms in `create_close_case_received_tree`
are discriminated by Selector **fall-through**, not by mutually exclusive guards, so a failure
in any *later* node of the owner arm is reinterpreted as "the sender is not the Case Owner":
the fallback arm runs and the tree reports SUCCESS with an empty failure reason. Both advance
nodes can already fail, so it is live, not hypothetical. Making the new node best-effort
removed this PR's contribution to it; fixing the structure changes how a tree all nine
scenarios run through discriminates its arms, so it is filed rather than folded in.

#3400 — CM-23-002 step 3 calls `case_fully_closed` "the last ledger entry for the case",
but CM-23-003 and CM-23-012 permit a bystander to `Leave` afterwards, and in FV one does
(`close_case` at index 19 follows `case_fully_closed` at 18). Which side is wrong is a
protocol decision, so it is filed rather than settled here.


sei-ahouseholder (commented):
## PR Triage: #3407 — fix(case-closure): record the CASE_MANAGER's own RM.CLOSED on the ledger

**Linked issues**: #2505 (FV demo fixture: CaseActor never reaches RM.CLOSED), #3385 (`test_invariant_per_actor_replica_divergence` xfail masks three unrelated replica-side invariants)
**Changed files**: 27 files — `core/behaviors/case`, `core/use_cases/received`, `core/predicates`, `adapters/driving/fastapi`, `demo/helpers`, `test/ci/invariants`, `notes`, `docs`
**CI status**: ✅ passing (all 22 checks, including all nine Demo Integration + Invariant Harness scenario jobs)
**Base branch**: `main`
**Needs integration tests**: yes — satisfied by CI

---

### Findings

|---|---|---|---|---|
| phase8-node-failure-suppresses-case-closure-0 | code-review | ❌ FAIL | A FAILURE from `CommitCaseActorRMClosedEntryNode` silently drops `case_fully_closed` while the tree reports SUCCESS | fix-now |
| phase7-agents-md-cell-without-note-writeup-0 | agents-md-compliance | ❌ FAIL | `AGENTS.md` cell extended with the #2505 clause, but no write-up in `notes/case-ledger-authority.md` | fix-now |
| phase8-new-commit-not-role-gated-0 | code-review | ⚠️ IMPROVE | The new canonical commit has no CASE_MANAGER role gate, unlike the `close_case` receipt in the same tree | fix-now-file |
| phase10-node-failure-branches-untested-0 | test-coverage | ⚠️ IMPROVE | None of the new node's non-SUCCESS branches are tested | fix-now |
| phase9-stale-test-class-reference-0 | notes-currency | ⚠️ IMPROVE | `notes/demo-ci-invariants.md:154` names a test class this PR deleted | fix-now |
| phase3-stale-verification-caveat-0 | pr-body-format | ⚠️ IMPROVE | The "Not verified locally" caveat is stale — CI ran all nine scenarios green | fix-now |

**Total**: 2 FAIL · 4 IMPROVE · 0 NEW-ISSUE

---

### FAIL detail

**`phase8-node-failure-suppresses-case-closure-0`** — `vultron/core/behaviors/case/nodes/leave/record.py:215`

`CommitCaseActorRMClosedEntryNode` is a child of `OwnerLeaveSeq` (a `Sequence`, `memory=False`) placed *before* `case_fully_closed_commit`. When it returns `FAILURE` the Sequence aborts, and the enclosing `OwnerOrNonOwnerEffects` Selector falls through to `NonOwnerLeaveFallbackSeq`, which succeeds. So a failure of the new node is reinterpreted as "the sender is not the Case Owner":

```text
CloseCaseReceivedUseCase(dl, owner-Leave event, sync_port=..., wire_render_port=None).execute()
  -> ledger: ['close_case']            # case_fully_closed MISSING, no fan-out
  -> tree status: SUCCESS, failure reason: ''

with wire_render_port=As2WireRenderAdapter():
  -> ledger: ['close_case', 'add_participant_status_to_participant', 'case_fully_closed']
```

The owner and the CaseActor end up `RM.CLOSED` in the local store, the case is never recorded as fully closed, nothing is fanned out, and `BTBridge.get_failure_reason()` is empty — so `CloseCaseReceivedUseCase.execute()` logs nothing at all. The node's comment says *"Fail loudly rather than commit an empty payload"*; the enclosing structure makes it fail silently and take case closure with it.

This does **not** require a wiring regression to reach. `create_commit_log_entry_tree` begins with `CheckLedgerFreshnessNode`, which returns `FAILURE` on a gapped local prefix *by design* (SYNC-10-001/002) — so this PR newly couples case closure to ledger freshness, silently. Before it, a gapped CaseActor ledger did not block `case_fully_closed`.

Two candidate fixes: treat a non-SUCCESS commit as best-effort, following the `_commit_one` precedent at `case_proposal_received_tree.py:872-905` (which commits the same event type and reserves hard failure for the load-bearing genesis entry); or make `OwnerLeaveSeq`'s failure propagate rather than be absorbed by the Selector, so the use case at least logs it.

**`phase7-agents-md-cell-without-note-writeup-0`** — `AGENTS.md:266`

`AGENTS.md:251-253` requires the write-up first: *"New pitfalls MUST be routed per `notes/agents-md-structure.md`: write-up in the nearest `notes/` or per-directory `AGENTS.md`, then extend a cell below."* The cell gained the #2505 clause, but `notes/case-ledger-authority.md` mentions neither 2505 nor CM-23-005. Every prior clause in that same cell honours the rule (#2824 → `case-ledger-authority.md:391`; #1689 → sections at `:670` and `:688`). The `plan/incoming/learnings/` entry is an input queue for `learn`, not the durable write-up, and it carries a different lesson.

---

### Verified clean

- **Merge state / CI**: `MERGEABLE` (`CLEAN`); all nine scenario jobs, `Tests (pytest)`, four linters and `docs-build-check` green.
- **ADR-0021 identity contract**: holds — `create_commit_log_entry_tree` is gated by `DeclineForeignLedgerCommitNode`, so a replica declines rather than forking the chain. No ADR was warranted; the PR implements ADR-0050/0051 rather than contradicting them.
- **BTND-07-004**: the `leave/` split is compliant (253 / 231 / 68 / 56 lines, from 310 + 231 = 541). The three class bodies are byte-identical to `origin/main:leave.py`; all public names re-exported with `__all__`; the `force_rm_state` quarantine was repointed with its count unchanged, exactly as the body claims.
- **`CLOSE_CASE` dispatch**: no set-membership bug. Removed from `_SYNC_AND_TRIGGER_PORT_SEMANTICS`, present only in `_CLOSE_CASE_SEMANTICS`, included in the disjoint guard, nothing double-registered, nothing lost relative to the old factory. `make_dispatcher()` is the only dispatcher construction in the codebase (`get_dispatcher` has no non-test callers), so the missing-port path is unreachable through the shipped adapter.
- **Exemption removal**: complete and consistent across all three sites. No inverted boolean, no new vacuity — `all_participants_rm_closed([])` still returns `True` as documented. The demo gate order is right: `wait_for_all_participants_rm_closed` polls before `verify_case_closed`.
- **`common.py` refactor**: behaviour-preserving — same replicas iterated, same four checks, same gate semantics, `check_fix_ready` still threaded per replica, violations keep their `Actor '<id>':` prefix. No check became vacuous.
- **Injected harness tests**: all four `test_invariant_per_actor_replica_*` attributes are really collected, and `test_diagnostic_map_sync.py` passes against the four new rows.
- **Ordering**: verified empirically as `close_case`(0) → `add_participant_status_to_participant`(1) → `case_fully_closed`(2); replay is idempotent (3 runs, 3 entries, stable indices).

---

*Triage artifact: `.claude/pr-3407-triage.json`*

sei-ahouseholder (commented):
## PR Execute: #3407

**Findings processed**: 6 of 6 — 6 fixed, 0 deferred, 0 skipped, 0 halted
**Commit**: `9f7207d5d` (plus PR-body-only edits)
**Sync**: merged `origin/main` cleanly (`e8c70e275`) — no conflicts
**Local suite**: `uv run pytest -m ""` → **10106 passed, 447 skipped, 20 xfailed, 5557 subtests passed**, 0 failed
**Linters**: Black, flake8, mypy (730 files), pyright (0 errors), markdownlint — all clean
**CI**: ✅ 22/22 passing, including all nine Demo Integration + Invariant Harness scenario jobs
**Merge state**: ✅ MERGEABLE (CLEAN) against `main`

---

### Results

| Finding | Outcome | Where |
|---|---|---|
| `phase8-node-failure-suppresses-case-closure-0` (FAIL) | ✅ fixed | `9f7207d5d` |
| `phase7-agents-md-cell-without-note-writeup-0` (FAIL) | ✅ fixed | `9f7207d5d` |
| `phase8-new-commit-not-role-gated-0` (IMPROVE) | ✅ fixed | `9f7207d5d`, closes #3417 |
| `phase10-node-failure-branches-untested-0` (IMPROVE) | ✅ fixed | `9f7207d5d` |
| `phase9-stale-test-class-reference-0` (IMPROVE) | ✅ fixed | `9f7207d5d` |
| `phase3-stale-verification-caveat-0` (IMPROVE) | ✅ fixed | PR body |

---

### The FAIL, and the decision behind the fix

`CommitCaseActorRMClosedEntryNode` returning `FAILURE` aborted `OwnerLeaveSeq`
before `case_fully_closed`, and the enclosing `OwnerOrNonOwnerEffects` Selector
then read the failed owner arm as *"the sender is not the Case Owner"* and
succeeded down the non-owner path. The result was a half-closed case reporting
`SUCCESS` with an empty failure reason: no `case_fully_closed`, no fan-out,
nothing logged. Reachable without any wiring regression, because
`create_commit_log_entry_tree` opens with `CheckLedgerFreshnessNode`, which fails
*by design* on a gapped local prefix (SYNC-10-001/002) — so the first cut had
quietly made case closure conditional on ledger freshness.

CM-23-002 mandates the four-step sequence and CM-23-005 mandates that every
transition be recorded, but neither says which promise to break when the
recording fails. The maintainer chose **best-effort recording**: losing one
transition's visibility beats losing the case's terminal anchor.

So `_best_effort()` warns and returns SUCCESS on every path where the entry
cannot be produced, and steps 3 and 4 always run. This matches `_commit_one` in
`case_proposal_received_tree.py`, which treats the same event type as
best-effort and reserves hard failure for the load-bearing genesis entry.

One thing worth flagging, because it nearly undermined the whole choice: the node
originally logged via `self.logger`, and `py_trees.behaviour.Behaviour.logger` is
a `py_trees.logging.Logger` that writes straight to the console and **never
reaches the stdlib `logging` tree**. "Log loudly and continue" would have been
invisible to deployment logs and CI artifacts — which is precisely the silent
drop #2505 already was. `_best_effort()` now logs through the module-level stdlib
logger as well, and a test pins that.

### The role gate turned out to be a MUST, not hardening

Triage recorded the missing CASE_MANAGER gate as unreachable defence-in-depth.
Reading the spec upgraded it: **CLP-09-001** requires every canonical-commit call
site to reach the commit through a check that the active actor holds
`CVDRole.CASE_MANAGER` at invocation time — explicitly *"a role-based check, not
an identity comparison, even where the CASE_MANAGER and the role holder happen to
coincide today."* This node was an unguarded call site.
`DeclineForeignLedgerCommitNode` does not cover it: that is a store-consistency
check, not an authority one (ARCH-24-005), and on a container co-hosting the
CaseActor with another actor it reports "not foreign" and would let a
non-CASE_MANAGER mint a canonical index. Now gated, closing #3417.

### Filed

- **#3417** (Bug) — the CLP-09-001 gap above. Fixed here; PR closes it. AC-3
  (bringing the sibling `case_fully_closed_commit` under the same gate) is left
  open deliberately and marked so in the issue.
- **#3418** (Concern) — the pre-existing half of the Selector hazard: the owner
  and non-owner arms are discriminated by fall-through, not mutually exclusive
  guards, so *any* owner-arm effect failure is reinterpreted as a non-owner
  close. Both advance nodes can already fail, so it is live. Making the new node
  best-effort removed this PR's contribution; fixing the structure changes how a
  tree all nine scenarios run through discriminates its arms, so it is filed
  rather than folded in.

Both added to Project #24 (Schedule=Someday).

---

*Execute artifact: `.claude/pr-3407-execute.json`*
*Next step: `/pr-verify`.*

sei-ahouseholder (commented):
## PR Verify: #3407

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, not a draft, no conflict markers in the tree
**CI status**: ✅ passing — 22 checks, 20 pass / 2 skipped (`Notify CI failure`), 0 failing
**Integration tests**: ✅ all nine Demo Integration + Invariant Harness scenario jobs green
**Artifact integrity**: 6 triage findings ↔ 6 execute results, IDs matched; execute recorded a clean sync

---

### Per-finding verdicts

|---|---|---|---|
| `phase8-node-failure-suppresses-case-closure-0` | ❌ FAIL | fixed | ✅ CONFIRMED |
| `phase7-agents-md-cell-without-note-writeup-0` | ❌ FAIL | fixed | ✅ CONFIRMED |
| `phase8-new-commit-not-role-gated-0` | ⚠️ IMPROVE | fixed (closes #3417) | ✅ CONFIRMED |
| `phase10-node-failure-branches-untested-0` | ⚠️ IMPROVE | fixed | ✅ CONFIRMED |
| `phase9-stale-test-class-reference-0` | ⚠️ IMPROVE | fixed | ✅ CONFIRMED |
| `phase3-stale-verification-caveat-0` | ⚠️ IMPROVE | fixed | ✅ CONFIRMED |

**6 CONFIRMED · 0 UNRESOLVED · 0 MISSING-COMMIT · 0 NOTED**

---

### What was checked at HEAD, not just in the commit diff

- **Best-effort recording**: `_best_effort` present (7 references) and **zero**
  `return Status.FAILURE` remain in `record.py` — the recording paths all return
  SUCCESS now. The stdlib-logger call (`logger.warning("%s", self.feedback_message)`)
  is present at `record.py:155`, which is the line that makes "log loudly" true
  rather than console-only.
- **Notes write-up**: `## A Store-Local State Change Is Invisible to Every Replica
  (Issue #2505)` present at `notes/case-ledger-authority.md:713`, satisfying the
  AGENTS.md routing rule that the cell extension required.
- **Role gate**: the `CVDRole.CASE_MANAGER not in (participant.case_roles or [])`
  check is live at `record.py:209`.
- **Tests**: both new classes collect — 7 tests, present at HEAD.
- **Stale reference**: `TestCheckPerActorReplicaDivergence` no longer appears
  anywhere in `notes/demo-ci-invariants.md`; `TestForEachReplica` is at line 154.
- **PR body**: carries `Closes #3417`, the refreshed `10106 passed` count, and the
  corrected "Verified in CI, not locally" block.
- **Filed issues are real and open**: #3417 (Bug), #3418 (Concern).

---

### For the record

Two things this pipeline changed that are worth a reviewer's attention, because
neither was in the original PR's scope:

1. The PR as first pushed had made **case closure conditional on ledger
   freshness** — silently, via a Selector that reinterpreted the new node's
   failure as "the sender is not the Case Owner". That is now fixed by
   maintainer decision (best-effort recording), and #3418 tracks the pre-existing
   half of the structure.
2. The new canonical commit was an **unguarded CLP-09-001 call site**. Triage
   filed it as defence-in-depth; the spec says MUST. Gated, closing #3417.

*Artifacts cleaned up.*

Issue #3471, opened 2026-09-21
as:Question has no dispatch path: the CBT-03-004 bootstrap replay request is unroutable


sei-ahouseholder:
## Summary

`as:Question` has **no dispatch path in Vultron at all** — no `ActivityPattern`
in `vultron/wire/as2/extractor/_instances.py`, no `MessageSemantics` member, no
received-side use case. Verified by grep on `origin/main`: the token `Question`
appears in none of those three places.

That would be harmless if nothing sent one. Something does.
`bootstrap_replay_question_activity` (`vultron/wire/as2/factories/case.py:835`)
builds a `Question[context=VulnerabilityCase]` asking a peer to resend a missing
bootstrap `Create(VulnerabilityCase)`. It is emitted for real from
`vultron/adapters/driving/fastapi/inbox_pending_queue.py:156` when an actor's
pre-bootstrap queue expires: the activity is saved, appended to the outbox, and
logged at INFO as `Sent bootstrap replay Question`.

The recipient cannot route it. So:

- CBT-03-004 — a `SHOULD` — is unsatisfiable end to end, and was before any of
  the implementation existed.
- The sending actor logs success and waits for a reply that cannot come. There is
  no timeout, no fault, and no second attempt path keyed on the absence of a
  response.

This is the same shape as CP-06-004 in
`plan/incoming/learnings/20260918-3399-bt18-has-no-contract-for-a-refusal-carrying-a-payload.md`:
a spec obliges an exchange whose other half has no mechanism.

Found while planning #3433 (the `ChoosePreferredEmbargo` poll). That issue named
one undispatchable `Question`; this is the other one, and unlike the poll this
one is live. ADR-0100 retires the poll and relies on the fact that
`Question[context=VulnerabilityCase]` is occupied by *this* activity, but
deliberately does not decide what should happen to it.

A second defect in the same code path: CBT-03-004 says the `target` of the
`as_Question` SHOULD be the `case_actor_id` recorded on the
`VultronPendingCaseInbox` entry. The factory sets `to` and `context` and never
sets `target`. So spec and code disagree about where the addressee lives, which
also means any future pattern keyed on `target_` would not match what is emitted.

## The open question this issue must answer first

**Is `Question` even the right vocabulary for a replay request?**

Vultron uses ActivityStreams activities as *state-change notifications* — "here
is what I did" — not as commands or requests (`notes/activitystreams-semantics.md`).
A replay request is a request/response exchange, and modelling it as a
[...]

Issue #3470, opened 2026-09-21
Resolve open embargo proposals in earliest-expiration order; prune the pending-proposal index


sei-ahouseholder:
## Summary

Make the code agree with the protocol's documented rule for resolving several
open embargo proposals. `docs/topics/process_models/em/defaults.md` is normative
and says Participants SHOULD accept the earliest-expiring open proposal and
handle the rest as revisions; `working_with_others.md` states the same heuristic
("The Shortest Embargo Proposed Wins"). ADR-0100 records the decision and
EP-08 makes it testable.

The implementation does something else. Three findings, measured on
`origin/main`:

1. **Order is arrival, not expiration.** `find_embargo_proposal_id`
   (`vultron/core/use_cases/triggers/_helpers.py:59`) returns the *first
   recorded* proposal from `case.pending_embargo_proposal_index`. So after a
   counter-proposal, an `accept-embargo` call that omits `proposal_id` accepts
   the *superseded* terms. The current behavior is asserted as intended by
   `test/core/use_cases/received/test_embargo_proposal_index.py` ("Counter-proposal
   case: first pending proposal used when no proposal_id given"), so that
   assertion changes with this work.
2. **The index is never pruned.** `_record_embargo_proposal_index`
   (`vultron/core/use_cases/received/embargo.py:143`) is the only writer, and
   nothing anywhere in `vultron/` removes an entry. Decided proposals survive
   acceptance, revision and teardown, so the set being ordered is polluted with
   entries that are no longer open.
3. **`proposed_embargoes` is pruned only on teardown — not on rejection.**
   Corrected during triage of PR #3476, which had claimed it was pruned in both
   places. The sole pruner is `RemoveFromProposedEmbargoesNode`
   (`behaviors/embargo/nodes/teardown.py`), wired into
   `remove_embargo_from_case_tree` and nowhere else.
   `reject_proposed_embargo_bt` omits it and `reject_proposed.py` only *reads*
   the list, so a **rejected** proposal survives in both records. This widens the
   issue rather than narrowing it: EP-08-003 governs every record of open
   proposals, so both need pruning on every decision path.

Findings 2 and 3 are why this is one issue and not three: ordering an unpruned
set by expiration still selects decided proposals.

## Acceptance Criteria

- [ ] AC-1: `find_embargo_proposal_id` selects the open proposal whose
  `EmbargoEvent.end_time` is earliest, not the first recorded. It gains the
[...]

Issue #3469, opened 2026-09-21
Retire ChoosePreferredEmbargo; fix as_Question option-field typing


sei-ahouseholder:
## Summary

Retire the multi-candidate embargo poll. Per ADR-0100, Vultron resolves several
open embargo proposals by presenting them sequentially in earliest-expiration
order, which is what `docs/topics/process_models/em/defaults.md` already
requires normatively. A "here are four, pick one" poll is the construct that
heuristic replaces, and it cannot be made to work: `_EmAcceptEmbargoActivity`
requires an inline `Invite` as its `object_`, so a candidate `EmbargoEvent`
inside a `oneOf` has no answer form.

The activity also cannot parse its own output today. `as_Question` declares
`anyOf`/`oneOf` as a single value where AS2 defines a collection, and
`_ChoosePreferredEmbargoActivity` added `any_of`/`one_of` alongside them — four
fields sharing two wire aliases. Serialization resolves the collision
last-writer-wins; validation binds the inherited singular field. Verified on
`origin/main`: an emitted example is refused at `parse_activity`, not merely
extracted as `UNKNOWN`.

Nothing emits the poll, so no wire-visible behavior changes.

## Acceptance Criteria

- [ ] AC-1: `_ChoosePreferredEmbargoActivity` is removed from
  `vultron/wire/as2/vocab/activities/embargo.py`, and
  `choose_preferred_embargo_activity` from
  `vultron/wire/as2/factories/embargo.py` and the `factories/__init__.py`
  re-export list.
- [ ] AC-2: The `choose_preferred_embargo()` example is removed from
  `vultron/wire/as2/vocab/examples/embargo.py`, along with its entry in the
  generator in `vocab_examples.py` and the committed artifact
  `docs/reference/examples/choose_preferred_embargo.json`. The stale 2023
  artifact `doc/examples/037_choosepreferredembargo.json` is deleted too — it
  still advertises the retired wire type name `ChoosePreferredEmbargo`, which
  the class never actually serialized as.
- [ ] AC-3: Its coverage is removed from
  `test/wire/as2/factories/test_embargo_factories.py` and
  `test/wire/as2/vocab/test_vocab_embargo_examples.py`, and the
  `choose_preferred_embargo` entry is removed from `_KNOWN_UNDISPATCHABLE` in
  `test/architecture/test_vocab_examples_dispatchable.py`. That test's module
  docstring is updated so it no longer describes the exemption as open.
[...]

Issue #3392, opened 2026-09-18
Carry a proposed EmbargoEvent on the report Offer; widen EmbargoEvent.context; implement shortest-wins at case creation


sei-ahouseholder:
## Summary

A Reporter has no way to state embargo terms when submitting a Report. EP-04-004 recorded this
as the contingency blocking EP-04-003 (shortest-proposal-wins at case creation), so the
comparison has never been reachable. ADR-0096 decided the mechanism: embed a proposed
`EmbargoEvent` in the `Offer(VulnerabilityReport)` activity.

This requires widening `EmbargoEvent.context` from *always a Case* to **the thing the embargo is
about** — the Report before a Case exists, the Case afterwards, rewritten at case creation. The
widening is a semantic change, not a type change: `context` is a `NonEmptyString` and is only
ever *cast* to a case, never validated as one.

Two alternatives were rejected and should not be reintroduced: an activity-level `end_time` on
the `Offer` (adds a third meaning to a field CM-28-001 already calls a critical naming hazard),
and inlining the Reporter's own `EmbargoPolicy` (a standing preference, not terms for this
Report).

## Acceptance Criteria

- [ ] AC-1: `EmbargoEvent.context` accepts a Report URI as well as a Case URI, and readers no
      longer assume a Case without checking (EP-04-009). The two
      `cast("VultronCase | None", self.context)` sites in
      `vultron/core/models/events/embargo.py` are the known consumers.
- [ ] AC-2: `Offer(VulnerabilityReport)` can carry a proposed `EmbargoEvent`, and the wire
      pattern matches it. `CreateEmbargoEventPattern`'s `context_=VULNERABILITY_CASE` constraint
      widens to admit a report context (VAM-05-001).
- [ ] AC-3: Semantic extraction surfaces the proposed terms to core, so a use case need not
      re-read the stored wire activity (ADR-0035 / DL-06-002).
- [ ] AC-4: The CASE_MANAGER rewrites the proposal's `context` from the Report URI to the Case
      URI at case creation (EP-04-004).
- [ ] AC-5: EP-04-003's shortest-wins merge is implemented at case creation: the shorter of the
      sender proposal and the actor default becomes `EM.ACTIVE`; the longer is registered as
      `EM.REVISE`.
- [ ] AC-6: The merge treats the **actor default** as a candidate and the **protocol default**
      as not a candidate (EP-04-006). Depends on the naming split in the sibling Task.
[...]

PR #3476, opened 2026-09-21
docs: plan issue #3433 — retire the multi-candidate embargo poll

### LINKED ISSUE #3433: ChoosePreferredEmbargo is emit-only: no ActivityPattern, no MessageSemantics

### LINKED ISSUE #3513: A linter version bump can merge without running its own lint gate (pyright 1.1.414 broke Lint (pyright) on every open PR)
## Summary

The dependabot bump of pyright 1.1.411 → 1.1.414 (#3501) landed on `main` while
breaking the `Lint (pyright)` check for every other open PR. Four deliberate
negative tests construct a model with a required field omitted, inside
`pytest.raises(ValidationError)`, and 1.1.414 newly reports that omission as
`reportCallIssue`:

```text
test/core/models/test_base.py:107              - Argument missing for parameter "type_"
test/core/models/test_case_reference.py:41     - Argument missing for parameter "url"
test/wire/as2/vocab/test_case_reference.py:48  - Argument missing for parameter "url"
test/wire/as2/vocab/test_vulnerability_record.py:48 - Argument missing for parameter "name"
```

Fixed in PR #3476 (which the breakage was blocking) with
`# pyright: ignore[reportCallIssue]` on each call, matching the pre-existing
precedent at `test/metadata/specs/test_schema.py:778` — the identical
argumentless-construction-in-a-negative-test pattern, suppressed with the same
rule. The tests still pass and still assert what they claim (106 tests green
across the four files).

## The root cause is the gap that let it land

**#3501 ran no `Lint (pyright)` check at all.** The `Build` workflow that hosts
the four lint jobs did not trigger for that dependabot PR, so the bump merged
green while making the check fail on every branch that *does* run it. A
dependency bump to a linter is exactly the change whose own gate must run, and
this one was invisible to it.

That is the part worth fixing here. The four suppressions are already done;
what remains is making a linter-version bump self-gating.

## Acceptance Criteria

- [ ] AC-1: Determine why the `Build` workflow (which hosts `Lint (black)`,
  `Lint (flake8)`, `Lint (mypy)`, `Lint (pyright)`) did not run on #3501 —
  likely a `paths:` / `paths-ignore:` filter that excludes `uv.lock` and/or
  `pyproject.toml`, or a dependabot-actor condition.
- [ ] AC-2: A change to `uv.lock` or `pyproject.toml` triggers the full lint
  suite, so a tool bump cannot merge without its own gate. Asserted by
  inspecting the workflow triggers, not only by observing one green run.
- [ ] AC-3: Confirm the same blind spot does not apply to the other three lint
  jobs or to `Tests (pytest)` — if a bump to `black`, `mypy` or `pytest` would
  land ungated the same way, that is the same defect.
- [ ] AC-4: Record the outcome in `notes/ci-workflow-authoring.md`, whose
  existing pitfall "a red job may never have run its assertions (and an
  all-skipped run is green)" is the same family of failure: here the job did not
  run at all, and its absence read as success.

## Prior art

- `notes/ci-workflow-authoring.md` — "a red job may never have run its
  assertions (and an all-skipped run is green)"
- `test/metadata/specs/test_schema.py:778` — the suppression precedent applied
- ADR-0014 pins GitHub Actions to commit SHAs; this is the analogous
  "dependency changes must be gated" concern one layer up

## Reference

Found while resolving merge conflicts on PR #3476, after #3501 merged to `main`.

sei-ahouseholder:
- Closes #3433
- Closes #3513

## Summary

Decides that the multi-candidate embargo poll is out of scope. ADR-0100 retires
`ChoosePreferredEmbargo`, and EP-08 states the resolution rule it contradicted —
open embargo proposals resolve **earliest-expiration first** — as a testable spec
requirement rather than prose.

## Motivation

#3433 reported that `ChoosePreferredEmbargo` has a wire class and a factory but no
`ActivityPattern` and no `MessageSemantics`, and asked whether the poll is in
scope. Measuring it found three things the issue did not state, each of which
pushed the answer toward retirement:

1. **It does not extract as `UNKNOWN` — it is refused at `parse_activity`.**
   `as_Question` types `anyOf`/`oneOf` as a single value where AS2 defines a
   collection, and the subclass declared `any_of`/`one_of` *beside* the inherited
   fields — four fields, two wire aliases. Serialization resolved the collision
   last-writer-wins; validation bound the inherited singular field. The class
   could emit and never parse, including its own output.
2. **There was no answer form.** `_EmAcceptEmbargoActivity.object_` is a required,
   inline-typed `Invite`. A candidate `EmbargoEvent` inside a `oneOf` is not one,
   so adding a pattern and a semantic would have left the poll askable and still
   unanswerable.
3. **The protocol already answers this, sequentially.**
   `docs/topics/process_models/em/defaults.md` is normative and requires that a
   Participant accept the earliest-expiring open proposal and handle the rest as
   revisions. A "here are four, pick one" poll is what that heuristic replaces.

A fourth consideration is a cost rather than a blocker, and PR triage corrected an
earlier overstatement of it. A poll pattern would be the **first** `as_Question`
pattern in the registry — `_instances.py` has none — arriving alongside the one
CBT-03-004's bootstrap-replay request still needs (#3471). Those two are
separable: SE-08-001 *permits* patterns sharing a verb given an extra
discriminator and registry ordering, `ActivityPattern` has `target_`, and
CBT-03-004 requires the replay request's `target` be the `case_actor_id`. So
building the poll would mean designing two `Question` semantics and their registry
ordering to gain nothing the sequential `Invite` path already gives — not, as
first drafted, an impossibility.

## Changes

- **`docs/adr/0100-no-multi-candidate-embargo-poll.md`**: the decision, the
  rejected build-it option, and the measured facts above. Numbered 0100 rather
  than 0099 because PR #3468 holds 0099.
- **`specs/embargo-policy.yaml`**: new **EP-08 "Open Proposal Resolution Order"**
  (version 1.1.0 → 1.2.0). EP-08-001 restates the normative heuristic, EP-08-002
  forbids selecting by arrival order when no proposal is named, EP-08-003 requires
  decided proposals to leave the open-proposal record. The rule previously existed
  only in `docs/topics/`, which is why nothing enforced it and the code drifted.
  EP-04-003 is wired as `refines EP-08-001` — it is the two-party case-creation
  instance of the general rule — and EP-08-003's note records that **neither**
  record of open proposals is fully pruned: `pending_embargo_proposal_index` has no
  remover at all, and `proposed_embargoes` is pruned only on teardown, not on
  rejection.
- **`test/core/use_cases/triggers/embargo/test_proposal_resolution_order.py`**:
  strict-`xfail` coverage markers for all three EP-08 IDs. Asserted against the
  use case rather than the private helper, so they survive the signature change
  #3470 needs. Each drives the path EP-08 actually governs — a default selection
  with no `proposal_id`, and an *owner* accept that moves EM PROPOSED → ACTIVE —
  and the EM transition is asserted before the record check so the second test
  cannot pass or fail vacuously. Keeps the protocol-coverage ratchet flat at 933
  uncovered rather than spending 1 of its 4 remaining headroom.
- **`notes/embargo-lifecycle.md`**: new section on earliest-expiration resolution,
  the unpruned open-proposal records, and the rule that #3392 must share one
  comparator.
- **`notes/activitystreams-semantics.md`**: replaces the "ChoosePreferredEmbargo Is
  a Corner Case" section with the decision, and records the general trap — a
  subclass field whose wire alias collides with an inherited one is write-only.
- **`notes/protocol-asks.md`**: "`Question` for *which one*" loses its only
  instance; records that the surviving `as_Question` is not a precedent to copy.
  Frontmatter gains `specs/embargo-policy.yaml` and the two notes it now
  references, so those cross-links are two-way.
- **`notes/message-type-reference.md`**: records that the dispatchable-example
  gate's reach stops at the example corpus, which is how a live undispatchable
  activity went unnoticed, and that `_KNOWN_UNDISPATCHABLE` becomes empty once
  #3469 lands. Includes the near-miss diagnosis: the replay factory's three
  required arguments look like the cause, but the collector never walks
  `factories/case.py` at all, so relaxing that rule would not reach it.
- **`notes/embargo-default-semantics.md`**: the note that owns shortest-wins gains
  a pointer to EP-08 as its N-proposal generalization, plus the rule that #3392 and
  #3470 must share one earliest-end-date comparator.
- **`docs/howto/activitypub/activities/establish_embargo.md`**: the warning now
  says the activity is being retired rather than promising the gap will close, and
  the "one set of terms at a time" advice is restated as the protocol's resolution
  order rather than a temporary workaround. The `as:Question` subgraph and its two
  flowchart edges are removed — the diagram had still routed readers through the
  poll to an `Accept?` decision that cannot happen (#3469 AC-5's docs half).
- **`docs/reference/messages/em.md`**: drops the claim that "the answer arrives as
  an `EA` or `ER` against whichever candidate is proposed". It cannot: `EA`/`ER`
  take the `Invite` being answered as their object, and a candidate inside a
  `oneOf` is not one.
- **`notes/activity-factories.md`**, **`docs/adr/index.md`**, **`mkdocs.yml`**:
  inventory annotation and ADR wiring.
- **Four test files under `test/core/models/` and `test/wire/as2/vocab/`**
  (#3513): `# pyright: ignore[reportCallIssue]` on four deliberate negative
  tests. Unrelated to this PR's subject — pyright 1.1.414 landed on `main` via
  #3501 and broke `Lint (pyright)` on every open PR, including this one. Fixed
  rather than filed-and-left because it blocks this PR's CI and the fix is four
  lines with an established precedent (`test/metadata/specs/test_schema.py:778`).
  #3513 covers the actual gap: #3501 ran no pyright check, so a linter bump can
  merge without its own gate.

## Verification

- Unit suite: **9085 passed**, 442 skipped, 22 xfailed (2 new, both failing on
  their intended assertions — verified with `--runxfail`). Re-run after merging
  `main`, which brought in 84 tests of its own.
- Integration suite: **1284 passed**, 1 xfailed
- Black, flake8, mypy, pyright clean; `markdownlint-cli2` clean; `spec-lint`
  exit 0; `mkdocs build --strict` exit 0; all pre-commit hooks pass including the
  ADR index/nav sync check and the notes-frontmatter validator

## Implementation Issues

- #3469 — Retire ChoosePreferredEmbargo; fix `as_Question` option-field typing
- #3470 — Resolve open embargo proposals in earliest-expiration order; prune the
  open-proposal records

## Filed separately

- #3471 — **`as:Question` has no dispatch path at all**, and CBT-03-004's
  bootstrap-replay request is emitted live from `inbox_pending_queue.py` and
  cannot be routed by any recipient. Unlike the poll, that one is really sent.
  Whether `Question` is even the right vocabulary for a request/response exchange
  is part of that issue; this ADR relies only on the fact that a second `Question`
  semantic would have to be registered and ordered alongside it.

## Review corrections

PR triage found three factual errors in this PR's own reasoning, plus seven smaller
gaps; all ten are fixed in `11e5300`. **The decision is unchanged** — it rests on
the answer-form argument and the normative sequential-resolution rule, both of
which the review confirmed against source. What changed:

- The `Question`-pattern collision argument was overstated. No `Question` pattern
  is registered at all, `ActivityPattern.target_` exists, and SE-08-001 blesses
  exactly that kind of discrimination (SE-08-004 says how). Re-scored from a
  blocker to a cost.
- `proposed_embargoes` is **not** pruned on rejection, contrary to what EP-08-003's
  note, `notes/embargo-lifecycle.md` and the history entry all claimed. That widens
  #3470 rather than narrowing it; the issue gains AC-3b.
- The EP-08-003 `xfail` was asserting against an already-decided proposal, so a
  correct #3470 fix would not have promoted it. Rebuilt to drive an owner accept.

Triage detail: <https://github.com/CERTCC/Vultron/pull/3476#pullrequestreview-5270882328>


sei-ahouseholder (commented):
## PR Triage: #3476 — docs: plan issue #3433 — retire the multi-candidate embargo poll

**Linked issues**: #3433 (ChoosePreferredEmbargo is emit-only: no ActivityPattern, no MessageSemantics)
**Follow-ups filed by this PR**: #3469, #3470 (implementation) · #3471 (filed separately)
**Changed files**: 13 files — `specs/embargo-policy`, `docs/adr`, `notes/`, `core/use_cases/triggers/embargo`, `wire/as2`
**CI status**: ✅ passing (15/15 checks)
**Base branch**: `main`

---

### Findings

|---|---|---|---|---|
| phase5-proposed-embargoes-prune-claim-false-0 | spec-conformance | ❌ FAIL | EP-08-003's `note` says `proposed_embargoes` is "already pruned … on rejection" — it is not; only teardown prunes | fix-now |
| phase6-adr0100-question-shape-not-occupied-0 | adr-check | ❌ FAIL | ADR-0100 asserts `Question[context=VulnerabilityCase]` "is already occupied" — no `as_Question` pattern is registered at all, which the ADR itself says 110 lines later | fix-now |
| phase6-adr0100-activitypattern-discriminator-wrong-0 | adr-check | ❌ FAIL | "Distinguishing them needs a field `ActivityPattern` does not have" — `target_` exists, and CBT-03-004 mandates that very field | fix-now |
| phase6-se08001-mischaracterized-0 | adr-check | ⚠️ IMPROVE | SE-08-001 is cited as forbidding two patterns per activity; it permits them with a discriminator | fix-now |
| phase9-establish-embargo-flowchart-contradicts-admonition-0 | notes-docs-currency | ⚠️ IMPROVE | How-to flowchart still routes `ChoosePreferredEmbargo --> Accept?`, contradicting the "Do not use it" admonition below it | fix-now |
| phase10-ep08003-xfail-may-not-promote-0 | test-coverage | ⚠️ IMPROVE | The EP-08-003 strict xfail asserts against an already-decided proposal, so a correct #3470 fix may not flip it | fix-now |
| phase9-message-type-reference-wrong-gate-reason-0 | notes-docs-currency | ⚠️ IMPROVE | The note blames the gate's blind spot on the required-argument rule; the real cause is corpus scope | fix-now |
| phase5-ep08001-refines-direction-inverted-0 | spec-conformance | ⚠️ IMPROVE | `EP-08-001 refines EP-04-003` runs broad→narrow, against 879 corpus usages | fix-now |
| phase9-embargo-default-semantics-not-updated-0 | notes-docs-currency | ⚠️ IMPROVE | `notes/embargo-default-semantics.md` owns shortest-wins and gained no pointer to EP-08 / ADR-0100 | fix-now |
| phase9-protocol-asks-frontmatter-0 | notes-docs-currency | ⚠️ IMPROVE | `notes/protocol-asks.md` body cites EP-08/ADR-0100 but its frontmatter was not updated, leaving a one-way cross-link | fix-now |

**Total**: 3 FAIL · 7 IMPROVE · 0 NEW-ISSUE

---

### The three FAILs are one theme

All three sit in the *rejection rationale*, not the decision. ADR-0100 argues the poll
cannot be built partly because a `Question` pattern would collide with an occupied
shape that SE-08-001 forbids, and because no discriminator field exists. Measured
against the code: `_instances.py` registers **zero** `as_Question` patterns, so nothing
is occupied; `ActivityPattern.target_` exists and `match()` reads `activity.target`;
CBT-03-004 explicitly *requires* the replay Question's `target` be the `case_actor_id`;
and SE-08-001 names `target_` as an acceptable discriminator while SE-08-004 specifies
how to use it safely. The ADR is refuted by the spec it cites.

**This is not an inversion.** The decision rests on two independent legs that the review
confirmed hold: `EA`/`ER` require an inline `Invite`, so a `oneOf` candidate has no
answer form; and `defaults.md` normatively specifies sequential earliest-expiration
resolution. Retirement is still the right call — the reasoning needs correcting, not
reversing.

The third FAIL is separate and has a code consequence: `proposed_embargoes` is pruned
**only** by `RemoveFromProposedEmbargoesNode` in the teardown tree.
`reject_proposed_embargo_bt` does not include it. So on rejection *both* records leak,
not just `pending_embargo_proposal_index` — which makes #3470's real scope wider than
EP-08-003 currently describes.

---

### What was verified clean

The measured parts of this PR hold up well:

- **Ratchet arithmetic is exact** — 933 uncovered protocol specs against the 937 ceiling; the PR body's "flat at 933, 4 remaining headroom" is correct to the unit.
- **Both new strict xfails fail on their intended assertions**, confirmed with `--runxfail` — not incidentally, which is the usual failure mode for this pattern.
- **Unit suite reproduced locally**: 9001 passed, 442 skipped, 22 xfailed — matches the PR body exactly.
- **The hard wire-layer claims all check out**: the four-fields/two-aliases collision reproduces, `parse_activity` genuinely refuses the example's own output, `find_in_vocabulary("Question")` resolves to the base class, `_EmAcceptEmbargoActivity.object_` is a required inline `Invite`.
- **`find_embargo_proposal_id` does select by arrival order**, and `pending_embargo_proposal_index` has exactly one writer and no remover.
- Both `defaults.md` block quotes are verbatim; ADR-0100's number is correct (0099 is reserved by open PR #3468); `spec-lint` exit 0 with no new warnings; `mkdocs build --strict` clean; markdownlint clean; all five modified notes carry valid frontmatter; #3469/#3470/#3471 are wired to Epic parents and on Project #24.

---

*Triage artifact: `.claude/pr-3476-triage.json`*

sei-ahouseholder (commented):
## PR Execute: #3476 — 10/10 findings fixed

**Fix commit**: `11e5300` · **Sync commit**: `038733` (merged `origin/main` cleanly)
**CI**: ✅ 15/15 passing · **Merge state**: ✅ MERGEABLE (CLEAN) on `main`
**CI iterations**: 1 of 4 · **Gates**: neither triggered

---

### Findings

| # | Severity | Outcome | What changed |
|---|---|---|---|
| phase5-proposed-embargoes-prune-claim-false-0 | ❌ FAIL | ✅ fixed | Spec note, `embargo-lifecycle.md` and the history entry now say **neither** open-proposal record is fully pruned; #3470 gains AC-3b |
| phase6-adr0100-question-shape-not-occupied-0 | ❌ FAIL | ✅ fixed | The `Question`-shape claim is now prospective, not present-tense; ADR no longer contradicts itself |
| phase6-adr0100-activitypattern-discriminator-wrong-0 | ❌ FAIL | ✅ fixed | Re-scored Bad → Neutral; `target_` exists and CBT-03-004 mandates it |
| phase6-se08001-mischaracterized-0 | ⚠️ IMPROVE | ✅ fixed | SE-08-001 now described as permitting discriminated patterns, citing SE-08-004 |
| phase9-establish-embargo-flowchart-contradicts-admonition-0 | ⚠️ IMPROVE | ✅ fixed | `as:Question` subgraph and both edges removed from the mermaid diagram |
| phase10-ep08003-xfail-may-not-promote-0 | ⚠️ IMPROVE | ✅ fixed | Rebuilt at `EM.PROPOSED` with an **owner** accept; EM-transition asserted first |
| phase9-message-type-reference-wrong-gate-reason-0 | ⚠️ IMPROVE | ✅ fixed | Correct cause (corpus scope) given; argument-count framing recorded as a near-miss |
| phase5-ep08001-refines-direction-inverted-0 | ⚠️ IMPROVE | ✅ fixed | `EP-04-003 refines EP-08-001`; EP-08-001 keeps a `constrains` link with a note |
| phase9-embargo-default-semantics-not-updated-0 | ⚠️ IMPROVE | ✅ fixed | Section added tying EP-04-003 to EP-08-001 plus the one-comparator rule |
| phase9-protocol-asks-frontmatter-0 | ⚠️ IMPROVE | ✅ fixed | Frontmatter updated; every cross-reference this PR adds is now reciprocal |

**Fixed**: 10 · **Deferred**: 0 · **Skipped**: 0

---

### Why this was not an inversion-halt

The three FAILs all sit in ADR-0100's *rejection reasoning*, not its decision. The
decision rests on two legs that were verified against source and stand
untouched: `EA`/`ER` require an inline `Invite`, so a `oneOf` candidate has no
answer form; and `docs/topics/process_models/em/defaults.md` normatively specifies
sequential earliest-expiration resolution. The third leg — "you cannot register a
pattern for the poll" — was wrong in three linked ways (nothing occupies the
`Question` shape, `ActivityPattern.target_` exists, and SE-08-001 blesses exactly
that discrimination while SE-08-004 specifies how). It is rewritten as a *cost*
rather than a blocker. No premise the issue or its backing specs rest on was
overturned, so Gate 2 does not apply.

### The one finding with teeth beyond the docs

`proposed_embargoes` is pruned **only** on teardown — `reject_proposed_embargo_bt`
omits `RemoveFromProposedEmbargoesNode` and `reject_proposed.py` only reads the
list. Three files in this PR claimed it was pruned on rejection too. That
correction *widens* #3470: EP-08-003 governs every record of open proposals, so
both need pruning on every decision path. #3470 gained AC-3b accordingly.

### Verification

- Unit suite after the merge: **9085 passed**, 442 skipped, 22 xfailed, 5557 subtests
- Both EP-08 xfails confirmed failing on their **intended** assertions via `--runxfail`, before and after the rewrite. The rebuilt EP-08-003 test now fails at the record assertion *after* its `EM == ACTIVE` guard passes — so a real PROPOSED→ACTIVE decision occurs and the test cannot pass or fail vacuously
- black, flake8, mypy (746 files), pyright all clean; markdownlint clean across 2387 files
- `spec-lint` exit 0 with no new warnings; `mkdocs build --strict` exit 0
- Protocol-coverage ratchet still 933 against the 937 ceiling
- Integration suite not run — `needs_integration_tests` is false; no `vultron/` source, adapter, behavior or demo files are touched

### Notes for the record

- The PR body was rewritten: its Motivation had asserted the claim the ADR no longer makes, and its Changes section omitted three changed files. Mid-edit a `gh` call failed with an HTTP/2 error and my script blanked the body; it was reconstructed in full from the triage capture, with `- Closes #3433` restored on line 1.
- `git push` initially had no upstream ref. I confirmed `021eef44d` was an ancestor of local `HEAD` before pushing — a plain fast-forward, no force-push.
- One suite run reported a spurious timeout while pyright ran concurrently; its stack sat in pytest's own report hook. Clean when re-run alone. Flagged because `tail` had reported exit 0 on the killed run.

---

*Execute artifact: `.claude/pr-3476-execute.json`*
*Next step: run `/pr-verify` to spot-check each fix against HEAD.*

sei-ahouseholder (commented):
## PR Verify: #3476 — ✅ READY-TO-MERGE

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, not a draft, up to date with `main`
**CI**: ✅ 14 passing, 1 skipped (`Notify CI failure`, expected), 0 failing, 0 pending
**Conflict markers**: none (`git grep` clean)
**Closing reference**: `Closes #3433` present on line 1 and parsed by GitHub
**Findings**: 10/10 CONFIRMED at HEAD · 0 unresolved · 0 deferred · 0 halted

---

### Per-finding verdicts

Each was checked against **HEAD**, not just the commit diff.

| # | Severity | Verdict | Evidence at HEAD |
|---|---|---|---|
| phase5-proposed-embargoes-prune-claim-false-0 | ❌ FAIL | ✅ CONFIRMED | All three copies rewritten. EP-08-003's note now reads "neither is fully pruned"; `embargo-lifecycle.md` and the history entry name teardown as the only pruner. #3470 carries AC-3b |
| phase6-adr0100-question-shape-not-occupied-0 | ❌ FAIL | ✅ CONFIRMED | "Nothing registers a `Question` pattern today … a claim on a shape rather than a present collision". Grep for `already occupied` / `is occupied` in the ADR: 0 hits |
| phase6-adr0100-activitypattern-discriminator-wrong-0 | ❌ FAIL | ✅ CONFIRMED | Bullet re-scored Bad → **Neutral**, citing `target_`, CBT-03-004's `target` requirement and SE-08-004's `strict=True`. Grep for "does not have": 0 hits |
| phase6-se08001-mischaracterized-0 | ⚠️ IMPROVE | ✅ CONFIRMED | Note now says SE-08-001 *permits* patterns sharing a verb; "SE-08-001 forbids" is gone from both files |
| phase9-establish-embargo-flowchart-contradicts-admonition-0 | ⚠️ IMPROVE | ✅ CONFIRMED | Mermaid block contains 0 references to `as:Question` / `ChoosePreferredEmbargo`; the retirement admonition is intact |
| phase10-ep08003-xfail-may-not-promote-0 | ⚠️ IMPROVE | ✅ CONFIRMED | Uses `owner_actor_and_dl` + `_build_case_with_one_open_proposal` at `EM.PROPOSED`, with `EM == ACTIVE` asserted before the record check |
| phase9-message-type-reference-wrong-gate-reason-0 | ⚠️ IMPROVE | ✅ CONFIRMED | "stops at the example corpus" present; "stops at zero-required-argument" gone |
| phase5-ep08001-refines-direction-inverted-0 | ⚠️ IMPROVE | ✅ CONFIRMED | Registry load confirms `EP-04-003 → refines EP-08-001` and `EP-08-001 → constrains EP-04-003` (with note) |
| phase9-embargo-default-semantics-not-updated-0 | ⚠️ IMPROVE | ✅ CONFIRMED | "EP-04-003 is the two-party instance of a general rule" section present |
| phase9-protocol-asks-frontmatter-0 | ⚠️ IMPROVE | ✅ CONFIRMED | All three frontmatter additions present; the reciprocal link was added to `embargo-lifecycle.md` too |

---

### Independent re-derivation of the spec relationships

Rather than trust the diff text, the relationship graph was reloaded through
`load_registry` and read back:

```text
EP-08-001 kind=protocol priority=SHOULD rels=[('constrains', 'EP-04-003', note=True)]
EP-04-003 kind=protocol priority=MUST   rels=[('refines',    'EP-08-001', note=False)]
EP-08-002 kind=protocol priority=MUST   rels=[('refines',    'EP-08-001', note=False)]
EP-08-003 kind=project  priority=MUST   rels=[('constrains', 'EP-08-002', note=False)]
```

`refines` now runs narrow → broad throughout, matching all 879 corpus usages.

### One artifact nit, non-blocking

The execute artifact records the sync confirmation as `merge_state.synced_with_base`
rather than the `merge_state.synced` key this skill's Phase 1 looks for. The
condition itself is recorded and true, and Phase 2's live check is the
authoritative one, so this does not raise `UNSYNCED-EXECUTE` — noting it so the
key name gets aligned rather than silently mapped.

### Not re-run here

The integration suite was not run, and correctly so: `needs_integration_tests` is
false because no `vultron/` source, adapter, behavior or demo file is touched. The
PR body's integration figure (1284 passed) is carried over from before the `main`
merge and was not re-derived in this pass.

---

*Artifacts cleaned up.*

Issue #1991, opened 2026-08-05
Remove alias_generator=to_camel from core-layer classes (ARCH-12-004)


sei-ahouseholder:
## Summary

`test/architecture/test_hierarchy_invariants.py::TestCoreVocabularyHierarchy::test_no_core_object_has_to_camel_alias_generator` is `xfail` referencing #800 (closed without fixing this violation).

The following core-layer classes still carry `alias_generator=to_camel` inherited from `as_Base` (wire layer), violating ARCH-12-004:

- `CoreActorCollection`, `VultronPerson`, `VultronOrganization`, `VultronService`, `VultronApplication`, `VultronGroup`, `CaseStatus`, `ParticipantStatus`

`to_camel` is an AS2 serialization concern and must not appear in the core branch.

## Acceptance criteria

- All listed classes have `alias_generator` removed from `model_config`
- `test_no_core_object_has_to_camel_alias_generator` passes without `xfail` marker
- No regression in wire-layer serialization tests

sei-ahouseholder:
## Cross-reference from #2232 analysis

The acceptance criterion "All listed classes have `alias_generator` removed from `model_config`" is **not safely actionable as written** for `ParticipantStatus` (and likely `CaseStatus`): removing `alias_generator` changes the **persisted key shape** of every stored row (`caseEngagement` → `case_engagement`, `cvdRole` → `cvd_role`, `trackingId`, `embargoAdherence`, …). `AGENTS.md` forbids agents changing persistence schemas without explicit instruction, so this needs a migration plan first.

#2260 was filed during #2232 analysis to carry that plan (legacy-row read path, interaction with CM-18-006 ledger-snapshot key agreement and DRPT-02-008 extractor dual-shape handling, and the coupled `_migrate_flat_fields` validator). Treat #2260 as the planning prerequisite for this issue's `ParticipantStatus`/`CaseStatus` entries; the actor classes are likely independent and can proceed here.

Note this issue's title says ARCH-12-004 while the spec text quoted in #2260 is ARCH-12-003 — worth reconciling.

Issue #3161, opened 2026-09-04
BTBridge.execute_with_setup restores only managed_keys; /activity and context_data keys can leak across nested executions


sei-ahouseholder:
## Summary

`BTBridge.execute_with_setup` restores only its `managed_keys`
(`datalayer`, `trigger_activity_factory`, `sync_port`, `wire_render_port`, and
now `ledger_payload_object_override`) to their pre-execution state in its
`finally` block. Other keys that `setup_tree` writes to the process-global
`py_trees.blackboard.Blackboard.storage` — the `/activity` key (written when
`activity is not None`) and any `**context_data` keys (e.g. `/case_id`,
`/log_entry`) — are **not** snapshotted or restored.

Because `_BT_GLOBAL_LOCK` is a reentrant `RLock`, nested `execute_with_setup`
calls are a real, supported pattern (guarded-commit sub-trees invoked mid-node
in `case/nodes/lifecycle.py`, `case/nodes/actor.py`,
`status/nodes/case_status.py`, `case/nodes/accept_invite.py`,
`case/nodes/delegation.py`, etc.). If a nested call ever passes `activity=` or a
`**context_data` key that collides with a key the outer execution is still
relying on, the inner `setup_tree` overwrites the outer value and the inner
`finally` does not restore it — so the outer execution, on resuming after the
nested call, would read the inner execution's stale value.

## Current status: latent, not currently observable

Every existing nested `execute_with_setup` call site passes only `tree=`,
`actor_id=`, and (in one case) `sync_port=` — **never `activity=` and never
`**context_data`**. So none of them overwrite the outer `/activity` or context
keys today, and the keys they do write are all in `managed_keys` and correctly
restored. This is a latent architectural sharp edge, not a live defect: a future
nested call that passes `activity`/`context_data` would trip it.

## Why filed separately from #3101

PR #3160 (closes #3101) makes `execute_with_setup` reset the specific
execution-scoped hand-off key `ledger_payload_object_override` on every outcome,
which is correct and complete for #3101. Generalizing the bridge to
snapshot/restore *all* `setup_tree`-written keys across nested executions is a
distinct change that alters nested-execution semantics and carries its own risk;
it is out of scope for the focused #3101 fix and has no current trigger.

## Suggested resolution (for triage)

Either (a) snapshot/restore all `setup_tree`-written keys (`/activity` +
`context_data` keys) in the same `finally`, mirroring the `managed_keys`
[...]

Issue #3085, opened 2026-09-02
concern(bridge): setup_tree and execute_tree share one try block — execution errors labelled 'BT setup failed'


sei-ahouseholder:
## Summary

At `vultron/core/behaviors/bridge.py:617`, `setup_tree()` and `execute_tree()` are called inside the same `try` block. If `execute_tree` ever raises, the error is caught and labelled 'BT setup failed' rather than 'BT execution failed', misdirecting debugging.

## Details

`vultron/core/behaviors/bridge.py` line 617

`execute_tree` is designed to never raise and currently satisfies that invariant. But both calls sit inside the same try. If a future py_trees release causes `execute_tree` to raise, the exception is caught by the setup-phase except clauses and reported as 'BT setup failed with internal error'. Operators debugging the failure look at setup configuration rather than tree logic.

## Recommendation

Split into two separate try/except pairs: one for `setup_tree`, one for `execute_tree`.

Surfaced by code review of PR #3078.

sei-ahouseholder:
Planned jointly with #3084 and #3161 on branch `plan/3084-bridge-exception-handler-dedup`.

sei-ahouseholder:
✅ Planned.

- Docs PR: https://github.com/CERTCC/Vultron/pull/3509
- Implementation issue: #3510

Issue #3080, opened 2026-09-02
bug(bridge): VultronError in execute_tree logged at ERROR level, violating AGENTS.md rule 310


sei-ahouseholder:
## Summary

The `except VultronError` block at `vultron/core/behaviors/bridge.py:513` calls `logger.exception()` (ERROR level with full traceback) for an exception classified as `internal_error=False` — a protocol outcome, not an unexpected failure. This violates AGENTS.md rule 310: "designed self-healing recovery paths log WARNING/INFO, never ERROR".

## Details

`vultron/core/behaviors/bridge.py` line 513

The block comment says: "Treated as attributable to the protocol rather than to us". Yet `logger.exception()` emits at ERROR with a traceback. On a coordinator processing hundreds of reports, every expected protocol rejection floods structured-log dashboards at ERROR level, masking genuine internal errors.

## Fix

Change `logger.exception()` to `logger.warning()` (without exc_info) in the `except VultronError` block of `execute_tree()`.

Surfaced by code review of PR #3078.

PR #3533, opened 2026-09-22
refactor(bridge): harden BTBridge exception handlers and managed keys

### LINKED ISSUE #3510: refactor(bridge): harden BTBridge exception handlers and managed-keys (#3084 #3085 #3161)
## Summary

Three hardening changes to `vultron/core/behaviors/bridge.py`:

1. **Extract `_exception_result` helper** (#3084): A private method on `BTBridge` that builds the error message, logs at the correct level (WARNING for `VultronError`, ERROR+traceback for `Exception`), and returns `BTExecutionResult`. Collapses four near-duplicate except-block bodies to four one-line call sites, preventing future log-level divergence (which caused #3080).

2. **Split the combined try block** (#3085): `execute_with_setup` currently wraps both `setup_tree` and `execute_tree` in one try/except. Split into two separate try blocks — one for `setup_tree` (labelled "BT setup failed") and one for `execute_tree` (labelled "BT execution failed") — so execution errors are never mislabelled as setup errors. `execute_tree` has its own internal catch-all, but wrapping it in its own try in `execute_with_setup` keeps the "never escapes" contract even if that changes.

3. **Extend managed_keys to cover `activity` and `context_data`** (#3161): `setup_tree` writes the `activity` key and all `**context_data` keys to the blackboard. These are currently not snapshotted/restored by the `finally` block. A nested `execute_with_setup` call that passes the same keys would overwrite the outer execution's values and not restore them. Fix: dynamically extend `managed_keys` with `["activity"] + list(context_data.keys())` before `previous_values` is computed. `actor_id` is intentionally excluded (outer execution holds its own actor id as a Python local).

## Acceptance Criteria

- [ ] AC-1: `BTBridge._exception_result(self, e, *, prefix, internal_error=False)` exists; each of the four except blocks in `execute_tree` and `execute_with_setup` calls it
- [ ] AC-2: `execute_with_setup` uses two separate try blocks — one covering only `setup_tree`, one covering only `execute_tree`
- [ ] AC-3: `managed_keys` in `execute_with_setup` is extended with `["activity"] + list(context_data.keys())` before `previous_values` is snapshotted
- [ ] AC-4: Regression test: nested `execute_with_setup` where inner call passes `activity=<some value>` does not clobber the outer execution's `/activity` blackboard key after the inner call returns
- [ ] AC-5: All existing `test_bridge.py` tests pass; no test behaviour changes except those that previously tested the now-consolidated error construction

## Reference

Source: #3084, #3085, #3161
Docs PR: https://github.com/CERTCC/Vultron/pull/3509
Notes: `notes/bt-pitfalls.md` § "Execution-Scoped Hand-off Keys Are Reset by the Bridge, Not by a Node"

### LINKED ISSUE #3516: bug(bridge): nested execute_with_setup leaks /actor_id to later siblings of the calling node
## Summary

`BTBridge.execute_with_setup` did not restore the `/actor_id` blackboard key, so
a nested `execute_with_setup` running as a different actor left that actor
visible to every **later sibling** of the node that made the nested call.

## Why the exclusion looked safe

`actor_id` was deliberately left out of `managed_keys` on the reasoning that
"the outer execution holds its own actor id as a Python local, not from the
blackboard". That is true of `execute_with_setup` itself and **false of every
node**:

`DataLayerCondition`, `DataLayerAction`, `DataLayerConditionWithPorts` and
`DataLayerActionWithPorts` (all in `vultron/core/behaviors/helpers.py`) read
`/actor_id` into `self.actor_id` in **`initialise()`**, not `setup()`. py_trees
calls `initialise()` on every tick in which the node was not `RUNNING`, so the
read is once per tick, not once per execution.

## Reproduction

A probe over a `Sequence(memory=False)` whose first child runs a nested
`execute_with_setup` as actor B and whose second child is a `DataLayerAction`:

```text
outer status: Status.SUCCESS
sibling saw actor_id: ['https://b.example/actors/inner']
LEAK
post-run /actor_id: https://b.example/actors/inner
```

The sibling resolved actor B while its own tree was executing as actor A.

## Protocol consequence

Reachability required `self.datalayer.actor_id` to differ from the blackboard
actor, which is exactly `_store_for_actor`'s documented foreign-authority
fall-through — a case whose CASE_MANAGER sits on another container after a
handoff (CP-08-003). Three nested call sites pass `self.datalayer.actor_id`
rather than the executing actor: `case/nodes/leave/advance.py` (two) and
`sync/nodes/close_case_effect.py`.

On that path, in `OwnerLeaveSeq` (`case/receive_close_case_tree.py`), the
siblings downstream of `AdvanceParticipantToRMClosedNode` include
`CommitCaseActorRMClosedEntryNode`, which feeds its resolved `self.actor_id`
into `build_add_participant_status_snapshot(...)`. The leaked actor therefore
reaches `payloadSnapshot.actor` — a CLP-07-003 identity fault in a canonical
ledger entry.

## Fix

Add `actor_id` to `execute_with_setup`'s `managed_keys`. This is safe rather
than merely safer: the restore runs in the **nested** call's own `finally`,
after its tree has finished, and puts back exactly the value that same call
snapshotted, so it cannot overwrite a live execution.

## Notes

Found while implementing #3510, whose AC-3 states the exclusion is intentional.
That premise does not hold, so the fix ships in the same PR rather than
preserving a documented leak. Both suites pass unchanged with the restore in
place (unit 9110, integration 1284), confirming nothing depended on the leak.

Regression guards: `test_an_inner_actor_id_does_not_clobber_the_outer_one` and
`test_a_later_sibling_reads_its_own_trees_actor` in
`test/core/behaviors/test_bridge.py`. The second asserts on what a sibling
*node* resolves, because asserting on raw blackboard storage does not exercise
the `initialise()` re-read that made this reachable.

Normative requirement added: BT-17-007.

### LINKED ISSUE #3534: concern(bridge/inbox): blackboard snapshot-and-restore is implemented twice
## Summary

`BTBridge.execute_with_setup` and `inbox/_process_payload.process_payload` each
carry their own copy of the same blackboard snapshot-and-restore logic, including
the same `/`-prefixed-alias handling and the same `had_value` restore loop.

## The duplication

`vultron/core/behaviors/bridge.py` (inside `execute_with_setup`):

```python
key_aliases: set[str] = set()
for key in managed_keys:
    key_aliases.add(key)
    key_aliases.add(f"/{key}")
previous_values = {
    key: (key in storage, storage.get(key)) for key in key_aliases
}
...
for _key, (_had_value, _value) in previous_values.items():
    if _had_value:
        storage[_key] = _value
    else:
        storage.pop(_key, None)
```

`vultron/core/behaviors/inbox/_process_payload.py` (`_save_inbox_keys` /
`_restore_inbox_keys`):

```python
for key in ALL_INBOX_KEYS:
    for variant in (key, f"/{key}"):
        saved[variant] = (variant in storage, storage.get(variant))
...
for key, (had_value, value) in saved.items():
    if had_value:
        storage[key] = value
    else:
        storage.pop(key, None)
```

Same `(bool, value)` tuple shape, same both-aliases handling, same
restore-or-pop branch. CS-22-001 ("three similar lines is a signal to extract;
copy-pasting a function body is not acceptable").

## Why it matters beyond style

The two key sets are currently disjoint — `ALL_INBOX_KEYS`
(`inbox/nodes/pipeline.py`) is nine `inbox_*` keys, none of which appear in
`managed_keys` — so there is no live conflict. The risk is divergence:

- A future fix to one restore path does not reach the other. This already
  happened once in the other direction: `managed_keys` grew twice (#3161, #3516)
  because keys were missing from it, and neither change prompted a look at the
  second implementation.
- `process_payload` builds its own `py_trees.trees.BehaviourTree` and calls
  `bt.setup()` directly rather than going through `BTBridge`, so it also does not
  benefit from the bridge's classification, ports rebinding, or leadership guard.
  The duplicated snapshot logic is a symptom of that larger bypass.

## Suggested direction

Extract the snapshot/restore pair into one helper — the natural home is beside
`BTBridge` or in a small `blackboard_scope` module — taking an iterable of key
names and returning a context manager, then use it from both call sites. The
larger question of whether `process_payload` should run through `BTBridge` at all
is separate and probably wants its own issue.

## Provenance

Found during pre-PR review of #3510 / #3516 (PR
https://github.com/CERTCC/Vultron/pull/3533). Not fixed there: the inbox pipeline
is a subsystem that PR does not otherwise touch, and the PR had already expanded
once to absorb #3516.

### LINKED ISSUE #3545: bug(test-isolation): _STORE_CLAIMANTS is never reset, so an earlier test's actor id makes a later no-warning assertion fail
## Summary

`_STORE_CLAIMANTS` in `vultron/adapters/driven/datalayer_sqlite/engine.py:208` is a
module-level dict that is never reset between tests. It records the first actor id
seen for each resolved store URL, purely so `get_actor_engine` can warn on a
cross-authority slug collision (`engine.py:232-241`). Because it survives for the
whole pytest process, a test that merely *uses* one actor id poisons the guard for
every later test that asserts the warning is absent.

Two test fixtures collide on the slug `test-actor`:

- `https://example.org/actors/test-actor` — `test_actor_id` in `test/core/behaviors/test_bridge.py`
- `https://test.example/api/v2/actors/test-actor` — the `datalayer` fixture in the same file

Both resolve to `sqlite:///file:vultron-test-actor?mode=memory&cache=shared&uri=true`.

## Reproduction

```bash
uv run pytest test/core/behaviors test/core/use_cases/received/test_create_report_received.py
```

```text
FAILED test/core/use_cases/received/test_create_report_received.py::TestDuplicateReportHandling::test_submit_report_no_warning_on_duplicate_report
FAILED test/core/use_cases/received/test_create_report_received.py::TestDuplicateReportHandling::test_create_report_no_warning_on_duplicate_report
```

Both assert "No WARNING should be emitted …" and both receive the
`Store … is shared by two distinct actor ids` warning left behind by the earlier
directory. Each file passes on its own.

## Why CI is green

CI's job ordering does not put `test/core/behaviors` ahead of
`test/core/use_cases/received` in the same process, so the collision is latent
rather than active. It is a live flake risk for anyone running a subset locally,
and for any future change to test ordering, sharding, or `-p xdist` distribution.

## Not caused by a branch

Reproduced identically at `2244b3b6c` (the merge-base of PR #3533, where it was
found). Both colliding fixtures predate that commit.

## Suggested fix

Give the guard a reset seam and use it from an autouse fixture, the same way
`_ENGINES` disposal is already a supported reset. The docstring notes
`_STORE_CLAIMANTS` is deliberately kept separate from `_ENGINES` "so disposal (a
legitimate reset) does not erase the record of which id claimed a store" — so the
fix should add an explicit test-scoped reset rather than folding it into engine
disposal. Alternatively, give the two fixtures non-colliding slugs, though that
treats the symptom and leaves the next slug collision to be rediscovered.

## Acceptance criteria

- [ ] AC-1: `uv run pytest test/core/behaviors test/core/use_cases/received/test_create_report_received.py` passes.
- [ ] AC-2: The fix is an explicit reset seam for `_STORE_CLAIMANTS`, not a rename of the colliding fixtures alone.
- [ ] AC-3: A regression guard fails if the process-global is reintroduced without a reset.

## Routing note

Parented to #2619 (test correctness) rather than #3129, which is scoped to the
*demo runner* harness rather than the unit suite.

## Reference

Found during `/pr-ship` triage of #3533.

sei-ahouseholder:
- Closes #3510
- Closes #3516
- Closes #3534
- Closes #3545

## Summary

Collapses `BTBridge`'s near-duplicate exception handlers into one helper, splits
`execute_with_setup`'s combined `try` so a setup error and an execution error are
never confused, and extends the execution-scoped key restore to `activity`,
every `context_data` key, and `actor_id`.

## Changes

- **`vultron/core/behaviors/bridge.py`**
  - Adds `_exception_result(e, *, prefix, internal_error=False)`. One input now
    drives the message, the log level, and the `internal_error` flag, so the six
    classifying handlers cannot diverge again — the log-level divergence that
    caused #3080 is no longer expressible (CS-22-001). The `bt.shutdown()` guard
    in `execute_tree`'s `finally` deliberately does not route through it: it
    builds no result, because its job is to preserve the one already returned.
  - Splits `execute_with_setup`'s single `try` into a setup block and an
    execution block inside one shared `finally`. An error raised during the ticks
    was reported as `"BT setup failed"`, sending a reader to the wiring when the
    bug was in a node (#3085). `execute_tree` still has its own catch-all, so the
    execution-phase handlers are the guarantee rather than the mechanism — they
    keep "a BT execution never escapes the bridge" true if that catch-all is ever
    narrowed.
  - Adds `activity` and every `context_data` key to `managed_keys` before
    `previous_values` is snapshotted, so a nested `execute_with_setup` hands the
    outer execution's values back instead of leaving its own behind for the outer
    tree's remaining ticks (#3161).
  - Adds `actor_id` to `managed_keys` (#3516) — see below.
- **`specs/behavior-tree-integration.yaml`**: adds `BT-17-007` making the restore
  contract normative, and corrects `BT-17-003`'s rationale, which still claimed
  the bridge cleans only `datalayer` and `trigger_activity_factory`.
- **`notes/bt-pitfalls.md`**: drops the `pending #3510` marker, documents the
  helper and the `/actor_id` restore, and fixes two stale passages the change
  exposed — the same "cleans only two keys" claim, and a worked example naming
  `broadcast_activity_id`, which no longer exists anywhere in `vultron/`.
- **`test/architecture/test_no_broad_except_outside_bt_update.py`**: raises the
  CS-23-001 exclusion count for `bridge.py` from 3 to 4 for the handler the split
  adds, and records which of the four classify.
- **`test/core/behaviors/test_bridge.py`**: new tests covering phase
  labelling, helper log levels and message/`errors` agreement at every boundary,
  and nested-execution key isolation for `activity`, `context_data` keys and
  `actor_id`.
- **`test/architecture/test_bridge_managed_keys_cover_setup_writes.py`** (new):
  asserts `setup_tree`'s literal `register_key()` names are a subset of
  `managed_keys`. #3161 and #3516 were the same defect — a key `setup_tree` writes
  that `execute_with_setup` does not restore — and nothing connected the two
  lists, so each omission was a silent one-line mistake. Verified to catch both.

- **`vultron/core/behaviors/blackboard_scope.py`** (new): extracts the blackboard
  snapshot-and-restore that `execute_with_setup` and `inbox/_process_payload` each
  carried a copy of (#3534, CS-22-001). Folded in here rather than deferred because
  this PR is the second witness that the logic drifts silently — #3161 and #3516
  were both one-line omissions in the bridge copy alone. Both call sites now
  delegate; 11 new tests pin the two details a re-implementation gets wrong
  (absent ≠ `None`, and both key spellings).
- **`vultron/adapters/driven/datalayer_sqlite/engine.py`**: adds
  `reset_store_claimants()`, and `test/conftest.py` calls it from the existing
  autouse store-disposal fixture (#3545). `_STORE_CLAIMANTS` is deliberately not
  cleared by engine disposal — a claim must outlive an engine for the
  cross-authority warning to mean anything — so it lived for the whole pytest
  process, and one test using `.../actors/test-actor` made a later test's
  "no WARNING" assertion fail on a collision it did not cause. Found while
  reviewing this PR; unrelated to the bridge, hence its own issue and its own
  closing bullet.
- **Review follow-ups** (no issue filed — the diff is the record): rebuilt
  `test_every_fixed_managed_key_is_restored`, which was vacuous for 5 of its 7
  keys (removing them from `managed_keys` left it green); taught
  `test_bridge_managed_keys_cover_setup_writes.py` to read positional
  `register_key("x", ...)` calls and to fail on a key it cannot resolve
  statically; added
  `test/architecture/test_bridge_exception_handlers_use_helper.py`, since the
  pre-existing broad-except ratchet counts only `except Exception` and so pinned
  3 of the 6 classifying handlers — the three `except VultronError` ones, where
  #3080 actually happened, were unguarded; added restore coverage for the FAILURE,
  tick-raises and setup-raises outcomes that BT-17-007 requires.

## Why #3516 is in this PR

#3510's AC-3 states that `actor_id` is "intentionally excluded" from the restore
because the outer execution holds its actor id as a Python local. That is true of
`execute_with_setup` and **false of every node**: all four node bases in
`behaviors/helpers.py` re-read `/actor_id` into `self.actor_id` in
`initialise()`, which py_trees runs on every tick in which the node was not
`RUNNING`. A probe over a `Sequence(memory=False)` confirmed a sibling
`DataLayerAction` resolving the *inner* actor after a nested call.

Reachability is narrow but real: three nested call sites pass
`self.datalayer.actor_id`, which diverges from the blackboard actor only on
`_store_for_actor`'s documented foreign-authority fall-through (a case whose
CASE_MANAGER sits on another container after a handoff, CP-08-003). On that path
`OwnerLeaveSeq` ticks its ledger-committing node downstream of
`AdvanceParticipantToRMClosedNode`, so the leaked actor reached
`payloadSnapshot.actor` — a CLP-07-003 identity fault in a canonical entry.

Restoring is safe rather than merely safer: the restore runs in the **nested**
call's own `finally`, after its tree has finished, and puts back exactly the
value that same call snapshotted, so it cannot overwrite a live execution. Both
suites pass unchanged with it in place, confirming nothing depended on the leak.

## Verification

- Unit suite: 9413 passed, 443 skipped, 21 xfailed, 0 failed
- Full suite incl. integration: 10697 passed, 443 skipped, 22 xfailed, 0 failed
- Black, flake8, mypy, pyright clean; `markdownlint-cli2` clean;
  `spec-lint` at its pre-change warning baseline
- Every test here was confirmed non-vacuous by reverting the corresponding fix:
  removing the `managed_keys` line fails the nested-isolation tests, collapsing
  the split `try` fails the phase-labelling tests, and removing `"actor_id"`
  fails both `actor_id` tests
- [x] AC-1: `_exception_result` exists; every classifying handler calls it. AC-2's
  split raises the count from four to six, and all six route through it
- [x] AC-2: `execute_with_setup` uses two separate try blocks, labelled
  `"BT setup failed"` and `"BT execution failed"`
- [x] AC-3: `managed_keys += ["activity", *context_data.keys()]` before
  `previous_values` is snapshotted. `actor_id` is now included rather than
  excluded — see "Why #3516 is in this PR"
- [x] AC-4: `test_an_inner_activity_does_not_clobber_the_outer_one`
- [x] AC-5: all pre-existing `test_bridge.py` tests pass unchanged; no test
  behaviour was altered

## Follow-ups filed

- **#3535** — audit the codebase's "intentionally excluded / no cleanup needed"
  justifications. #3516 is the second witness for the claim that a confident
  rationale attached to an exemption suppresses scrutiny without earning it;
  the first is already in the learnings queue. Not fixed here: it is a
  codebase-wide audit, not a change to this code.


sei-ahouseholder (commented):
## PR Triage: #3533 — refactor(bridge): harden BTBridge exception handlers and managed keys

**Linked issues**: #3510 (harden BTBridge exception handlers and managed-keys), #3516 (nested execute_with_setup leaks /actor_id to later siblings)
**Changed files**: 7 files — core/behaviors, specs, notes, test/architecture
**CI status**: ✅ passing (24 checks)
**Base branch**: main
**Needs integration tests**: yes — ran (1284 passed), demo-integration jobs green

---

### Findings

|---|---|---|---|---|
| phase10-vacuous-managed-key-parametrizations-0 | test-coverage | ❌ FAIL | `test_every_fixed_managed_key_is_restored` is vacuous for 5 of 7 keys — deleting them from `managed_keys` leaves it green | fix-now |
| phase10-ratchet-misses-positional-register-key-0 | test-coverage | ⚠️ IMPROVE | New ratchet reads only `register_key(key=…)`; a positional call is silently uncounted | fix-now |
| phase5-vultron-error-handlers-unratcheted-0 | spec-conformance | ⚠️ IMPROVE | The "divergence no longer expressible" guarantee is ratcheted for 3 of 6 handlers | fix-now |
| phase5-restore-untested-on-failure-outcomes-0 | spec-conformance | ⚠️ IMPROVE | BT-17-007 says "every outcome"; new keys tested only on SUCCESS | fix-now |
| phase9-suggested-roles-example-ambiguous-0 | notes-currency | ⚠️ IMPROVE | `suggested_roles` spans a managed and an unmanaged key; example doesn't say which | fix-now |
| phase2-3534-deferral-not-gated-0 | issue-linkage | ⚠️ IMPROVE | #3534 deferred without a Gate-1 measured remainder | defer-ask |
| phase11-preexisting-order-dependent-failure-0 | linter-ci-status | 🎫 NEW-ISSUE | Pre-existing order-dependent failure via `_STORE_CLAIMANTS` global | defer-ask |

**Total**: 1 FAIL · 5 IMPROVE · 1 NEW-ISSUE

---

### The FAIL, in detail

`test_every_fixed_managed_key_is_restored` parametrizes over all seven fixed managed keys, but only `datalayer` and `actor_id` have their sentinel actually overwritten during the execution. For the other five the pre-value equals the post-value whether or not the key is in `managed_keys`. Verified by deleting `is_leader`, `sync_port`, `wire_render_port`, `trigger_activity_factory` and `ledger_payload_object_override` from `managed_keys` (`bridge.py:729-745`) — all seven parametrizations still pass.

Causes, per key:

- `is_leader` — the fixture is `BTBridge(datalayer=…)`, so `_is_leader_explicit` is `False` and `setup_tree` never writes `/is_leader`.
- `trigger_activity_factory` / `sync_port` / `wire_render_port` — with no injected port, `_inherited_port()` reads the sentinel back off the blackboard and writes the same object again.
- `ledger_payload_object_override` — `AlwaysSucceed` never writes it.

This is the test that carries BT-17-007's "for each managed key" verification clause, so that clause is not yet satisfied. The production behavior is correct — this is a coverage defect only.

---

### Verified correct

Claims that were attacked and held up:

- **The `errors`-is-empty claim** (`bridge.py:594-599`) is **true**. `errors` is appended at exactly two sites (:564 INVALID, :579 max-iterations), each immediately followed by a `return`. The removed `errors.append(...)` dropped nothing.
- **`logger.exception` context** — all three `internal_error=True` call sites (:607, :812, :828) are lexically inside `except Exception as e:`, so tracebacks are real; the new tests assert level *and* `exc_info`.
- **The `actor_id` fix is real.** Traced at the merge base: the later sibling resolves the inner actor (leak reproduced). On this branch it resolves its own tree's actor. `_RecordResolvedActor(DataLayerAction)` exercises the genuine `initialise()` re-read, and `assert inner_seen == [inner_actor]` blocks a vacuous pass — exactly what BT-17-007's verification clause asks for.
- **Restore logic** — absent-before keys correctly `pop` rather than set-to-`None`; `key_aliases` is a `set`, so a `context_data`/fixed-key collision is harmless; `+=` precedes the alias loop and snapshot, so dynamic keys get identical `/`-prefix handling; non-identifier `context_data` keys (`"a/b"`, `"1bad"`) work end-to-end.
- **Try-split** — the shared `finally` wraps both inner `try`s, so all four early returns pass through the restore; `bt` can never be referenced unbound; no new escaping path.
- **No message-format dependents** outside this branch; setup-phase strings are byte-identical.
- **Ratchet counts reconcile** — 4 broad (3 classifying + the `bt.shutdown()` guard) and 6 classifying (3 broad + 3 `VultronError`).
- **No ADR warranted** — MS-11-005 applies; the spec entry is the right vehicle.
- PR body format compliant, notes frontmatter updated, `broadcast_activity_id` fully removed from the tree, learnings file matches peer format.

---

*Triage artifact: `.claude/pr-3533-triage.json`*

sei-ahouseholder (commented):
## PR Execute: #3533

**Findings processed**: 7 of 7 — 7 fixed, 0 skipped, 0 deferred, 0 halted
**CI**: ✅ 24/24 passing (1 iteration)
**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, synced at `7bf7f3634`
**Commit**: `4c97a088d`

---

### Fixes applied

| Finding | Sev | Outcome | What changed |
|---|---|---|---|
| phase10-vacuous-managed-key-parametrizations-0 | ❌ FAIL | fixed | Rebuilt `test_every_fixed_managed_key_is_restored` so all 7 keys are load-bearing |
| phase10-ratchet-misses-positional-register-key-0 | ⚠️ IMPROVE | fixed | Ratchet reads positional `register_key`, fails on unresolvable keys, pins call-site scope |
| phase5-vultron-error-handlers-unratcheted-0 | ⚠️ IMPROVE | fixed | New ratchet covers all 6 classifying handlers, not just the 3 broad ones |
| phase5-restore-untested-on-failure-outcomes-0 | ⚠️ IMPROVE | fixed | Restore coverage for FAILURE, tick-raises and setup-raises outcomes |
| phase9-suggested-roles-example-ambiguous-0 | ⚠️ IMPROVE | fixed | Notes name the key *form*, not the bare noun |
| phase2-3534-deferral-not-gated-0 | ⚠️ IMPROVE | fixed → closes #3534 | Extracted shared `blackboard_scope` module |
| phase11-preexisting-order-dependent-failure-0 | 🎫 NEW-ISSUE | fixed → closes #3545 | `reset_store_claimants()` + autouse reset |

**Every fix was verified by mutation** — reverting the corresponding production change fails the new assertion. The three that matter most:

- Deleting `is_leader`, `sync_port`, `wire_render_port`, `trigger_activity_factory` and `ledger_payload_object_override` from `managed_keys` now fails exactly those 5 parametrizations. **Before this change all 7 passed** — the test asserted nothing for 5 of its keys.
- A positional `register_key("positional_key", ...)` now fails the ratchet. Before, `missing == set()` and `len(registered) == 7` — a silent pass, in the ratchet whose entire purpose is catching that omission.
- Replacing the setup-phase `except VultronError` with an inline result logged at ERROR — **#3080's exact shape** — now fails.

### Two gates run

Both were `defer-ask`, and the user chose **fix now** on both:

- **#3534** — measured remainder presented (2 call sites, ~25 lines, 1 helper). Extracted `vultron/core/behaviors/blackboard_scope.py`; both `execute_with_setup` and `inbox/_process_payload` delegate to it. 11 new tests pin the two details a re-implementation gets wrong: absent ≠ `None`, and both key spellings.
- **#3545** — filed first (Bug, parent #2619, milestone 25, on Project #24), then a size estimate given (~70 diff lines, ~4%, doesn't move the size label). `_STORE_CLAIMANTS` is deliberately not cleared by engine disposal — a claim must outlive an engine for the cross-authority warning to mean anything — so the reset is its own named seam, called from the existing autouse fixture.

### Verification

- Unit: **9413 passed**, 443 skipped, 21 xfailed, 0 failed
- Full incl. integration: **10697 passed**, 443 skipped, 22 xfailed, 0 failed
- black, flake8 (CC gate), mypy, pyright, markdownlint-cli2 all clean; `spec-lint` exit 0, no warning on `BT-17-007`
- `pytest test/core/behaviors test/core/use_cases/received/test_create_report_received.py` → 2302 passed (2 failed before the #3545 fix)

### xfail ratchet: 22 xfails, 0 unmanaged

Two apparent hits, both cleared on inspection rather than filed:

- `test_no_core_object_has_to_camel_alias_generator` names closed #1991 — but its reason reads "Goal state tracked in #2288 and #2289 (supersedes closed #1991)", both **OPEN**. The closed number is a deliberate historical note.
- `choose_preferred_embargo` carries no issue number **by design**: its reason cites ADR-0100, "emit-only by decision". It asserts a negative that is permanently correct, so a tracking issue could never close.

### Sync

Merged `origin/main` (`1412a1d71`) cleanly — ADR-0101 (spec-item format is field presence) and `notes/spec-authoring-rules.md`. `spec-lint` was re-run after the merge to confirm `BT-17-007` still conforms to the new format rule.

---

*Execute artifact: `.claude/pr-3533-execute.json`*
*Next step: `/pr-verify`*

sei-ahouseholder (commented):
## PR Verify: #3533

**Merge state**: ✅ MERGEABLE (CLEAN) — base `main`, not a draft, no conflict markers
**CI**: ✅ 22 passing, 2 skipped, 0 failing (the two `Notify CI failure` jobs are `skipping` by design)
**Integration tests**: ✅ all four Demo Integration jobs green (`fcv-reject`, `fcvcv`, `fv`, `fvcv-handoff`)
**Artifact integrity**: ✅ 7 results = 7 triage findings; `merge_state.synced: true`

---

### Per-finding verdicts

| Finding | Sev | Execute outcome | Verdict |
|---|---|---|---|
| phase10-vacuous-managed-key-parametrizations-0 | ❌ FAIL | fixed | ✅ CONFIRMED |
| phase10-ratchet-misses-positional-register-key-0 | ⚠️ IMPROVE | fixed | ✅ CONFIRMED |
| phase5-vultron-error-handlers-unratcheted-0 | ⚠️ IMPROVE | fixed | ✅ CONFIRMED |
| phase5-restore-untested-on-failure-outcomes-0 | ⚠️ IMPROVE | fixed | ✅ CONFIRMED |
| phase9-suggested-roles-example-ambiguous-0 | ⚠️ IMPROVE | fixed | ✅ CONFIRMED |
| phase2-3534-deferral-not-gated-0 | ⚠️ IMPROVE | fixed → closes #3534 | ✅ CONFIRMED |
| phase11-preexisting-order-dependent-failure-0 | 🎫 NEW-ISSUE | fixed → closes #3545 | ✅ CONFIRMED |

**7 CONFIRMED · 0 UNRESOLVED · 0 MISSING-COMMIT · 0 NOTED**

### How each was checked at HEAD

Checked against `HEAD` (`7bf7f3634`), not just the commit diff, so a fix applied and later reverted would show up:

- **FAIL fix** — all five load-bearing constructs present in `test_bridge.py` at HEAD: `is_leader=lambda: True` (1663), `trigger_activity=cast(Any, object())` (1664), `WriteLedgerOverride()` (1674), `RecordManagedKeyDuringTick(key, observed)` (1675), and the non-vacuity assertion `observed["during"] is not sentinel` (1686).
- **Ratchet fix** — `call.args[0] if call.args else None`, `_SANCTIONED_DYNAMIC_KEYS`, and `test_every_register_key_call_lives_in_setup_tree` all at HEAD.
- **Handler ratchet** — `test/architecture/test_bridge_exception_handlers_use_helper.py` exists at HEAD.
- **Failure-outcome coverage** — both new test functions at HEAD.
- **#3534** — `vultron/core/behaviors/blackboard_scope.py` at HEAD; `bridge.py` calls both `snapshot_keys(storage, managed_keys)` and `restore_keys(storage, previous_values)`.
- **#3545** — `reset_store_claimants` defined in `engine.py` and called from `test/conftest.py`, both at HEAD.
- Re-ran every touched test file at HEAD: **171 passed**.
- #3534 and #3545 both confirmed **OPEN** (they close on merge via their `Closes` bullets).

### Notes

- **Execute artifact schema corrected.** The artifact was first written with `findings` instead of `results` and a `merge_state` block lacking `synced`/`sync_commit_ref`. That was a serialization error in execute's Phase 6, not a defect in the PR — normalized to the documented schema so this phase's integrity and sync gates read real data rather than passing on absent fields.
- **PR body counts refreshed.** The Verification section still read `9391 passed … (25 new)` and a separate `1284` integration line from before execute's fixes. Now `9413` unit / `10697` full-suite, matching the actual runs.
- **xfail ratchet**: 22 xfails, 0 unmanaged. Two apparent hits were cleared rather than filed — `test_no_core_object_has_to_camel_alias_generator` names closed #1991 only as a "supersedes" note while tracking live #2288/#2289, and `choose_preferred_embargo` is an ADR-0100 "by decision" exemption asserting a permanently-correct negative, so a tracking issue could never close.

---

*Verified against HEAD `7bf7f3634`. Session artifacts cleaned up.*

Issue #2964, opened 2026-09-01
bug(case_status): _coerce_vf_or_none and _coerce_d_or_none raise KeyError instead of ValueError on unknown names


sei-ahouseholder:
## Problem

`vultron/wire/as2/vocab/objects/case_status.py` has two Pydantic field validators with the same `KeyError` bug that was fixed in `dimensions.py` (#2892):

1. **Line ~87** — `_coerce_vf_or_none`: uses `CS_vf[s]` (enum `__getitem__`) which raises `KeyError` on unknown input. Pydantic v2 only catches `ValueError`/`AssertionError` from field validators, so `KeyError` propagates and crashes the inbox pipeline with a 500-class error instead of a clean 422 `ValidationError`.

2. **Line ~97** — `_coerce_d_or_none`: same `KeyError` pattern for `CS_d[s]`.

`dimensions.py` already has the correct `try/except KeyError → raise ValueError` wrapping (the fix from #2892). These sibling validators were missed.

## Failure scenario

Inbound `CaseStatus` wire object with unrecognised `vfState` or `dState` string causes uncaught `KeyError` to propagate through the inbox pipeline, producing a 500-class error instead of a 422 `ValidationError`.

## Fix

Wrap both coerce functions with `try: ... except KeyError as e: raise ValueError(...) from e` — identical pattern to the fix already applied in `dimensions.py`.

## Related

- #2892 (closed — same bug in `dimensions.py`)
- Discovered during code review of PR #2928 (ISSUE-2458)

sei-ahouseholder:
Claimed on branch `bug/2964-case-status-keyerror`.

Issue #2692, opened 2026-08-26
Wire-type namespace and immutability: complete alias removal and freeze the wire/core boundary


sei-ahouseholder:
## Design idea

Complete the alias-generator removal chain and enforce wire-type immutability so the wire/core boundary is structurally irrevocable.

## Rationale

Calved 2026-08-26 from **#2222 Enforce the wire/domain/persistence layer boundary**, which grew beyond a single coherent design idea. This epic takes the wire-type structural side: removing `alias_generator` from core classes, separating the `type_` namespaces, and marking wire branch objects `frozen=True` so crossing the boundary becomes a type error rather than a runtime surprise.

## Tracked Work

- [ ] #2288 — Remove alias_generator from the six core actor classes; render actor AS2 responses from as_* wire types *(size:M)*
- [ ] #2289 — Remove alias_generator from ParticipantStatus/CaseStatus, delete the flat-field shim behind reject-guards, retire the #1991 ratchet *(size:L)*
- [ ] #2403 — Structural fix: disjoint type_ namespaces to eliminate wire/core VOCABULARY collision
- [ ] #2652 — Add frozen=True to wire branch as_Object and ratchet test *(size:S)*
- [ ] #2653 — Redesign TriggerActivityPort to return frozen wire blob instead of dict *(size:L)*

## Reference

- Parent epic: #2222 (Enforce the wire/domain/persistence layer boundary)
- ADRs: 0017, 0034, 0035, 0063
- Specs: ARCH-12 (ARCH-12-003, ARCH-12-004), ARCH-20
- Notes: notes/domain-model-separation.md, notes/core-wire-rendering-port.md

sei-ahouseholder:
## Closing: the authority for this epic has been reversed

This epic is being closed as **not planned**, not because its problems were imaginary but because the architecture it was built to enforce no longer exists. Writing down why, in detail, so the decision is legible to anyone who finds a survivor issue and wonders where its parent went.

### What happened

This epic was calved from #2222 with **ADR-0017** as its authority — the wire/core split — and its thesis was to *strengthen* that boundary: remove `alias_generator` from core so core is alias-free, keep the wire and core `type_` namespaces disjoint, and freeze wire objects so received artifacts cannot be mutated.

[...]

PR #3540, opened 2026-09-22
feat: delete 25 paired as_* domain classes (ADR-0099 detail 3)

### LINKED ISSUE #3487: Delete the 25 straightforward paired as_* domain classes
Source: **ADR-0099** detail 3 (#2670).

## Scope

27 `as_*` classes pair with a core domain class. This issue covers the **25 that differ from their core counterpart only by field spelling and validation strictness**. The two status classes (`as_CaseStatus`, `as_ParticipantStatus`) differ by field *name* and drag the `as_CaseParticipant` cascade with them; they are a separate issue.

The message-shape classes stop naming the wire copy and name the core class instead:

```python
class _AddReportToCaseActivity(as_Add):
    object_: VulnerabilityReport       # was as_VulnerabilityReport
    target:  VulnerabilityCase         # was as_VulnerabilityCase
```

The 49 unpaired `as_*` classes are pure AS2 vocabulary (`as_Like`, `as_Follow`, `as_Collection`, `as_Link`, `as_Tombstone`, …) and stay. The 47 `_XxxActivity` message shapes stay. Only the duplicates go.

## Why this is safe to do in batches

Each pair is independent. Take one class per commit or per small batch, and the suite tells you immediately. Recommended order: start with the ones whose projection overrides have the least structural work (`case_ledger_entry` has zero), and leave `vulnerability_case` (20 structural operations) until last.

## Acceptance Criteria

- [ ] AC-1: The 25 paired classes are deleted; no `as_*` class duplicates a core domain type except the two status classes.
- [ ] AC-2: Every message-shape slot that named a deleted class now names the core class.
- [ ] AC-3: For each deleted pair, a test asserts the serialized AS2 payload is unchanged — the same acceptance test shape as `test/core/models/test_dimension_bare_serialization.py::test_core_participant_status_as2_output_matches_wire`.
- [ ] AC-4: No projection behaviour is dropped silently. Any `from_core`/`to_core` body doing structural work has that work either relocated or explicitly declared unnecessary in the commit message.
- [ ] AC-5: Full unit suite and integration suite pass.

## Blocked by

- the ratchet replacement — the ARCH-22 ratchet is two-sided and fails when the violation set **shrinks**, so deleting a wire file breaks CI until it is replaced
- the materialisation owner — deleting projection methods before that is named drops behaviour silently

## Reference

ADR-0099 detail 3. Sibling: the status-class issue.

### LINKED ISSUE #3488: Delete as_ParticipantStatus and as_CaseStatus, and the as_CaseParticipant cascade
Source: **ADR-0099** detail 3 (#2670).

## Scope

The two remaining paired classes, and the only two of the 27 whose fields differ by **name** rather than spelling:

| pair | wire has | core has |
|---|---|---|
| `as_CaseStatus` / `CaseStatus` | `em_state`, `pxa_state` | `em`, `pxa` |
| `as_ParticipantStatus` / `ParticipantStatus` | `rm_state`, `vf_state`, `d_state`, `em_consent_state` | `rm`, `vf`, `d`, `consent` |

**That difference is already solved.** PR #3468 collapsed the dimension wrapper so a dimension serializes to its bare state value, and declared the aliases (`rm` → `rmState`, and so on). Core `ParticipantStatus` already produces AS2 output identical to `as_ParticipantStatus`. What remains is removing the now-redundant classes.

## Why this is its own issue, sized XL

Measured during the ADR-0099 spike. `as_ParticipantStatus` is referenced in **20 production files and 43 test files**. The structural blocker is:

```python
class as_CaseParticipant(...):
    participant_statuses: list[as_ParticipantStatus]
```

so deleting it forces `as_CaseParticipant` to change, which cascades into its activities, its factories, the FastAPI example routes (`examples.py` uses `as_ParticipantStatus` as a `response_model`), and two demo scripts (`status_updates_demo.py`, `manage_participants_demo.py`).

Also in scope: the two `_migrate_core_dimension_format` validators and `ParticipantStatus._migrate_flat_fields` become dead once there is no second shape to migrate between. They are three copies of one rule — PR #3468 made them agree; this issue deletes them.

## Acceptance Criteria

- [ ] AC-1: `as_ParticipantStatus` and `as_CaseStatus` are deleted, along with `as_ParticipantStatusRef` and `as_CaseStatusRef`.
- [ ] AC-2: `as_CaseParticipant.participant_statuses` names the core `ParticipantStatus`; its activities, factories, FastAPI example routes and the two demo scripts are updated.
- [ ] AC-3: Both `_migrate_core_dimension_format` validators and `ParticipantStatus._migrate_flat_fields` are deleted.
- [ ] AC-4: `test/core/models/test_dimension_bare_serialization.py` still passes, with the wire-comparison test replaced by a golden-payload assertion once there is no wire class to compare against.
- [ ] AC-5: The AS2 payload for a status is byte-identical to today's. This is the hard gate — the wire form must not change.
- [ ] AC-6: Full unit suite and integration suite pass, including the demo scenario integration jobs.
- [ ] AC-7: **Blocking gate — `CaseParticipant` must carry complete AS2 aliases before `as_CaseParticipant` is deleted.** Core `CaseParticipant` has no `alias_generator` and no explicit aliases on 10 fields, so its own `by_alias` dump is not valid AS2. Measured on `task/3437-add-object-to-case-to-wire`:

      ```
      wire-only keys (lost):   ['@context', 'acceptedEmbargoIds', 'attributedTo', 'caseRoles',
                                'embargoConsentState', 'mediaType', 'participantStatuses']
      core-only keys (leaked): ['accepted_embargo_ids', 'attributed_to', 'case_roles',
                                'embargo_consent_state', 'media_type', 'participant_statuses']
      ```

      `@context` is dropped entirely and six keys are misspelled. `CaseParticipant` is rendered through the wire-render port at `vultron/core/behaviors/case/ledger_snapshots.py:121`, so this output lands in `CaseLedgerEntry.payloadSnapshot` — which **CLP-07-001** requires to be receiver-reconstitutable AS2, in a tamper-evident ledger.

      **This fails silently, not loudly.** PR #3540 changed `AS2WireRenderAdapter.render()` (`vultron/adapters/driven/wire_render/as2.py:83`) so a missed `WIRE_TYPE_MAP` lookup falls back to `obj.model_dump(by_alias=True, ...)` instead of raising `VultronValidationError`. Its guard is `isinstance(obj, BaseModel)`, which asks "is this a Pydantic model", not "does this serialize to valid AS2" — every core class passes. Today `as_CaseParticipant` still exists so the lookup hits and the gap is masked; deleting it in this issue removes the mask. Re-measure before scoping (MS-16-001), and see the review finding on #3540.
- [ ] AC-8: A test asserts the rendered `CaseParticipant` payload carries `@context` and camelCase keys, driven through the wire-render port rather than by calling `model_dump` directly — so the port's fallback path is what is under test.

## Blocked by

- #3578 — the `ARCH-12-003` rewrite decides whether `CaseParticipant` gets a blanket `alias_generator` (the mechanism PR #3540 uses for `VulnerabilityCase`) or explicit per-field aliases. AC-7 cannot be implemented without that answer.
- the ratchet replacement
- the materialisation owner
- the 25-class deletion — do the easy ones first, so the cascade is the only thing left that can go wrong

## Reference

ADR-0099 detail 3 and detail 5. Serialization groundwork landed in PR #3468.

sei-ahouseholder:
- Closes #3487
- Closes #3488

## Summary

Implements ADR-0099 detail 3: deletes **all 27** paired `as_*` domain classes, each replaced by a `TypeAlias` (`as_Foo = CoreFoo`) so existing import sites stay green while the dual model is eliminated.

Scope grew from 25 to 27 during review, which is why this closes #3488 as well. #3487 held back `as_CaseStatus`, `as_ParticipantStatus` and `as_CaseParticipant` because they "differ by field *name*" rather than by spelling. That had already stopped being true: ADR-0099 detail 5 makes a dimension serialize to a bare state value, and the core fields carry the `emState`/`rmState` aliases, so both sides already emitted byte-identical AS2. Collapsing them removes the seam instead of building a translation step to bridge it — the cheap option, not the expensive one.

## What review found, and what it cost

The first push was green on lint and red on CI: **25 integration failures plus a test that hung until killed at 8m47s**. The cause was a partial union widening. `as_Activity.actor` was widened to accept core objects but `target`/`origin`/`instrument` were not, so a promoted class in those slots escaped its declared union — pydantic emitted `PydanticSerializationUnexpectedValue` and shipped a payload shaped by the wrong schema outbound, and refused a well-formed payload inbound. Every recipient answered HTTP 422 to `Create(VulnerabilityCase)` and never seeded the case replica (ADR-0041, PCR-01-003). The hang was the same defect: an unbounded wait when the inbox rejected the activity.

**15 `# type: ignore[assignment]` markers were suppressing exactly the error that flags this**, which is why mypy and pyright were clean while the protocol did not work. All 15 are gone; widening the base union makes the subclass overrides genuine narrowings.

Four more behaviours were being dropped silently, each caught by a test rather than by inspection:

| dropped | restored as |
|---|---|
| Vultron JSON-LD `@context` on every promoted type (VM-10-001, MUST) | a `by_alias`-aware serializer on `CoreObject` — emitted on delivery, withheld from persistence, which is ADR-0099 detail 1's table |
| `attributedTo` → `attributed_to` on four types | the alias generator inherited once at `CoreObject` instead of ten per-class opt-ins |
| `+00:00` → `Z` timestamps | pinned back; all 226 timestamps in the published reference use the offset form, and `payloadSnapshot` bytes are compared across replicas |
| `_reject_retired_vfd_keys` (SDO-03-005) | relocated to `ParticipantStatus` — the one wire behaviour with no core equivalent |

Also fixed: the render port fails closed again (ARCH-20-003) — it was gated on `isinstance(obj, BaseModel)`, which asks "is this a Pydantic model" rather than "does this serialize to valid AS2", and `VultronOfferRecord` slipped through emitting snake_case with no `@context`; materialisation places promoted core objects instead of refusing every well-formed read (VM-06-007); RM and PXA coercion raise `ValueError` not `KeyError`, so a bogus state is a clean 422 rather than a 500 (#2964); three more `isinstance(x, as_Base)` guards that silently changed meaning, two of which had stopped generating example files entirely.

## Changes

- **`vultron/wire/as2/vocab/base/objects/activities/base.py`**: widen `target`/`origin`/`instrument` to admit `CoreObject`. This is the 422 fix.
- **`vultron/core/models/base.py`**: `alias_generator=to_camel` inherited by every `CoreObject`; `@context` and `+00:00` timestamps emitted on the `by_alias` path only; `local_only_fields` for fields that must persist but not ship.
- **`vultron/core/models/case_status.py`, `participant_status.py`**: flat `em_state`/`pxa_state`/`rm_state`/`vf_state`/`d_state` as read/**write** views onto the dimensions. Read-only properties satisfied every reader and then failed on the first writer with "property has no setter", which is a narrower break rather than compatibility.
- **`vultron/wire/as2/vocab/objects/case_status.py`, `case_participant.py`**: the last three pairs collapsed (#3488).
- **`vultron/adapters/driven/wire_render/as2.py`**: gate the fallback on `CoreObject`, not `BaseModel`.
- **`vultron/wire/as2/rehydration.py`**: admit promoted core classes in object-only slots and as resolved references.
- **`vultron/core/models/dimensions.py`**: `_coerce_rm`/`_coerce_pxa`/`_coerce_pec` wrap the enum lookup, matching `_coerce_vf`/`_coerce_d`.
- **`vultron/errors.py`**: `VultronValidationError` inherits `ValueError` (ARCH-23-006's stated precondition); new `VultronAlreadyExistsError` so "already stored" stays separable from "could not be projected".
- **60 example files regenerated**, plus specs, notes, ADR-0099 and both AGENTS.md files.

## Spec and decision changes

Four requirements are annotated because merged code would otherwise contradict them:

- **ARCH-23-006** — its precondition is now met, and enforcement moved rather than ended. The old test forbade core types in wire annotations, which detail 3 makes the intended shape; it is replaced by the invariant detail 3 states (every promoted class is exactly AS2-representable). The as-delivered alternative was a 59-entry allowlist, which records violations without checking anything.
- **ARCH-12-003 / ARCH-20-001 / ARCH-20-006** — each forbids a mechanism ADR-0099 requires, on a rationale that rests on the adapter-side translator ADR-0099 deleted. The parts that still bind (one rendering seam; never synthesise AS2 by dumping an arbitrary core class) are called out as surviving. **#3578 owns the full rewrite**; these notes exist so the requirements are not left silently false.
- **ADR-0099 detail 2 amended**: the AS2 spelling is *derived* by a generator plus three explicit exceptions, not enumerated per field. `to_camel` is correct for 27 of 30 real field names; the three it misses already carry explicit aliases. Enumerating buys nothing and costs the fields nobody remembers — which is precisely how four types shipped `attributed_to`.

`_TO_CAMEL_BACKLOG_1991` is retired rather than grown; its direction is inverted by ADR-0099, and a set documented "may only SHRINK" cannot express "all of them, by inheritance". #2288/#2289 are superseded (see #3578).

## Verification

- **Unit suite: 9364 passed**, 452 skipped, 20 xfailed — exit 0
- **Integration suite: 1284 passed** — exit 0, including `test_transfer_ownership_demo`, which previously hung
- black, flake8, mypy, pyright all clean, with **zero** added suppressions
- **Zero keys lost across all 60 published example files** versus `main`, verified by key-set comparison. The only differences are additions: the auto-seeded opening `CaseStatus`, and `vfState` auto-set for the VENDOR role (ADR-0075)
- AC-3's missing control is now in place: `test_committed_examples_have_the_same_wire_keys_as_generated` compares the key structure of every committed example against regenerated output, and a companion test pins the timestamp spelling. Filename-set equality — all the previous check did — passes while `@context` disappears from every object


sei-ahouseholder (commented):
## PR Triage: #3540 — feat: delete 25 paired as_* domain classes (ADR-0099 detail 3)

**Linked issues**: #3487 (Delete the 25 straightforward paired as_* domain classes)
**Changed files**: 100 files (+1156 / −1681) — wire/as2, core/models, adapters/driven, adapters/driving, core/behaviors
**CI status**: ❌ failing — Tests (pytest) + all 8 Demo Integration / Invariant Harness jobs
**Merge state**: ✅ mergeable (UNSTABLE)
**Base branch**: `main`

### Local reproduction

| Selection | Result |
|---|---|
| unit (`pytest`, the default `-m 'not integration'`) | ✅ 9361 passed, 450 skipped, 21 xfailed — exit 0 |
| integration (`pytest -m integration`) | ❌ **25 failed, 3 errors**, 1254 passed |
| hang | `test/demo/test_transfer_ownership_demo.py::test_demo` blocks until timeout (same stack as CI) |
| `main` baseline (ff4b39cb) | ✅ Python application + Demo Integration both green |

CI's gate runs `uv run pytest -m ""` (all markers). The unit selection is green, so validating only the default selection hides all of this. Every failure below is a regression introduced by this branch.

---

### Root cause of the integration breakage

`as_Activity.actor` was widened to `as_ActorRef | CoreActor`, but **`target`, `origin` and `instrument` were left as `as_Object | as_Link | str | None`**. Core objects placed in those slots escape the declared union in both directions:

```
PydanticSerializationUnexpectedValue(Expected `as_Object` … [field_name='target',
  input_value=VultronOrganization(…), input_type=VultronOrganization])
ERROR: Terminal delivery failure (HTTP 422) for activity … — not retrying (OX-13-005).
AssertionError: no VulnerabilityCase replica for offer … — the CaseActor's
  Create(VulnerabilityCase) never arrived (ADR-0041, PCR-01-003)
```

The 14 `# type: ignore[assignment]` markers added to `target`/`origin` overrides suppress exactly the type error that flags this, which is why mypy and pyright stay green.

Two further independent rejections: the promoted core `VulnerabilityCase` cannot accept the **deliberately retained** wire `as_CaseParticipant` / `as_CaseStatus` that the parser produces for inline `{"type":"CaseParticipant"}` / `{"type":"CaseStatus"}` (`issubclass(as_CaseStatus, CaseStatus)` is `False`), so the bootstrap `Create(VulnerabilityCase)` that CBT-01-007/CBT-05-008 *require* to carry inline participants is refused with 422.

---

### Findings

|---|---|---|---|---|
| phase2-issue-blockers-unresolved-0 | issue-linkage | ❌ FAIL | #3487's own two Blocked-by prerequisites (ratchet replacement, materialisation owner) were never resolved | **inversion-halt** |
| phase5-arch23006-precondition-unmet-0 | spec-conformance | ❌ FAIL | ARCH-23-006 says "Fix that before inverting this rule"; `VultronValidationError` is still not a `ValueError` subclass (verified) | fix-now |
| phase8-target-origin-union-escape-0 | code-review | ❌ FAIL | `target`/`origin`/`instrument` not widened with `CoreObject` — breaks send *and* receive | fix-now |
| phase7-type-ignore-masks-union-escape-0 | agents-md | ❌ FAIL | 20 production `# type: ignore` added; 14 hide the union escape above | fix-now |
| phase8-inline-case-participants-rejected-0 | code-review | ❌ FAIL | Inline `caseParticipants` → 422; violates CBT-01-007/CBT-05-008 | fix-now |
| phase8-inline-case-statuses-rejected-0 | code-review | ❌ FAIL | Inline `caseStatuses` → 422; `as_CaseStatus` is not a `CaseStatus` | fix-now |
| phase5-vm10001-context-lost-0 | spec-conformance | ❌ FAIL | Vultron JSON-LD `@context` gone from every promoted type — VM-10-001 (MUST) | fix-now |
| phase8-partial-to-camel-snake-on-wire-0 | code-review | ❌ FAIL | Partial `to_camel` migration puts `attributed_to`, `case_roles`, … on the wire; changes `payloadSnapshot` bytes | fix-now |
| phase5-arch20001-alias-generator-forbidden-0 | spec-conformance | ❌ FAIL | ARCH-20-001 (MUST, unamended) forbids `alias_generator` on core classes | fix-now |
| phase5-arch20003-render-fails-open-0 | spec-conformance | ❌ FAIL | Render port guard relaxed to `BaseModel`; no longer fails closed | fix-now |
| phase5-arch20006-core-dumped-as-as2-0 | spec-conformance | ❌ FAIL | `responses.py` now dumps core classes as synthesised AS2 — ARCH-20-006 (MUST) | fix-now |
| phase8-vulnerability-record-not-registered-0 | code-review | ❌ FAIL | Missing `WIRE_TYPE_MAP` line; inline `VulnerabilityRecord` silently loses `aliases` | fix-now |
| phase8-case-actor-inbox-outbox-projection-lost-0 | code-review | ❌ FAIL | `as_CaseActor` projection dropped: no `inbox`, core `VultronOutbox` on the wire | fix-now |
| phase8-embargo-event-set-name-dropped-0 | code-review | ❌ FAIL | `set_name` derivation dropped; protocol-visible via `payloadSnapshot` | fix-now |
| phase8-ledger-entry-id-derivation-broken-0 | code-review | ❌ FAIL | `_set_id_from_case` reads `case_id` while the class emits `caseId`; SYNC dedup key wrong | fix-now |
| phase5-cbt01007-example-bare-uris-0 | spec-conformance | ❌ FAIL | Reference example changed to bare URIs, which CBT-01-007 says MUST NOT be used | fix-now |
| phase9-docs-examples-json-diverged-0 | notes-docs | ❌ FAIL | `docs/reference/examples/create_case.json` diverged from generator output | fix-now |
| phase9-wire-core-boundary-note-contradicted-0 | notes-docs | ❌ FAIL | Active note says #3440 removed `\| CoreObject` "to prevent re-introduction"; this re-introduces it | fix-now |
| phase10-tests-weakened-to-pass-0 | test-coverage | ❌ FAIL | Four tests rewritten to stop asserting what they guarded | fix-now |
| phase2-ac3-no-payload-equivalence-tests-0 | issue-linkage | ❌ FAIL | AC-3 unmet — no payload-equivalence test for any of the 25 pairs | fix-now |
| phase2-ac4-projection-work-undeclared-0 | issue-linkage | ❌ FAIL | AC-4 unmet — three confirmed silent projection drops, none declared | fix-now |
| phase11-ci-all-suite-and-integration-failing-0 | ci-status | ❌ FAIL | 25 failed / 3 errors locally; all 8 integration jobs red; `main` green | fix-now |
| phase11-demo-transfer-ownership-hang-0 | ci-status | ❌ FAIL | Demo hangs instead of failing — unbounded wait when the inbox rejects | fix-now |
| phase2-ac5-suites-not-passing-0 | issue-linkage | ❌ FAIL | AC-5 unmet | fix-now |
| phase3-false-verification-claims-0 | pr-body | ❌ FAIL | "Integration tests pass (exit 0)" is false | fix-now |
| phase8-trigger-target-id-divergence-0 | code-review | ⚠️ IMPROVE | `target=case.id_` re-expands via `rehydrate_fields`; two representations disagree | fix-now |
| phase8-hydration-silent-except-pass-0 | code-review | ⚠️ IMPROVE | Bare `except Exception: pass` turns unreadable into absent | fix-now |
| phase8-proposal-ledger-bypasses-datalayer-0 | code-review | ⚠️ IMPROVE | Inline branch commits a ledger entry replicas cannot resolve | fix-now |
| phase8-actors-wrong-notfound-error-0 | code-review | ⚠️ IMPROVE | Wrong-shape case reported as 404 not-found | fix-now |
| phase8-set-cs-context-invariant-unenforced-0 | code-review | ⚠️ IMPROVE | `set_cs_context` has no core counterpart | fix-now |
| phase3-ratchet-count-understated-0 | pr-body | ⚠️ IMPROVE | Allowlist is 59 entries, not the "29" described | fix-now |
| phase3-em-state-kwarg-claim-false-0 | pr-body | ⚠️ IMPROVE | `em_state`/`pxa_state` are read-only properties, not `__init__` kwargs | fix-now |
| phase3-commit-msg-lists-undeleted-classes-0 | pr-body | ⚠️ IMPROVE | Commit message claims status/participant classes were deleted; they are identical to `main` | fix-now |
| phase9-agents-md-as-prefix-stale-0 | notes-docs | ⚠️ IMPROVE | `AGENTS.md:71` still documents the dual-model `as_` rule | fix-now |

**Total**: 25 FAIL · 9 IMPROVE · 0 NEW-ISSUE

The scope boundary was respected — `as_CaseStatus`, `as_ParticipantStatus` and `as_CaseParticipant` are byte-identical to `main`, correctly matching #3487's exclusion. The problem is that promoting `VulnerabilityCase` while those three are retained leaves the two models unable to interoperate.

---

*Triage artifact: `.claude/pr-3540-triage.json`*
*Execution is paused on the `inversion-halt` gate — see the follow-up comment.*

sei-ahouseholder:
## Review finding: the new `render()` fallback makes a future #3488 defect silent

`vultron/adapters/driven/wire_render/as2.py` now falls back to a raw core dump when the `WIRE_TYPE_MAP` lookup misses:

```python
        # Core types that have no wire wrapper (ADR-0099 detail 3): render
        # the domain object directly.  Unknown non-Pydantic objects still raise.
        if not isinstance(obj, BaseModel):
            raise VultronValidationError(...)
        return obj.model_dump(by_alias=True, exclude_none=True, mode="json")
```

That is correct **only if the core class carries complete AS2 aliases.** One of the four types that reach `render()` does not.

Measured the types rendered at `vultron/core/behaviors/case/ledger_snapshots.py:66-143`:

| rendered type | `alias_generator` | fields dumping snake_case |
|---|---|---|
| `ParticipantStatus` | yes | 0 — safe |
| `CaseStatus` | yes | 0 — safe |
| `VulnerabilityCase` | **added by this PR** | 17 → 0, fixed here |
| **`CaseParticipant`** | **no, and this PR does not add it** | **10** |

`CaseParticipant` is rendered at `ledger_snapshots.py:121`. Diffing its core dump against the wire form:

```
wire-only keys (lost):   ['@context', 'acceptedEmbargoIds', 'attributedTo', 'caseRoles',
                          'embargoConsentState', 'mediaType', 'participantStatuses']
core-only keys (leaked): ['accepted_embargo_ids', 'attributed_to', 'case_roles',
                          'embargo_consent_state', 'media_type', 'participant_statuses']
```

`@context` is dropped entirely and six keys are misspelled. That output lands in `CaseLedgerEntry.payloadSnapshot`, which **CLP-07-001** requires to be receiver-reconstitutable AS2, in a tamper-evident ledger.

### Why the suite is green

`as_CaseParticipant` still exists, so the lookup hits and takes the `from_core` path. It is not among this PR's 25 classes — it is in the **#3488** cascade. So the sequence is:

1. this PR installs the silent fallback,
2. #3488 deletes `as_CaseParticipant`,
3. ledger snapshots silently emit invalid AS2 — no exception, no failing test.

Before this PR the same situation raised `VultronValidationError`. The fallback is what converts it from loud to silent.

### Suggested change

Either keep raising for classes whose AS2 completeness is not established, or gate the fallback on a positive check rather than `isinstance(obj, BaseModel)` — the current predicate asks "is this a Pydantic model", not "does this serialize to valid AS2", and every core class passes it.

Tracked on #3488 as a blocking AC either way. Also note the lookup one line above (`WIRE_TYPE_MAP.get(type(obj).__name__)`) is the ARCH-23-001 name-coincidence resolution reported in #3571 — this PR does not change it, but it does make its failures invisible.

### Separately: the `_TO_CAMEL_BACKLOG_1991` growth

`test/architecture/test_hierarchy_invariants.py` grows the backlog from 8 entries to 12, against the set's own comment three lines above (*"This set may only SHRINK"*), while the `xfail` reason still reads *"Goal state tracked in #2288 and #2289"* — two issues that ADR-0099 supersedes. That is now owned by **#3578**, which annotates `ARCH-12-003` (a live MUST that still forbids `alias_generator` on core types, with no ADR-0099 `note:`) and deletes the ratchet outright. #2288 and #2289 are closed as superseded. No change needed in this PR beyond pointing the reason string at #3578.

sei-ahouseholder:
## Non-blocking review — three follow-ups filed, plus one merge-order note

To be clear up front: **this is not a request to change anything before merging.** This PR has been in flight long enough that gating it on further cleanup is the wrong trade. Everything below is either filed as follow-up work or is a mechanical rebase note. Treating the PR as landing as-is.

Context for why these are being raised now: #2692 (authority ADR-0017) and #2670 (ADR-0099) were in an unrecognised race, and this PR is where the two theses actually collide. The audit turned up a repeating shape rather than three unrelated nits — in each case a guard stayed green while the guarantee it existed to protect left.

### 1. `_TO_CAMEL_BACKLOG_1991` widens 8 → 12 — tracked in #3578

`test/architecture/test_hierarchy_invariants.py:59` carries the comment "This set may only SHRINK." This PR grows it by four (`CaseLedgerEntry`, `EmbargoEvent`, `EmbargoPolicy`, `VulnerabilityCase`) and updates the reason string from "8 known classes remain" to "12 known classes remain".

The substantive point is not the ratchet convention — it is that `ARCH-12-003` is a **live MUST** with no ADR-0099 `note:` on it:

> Core-branch types MUST NOT carry wire-specific concerns: `alias_generator=to_camel`, hardcoded AS2 `@context` namespace, or VOCABULARY registration are forbidden in core-branch types.

ADR-0099 never mentions `alias_generator` anywhere — which is precisely why this could widen without anyone noticing a conflict. The backlog entries also still say "Goal state tracked in #2288 and #2289", which are issues ADR-0099 supersedes; that is circular ownership. #3578 resolves the spec question and re-homes the backlog; #2288/#2289 are being closed.

### 2. `render()` swaps a raise for a silent fallback — tracked as AC-7 on #3488

`vultron/adapters/driven/wire_render/as2.py:83`. The previous behaviour raised `VultronValidationError` when no wire counterpart existed; the new behaviour falls through to `obj.model_dump(by_alias=True, ...)` for any `BaseModel`. That is reasonable for the paired types whose wrapper this PR deletes, but it also silently absorbs the genuinely-missing-counterpart case the raise was written to catch, and the four call sites in `vultron/core/behaviors/case/ledger_snapshots.py:66-143` cannot distinguish the two outcomes.

### 3. Frozen received evidence — filed as #3584, and the design question is now settled

Measured on this branch with a parsed inbound `Add` carrying a nested `VulnerabilityCase`:

```text
activity      : as_Add             frozen= True
nested object : VulnerabilityCase  frozen= None
mutate activity field : blocked -> ValidationError
mutate nested object  : SUCCEEDED -> act.object_.name = TAMPERED EVIDENCE
```

On `main`-descended branches the nested mutation raises. `test/architecture/test_wire_artifact_immutability.py:30` asserts `frozen` on the abstract `as_Object` root only, so once `as_VulnerabilityCase` is a core class it falls outside the assertion's scope and the file still reports **2 passed**.

Calibrating this honestly, because it is easy to overread: the primary integrity mechanism is `payload_snapshot: dict[str, Any]`, a value copy, and this PR does not touch it. **No current code mutates received evidence** — it is presently impossible, so by construction none does. What is lost is enforcement over the parse→snapshot window. Latent regression, not live corruption.

The resolution is recorded in #3584: **snapshot-at-parse.** `model_config` is per-class and Pydantic has no per-instance freezing, so under one object model "frozen artifact plus copy-before-modify" cannot be a class config — the snapshot simply moves earlier, to the moment of parse, where the raw body is already available via `_reparse_as_specific_type(nested, raw_obj)`. That is follow-up work sequenced after this PR, not a precondition for it.

### 4. Merge-order note (mechanical)

This branch is **51 commits behind** `main` and `git merge-tree` predicts **10 conflicts**, including a `modify/delete` on `test/architecture/test_normalize_wire_to_core_ratchet.py` — `main` has already deleted that file, so the resolution is to accept the deletion rather than restore it.

Worth knowing about separately: **six production files auto-merge silently** between this branch and #3575, including `vultron/wire/as2/parser.py` and `_builders.py`. Both PRs touch parser admission logic — this one adds `or obj_type in WIRE_TYPE_MAP` to the filter at `parser.py:84`, #3575 changes timestamp-absence handling — and the *combination* is exercised by neither PR's tests. A clean textual merge is not evidence of behavioural compatibility here. Tracked separately so it does not become this PR's problem; #3575 is much closer to `main` (1 conflict) and is the natural one to land first.

sei-ahouseholder (commented):
## PR Execute: #3540 — feat: delete 25 paired as_* domain classes (ADR-0099 detail 3)

**Fixes applied**: 3 commits (`9b28492ec`, `33eece131`, `dc9d0629d`)
**Excursions filed and fixed**: 0
**Deferred (you approved)**: 1
**Halted (inversion, awaiting you)**: 0
**Tests run**: unit + integration
**CI status**: ✅ passing
**Base sync**: ✅ already current with `main` @ `f56112ca6`

---

### Fixed

| Finding | Commit | Closes |
|---|---|---|
| phase8-case-actor-inbox-outbox-projection-lost-0: CaseActor emits `inbox`/`outbox` collection URIs again (`_as2_derived_fields` hook) | `33eece131` | — |
| phase8-embargo-event-set-name-dropped-0: EmbargoEvent derives its display name | `33eece131` | — |
| phase8-set-cs-context-invariant-unenforced-0: inline statuses point at their case | `33eece131` | — |
| phase8-ledger-entry-id-derivation-broken-0: ledger id derived from `caseId`/`logIndex` too | `33eece131` | — |
| phase8-proposal-ledger-bypasses-datalayer-0: reports resolved from the store | `33eece131` | — |
| phase8-actors-wrong-notfound-error-0: wrong-shape case → `VultronValidationError` | `33eece131` | — |
| phase5-arch20006-core-dumped-as-as2-0: `AS2JSONResponse` refuses non-AS2, non-CoreObject models | `33eece131` | — |
| phase10-tests-weakened-to-pass-0: `@context`, collection URIs, camelCase keys asserted | `33eece131` | — |
| phase2-ac3-no-payload-equivalence-tests-0: examples key-shape + `+00:00` ratchets (flag 14 files vs `main`) | `33eece131` | — |
| phase9-wire-core-boundary-note-contradicted-0, phase9-agents-md-as-prefix-stale-0: notes and AGENTS.md corrected | `33eece131` | — |
| phase2-ac4-projection-work-undeclared-0, phase3-false-verification-claims-0: PR body declares all relocated behaviours; verification rewritten | `33eece131` | — |
| phase11-ci-all-suite-and-integration-failing-0, phase2-ac5-suites-not-passing-0: pyright/mypy CI errors; suites green | `dc9d0629d` | — |
| 20 remaining findings (union escape, type-ignores, `@context`, to_camel, render fail-closed, inline participants/statuses, examples, hydration, etc.) | through `9b28492ec` | — |

Those 20 were resolved on the branch before this run. Three verification agents re-checked them against current `main`. ARCH-20-003's statement is amended to "raise when not a `CoreObject`", which matches the code.

### Deferred — You Approved

| Finding | Issue | Measured remainder |
|---|---|---|
| phase2-issue-blockers-unresolved-0 | #3578 | You chose to land as-is. The rewrite of ARCH-12-003 / ARCH-20-001 / ARCH-20-006 remains (annotated here, not rewritten). #3584 is a further follow-up. |

---

*Execute artifact: `.claude/pr-3540-execute.json`*
*Next step: `/pr-ship` continues with `/pr-verify`.*

sei-ahouseholder (commented):
## PR Verify: #3540 — feat: delete 25 paired as_* domain classes (ADR-0099 detail 3)

**Overall verdict**: ✅ READY-TO-MERGE
**CI status**: ✅ passing (all checks on `dc9d0629d`, including Tests (pytest) and 4 demo integration + invariant harnesses)
**Base sync in execute**: ✅ already current (`main` @ `f56112ca6`)
**Integrity check**: ✅ all 34 findings accounted for

---

### Finding Verdicts

|---|---|---|---|
| phase2-issue-blockers-unresolved-0 | ❌ FAIL | deferred-ask → #3578 (open) | 📋 NOTED |
| phase5-arch23006-precondition-unmet-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase8-target-origin-union-escape-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase7-type-ignore-masks-union-escape-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase8-inline-case-participants-rejected-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase8-inline-case-statuses-rejected-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase5-vm10001-context-lost-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase8-partial-to-camel-snake-on-wire-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase5-arch20001-alias-generator-forbidden-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase5-arch20003-render-fails-open-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase5-arch20006-core-dumped-as-as2-0 | ❌ FAIL | fixed @ `33eece131` | ✅ CONFIRMED |
| phase8-vulnerability-record-not-registered-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase8-case-actor-inbox-outbox-projection-lost-0 | ❌ FAIL | fixed @ `33eece131` | ✅ CONFIRMED |
| phase8-embargo-event-set-name-dropped-0 | ❌ FAIL | fixed @ `33eece131` | ✅ CONFIRMED |
| phase8-ledger-entry-id-derivation-broken-0 | ❌ FAIL | fixed @ `33eece131` | ✅ CONFIRMED |
| phase5-cbt01007-example-bare-uris-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase9-docs-examples-json-diverged-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase9-wire-core-boundary-note-contradicted-0 | ❌ FAIL | fixed @ `33eece131` | ✅ CONFIRMED |
| phase10-tests-weakened-to-pass-0 | ❌ FAIL | fixed @ `33eece131` | ✅ CONFIRMED |
| phase2-ac3-no-payload-equivalence-tests-0 | ❌ FAIL | fixed @ `33eece131` | ✅ CONFIRMED |
| phase2-ac4-projection-work-undeclared-0 | ❌ FAIL | fixed @ `33eece131` | ✅ CONFIRMED |
| phase11-ci-all-suite-and-integration-failing-0 | ❌ FAIL | fixed @ `dc9d0629d` | ✅ CONFIRMED |
| phase11-demo-transfer-ownership-hang-0 | ❌ FAIL | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase2-ac5-suites-not-passing-0 | ❌ FAIL | fixed @ `dc9d0629d` | ✅ CONFIRMED |
| phase3-false-verification-claims-0 | ❌ FAIL | fixed @ `33eece131` | ✅ CONFIRMED |
| phase8-trigger-target-id-divergence-0 | ⚠️ IMPROVE | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase8-hydration-silent-except-pass-0 | ⚠️ IMPROVE | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase8-proposal-ledger-bypasses-datalayer-0 | ⚠️ IMPROVE | fixed @ `33eece131` | ✅ CONFIRMED |
| phase8-actors-wrong-notfound-error-0 | ⚠️ IMPROVE | fixed @ `33eece131` | ✅ CONFIRMED |
| phase8-set-cs-context-invariant-unenforced-0 | ⚠️ IMPROVE | fixed @ `33eece131` | ✅ CONFIRMED |
| phase3-ratchet-count-understated-0 | ⚠️ IMPROVE | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase3-em-state-kwarg-claim-false-0 | ⚠️ IMPROVE | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase3-commit-msg-lists-undeleted-classes-0 | ⚠️ IMPROVE | fixed @ `9b28492ec` | ✅ CONFIRMED |
| phase9-agents-md-as-prefix-stale-0 | ⚠️ IMPROVE | fixed @ `33eece131` | ✅ CONFIRMED |

---

### Deferred Items — Maintainer Approved

- **phase2-issue-blockers-unresolved-0** → #3578 (open). You chose to land as-is. The rewrite of ARCH-12-003 / ARCH-20-001 / ARCH-20-006 for one object model is tracked there; #3584 is a further follow-up.

*Each fix was checked in the file at HEAD, not only in its commit diff.*

Issue #2992, opened 2026-09-01
fix(docs): AGENTS.md line 302 incorrectly describes WIRE_TYPE_MAP key naming


sei-ahouseholder:
## Problem

`AGENTS.md:302` states:

> `WIRE_TYPE_MAP` uses wire `type_` value keys (`"VulnerabilityCase"`)

This is incorrect for actor classes. The actual registration in `as_Base.__init_subclass__` is:

```python
WIRE_TYPE_MAP[cls.__name__.removeprefix("as_")] = cls
```

For most classes the key happens to equal the `type_` value (e.g. `as_VulnerabilityCase` → key `"VulnerabilityCase"`, `type_` = `"VulnerabilityCase"`). But for actor classes it diverges:

- `as_VultronPerson` → WIRE_TYPE_MAP key is `"VultronPerson"`, while `type_` is `"Person"`
- `as_VultronOrganization` → key is `"VultronOrganization"`, while `type_` is `"Organization"`

The explicit AS2 `type_` key overrides for actor classes are added separately in `vultron_actor.py` lines 103–107, not by the generic hook.

## Fix

Update AGENTS.md line 302 to say the key is `cls.__name__.removeprefix("as_")` (not the `type_` field value), and note the actor-class override in `vultron_actor.py`.

## Source

Pre-existing doc error discovered during code review of PR #2991.
Learning file: `plan/incoming/learnings/20260901-wire-type-map-key-naming-agents-md-wrong.md`

sei-ahouseholder:
Claimed on branch `bug/2553-wire-timestamps-carried-not-minted`. Bundled with #2553 in one PR.

sei-ahouseholder:
**History: implementation — WIRE_TYPE_MAP key naming documented correctly**

Issue #2992. Symptom: notes/vocabulary-registry.md, notes/README.md and registry.py said WIRE_TYPE_MAP is keyed by the wire type_ value. Root cause: as_Base.__init_subclass__ keys by cls.__name__.removeprefix('as_'); it diverges for as_VultronPerson (Person), the other Vultron actors and as_VulnerabilityCaseStub; the AS2 actor type_ keys are registered explicitly in vultron_actor.py/actors.py. Fix: corrected the three docs and added a pinning test. PR: https://github.com/CERTCC/Vultron/pull/3575

sei-ahouseholder:
[...]

Issue #2979, opened 2026-09-01
Fix current_status UUID tiebreaker in VulnerabilityCase


sei-ahouseholder:
## Summary
`VulnerabilityCase.current_status` uses `max(key=lambda cs: cs.updated or cs.published or cs.id_)` to determine the most-recent CaseStatus. The `cs.id_` fallback sorts auto-seeded UUID-based IDs (`urn:uuid:…`, starting with `'u'`) lexically higher than HTTPS-scheme IDs (`https://…`, starting with `'h'`), so an auto-seeded initial status beats a legitimately newer received status when both lack explicit timestamps. The fix: replace the `cs.id_` fallback with `datetime.min.replace(tzinfo=timezone.utc)` so timestampless statuses sort to the bottom.

## Acceptance Criteria
- [ ] AC-1: Replace `cs.id_` fallback with `datetime.min.replace(tzinfo=timezone.utc)` in `vultron/core/models/case.py` `VulnerabilityCase.current_status`
- [ ] AC-2: Apply the same fix to `vultron/wire/as2/vocab/objects/vulnerability_case.py`
- [ ] AC-3: Unit test: construct a VulnerabilityCase with an auto-seeded status (urn:uuid ID, no timestamps) and a received status (HTTPS ID, no timestamps); assert `current_status` returns the received status

## Reference
Source: #2832
Spec: `specs/case-management.yaml` CM-29-001
Docs PR: https://github.com/CERTCC/Vultron/pull/2978

sei-ahouseholder:
Claimed on branch `task/2979-fix-current-status-uuid-tiebreaker`.

sei-ahouseholder:
**History: implementation — Fix current_status UUID tiebreaker in VulnerabilityCase**


## Issue #2979 — Fix current_status UUID tiebreaker in VulnerabilityCase

Replaced the `cs.id_` recency fallback in `VulnerabilityCase.current_status`
with a shared `status_recency_key(updated, published)` helper in
`vultron/core/models/_helpers.py`, used by both the core and wire
implementations (CM-29-001). The key resolves `updated` then `published`,
floors timestampless statuses at `datetime.min` (UTC), and normalizes naive
wire timestamps to UTC so mixed naive/aware comparisons never raise
`TypeError` in `max()`.

Pre-PR code review surfaced three findings (naive-datetime crash, core/wire
duplication, per-element sentinel recompute); all resolved in-session via the
[...]

Issue #3571, opened 2026-09-23
concern: As2WireRenderAdapter resolves a core class name against the wire type-value index (ARCH-23-001)


sei-ahouseholder:
## Summary

`As2WireRenderAdapter.render()` resolves a core class to its wire counterpart with

```python
wire_cls = WIRE_TYPE_MAP.get(type(obj).__name__)
```

That is a **core class name** looked up in an index keyed by **wire `type` values**
(VM-01-007). It resolves only where the two strings happen to coincide, which is
the name-coincidence resolution ARCH-23-001 explicitly forbids:

> Resolving a counterpart by name coincidence between `VOCABULARY` and
> `CORE_VOCABULARY` is forbidden.

`notes/vocabulary-registry.md` documented this exact lookup as the sanctioned way
to resolve a counterpart, which is how it survived.

## Why it matters

For every core type whose name differs from its wire `type` value, the lookup
misses and `render()` raises `VultronValidationError("No wire counterpart for core
type 'X'.")` — a message that reads as "this type is not registered" when the real
cause is that the index cannot answer this question at all. The five core actor
types (`VultronPerson`, `VultronOrganization`, `VultronService`,
`VultronApplication`, `VultronGroup`) are in that position today.

Measured while fixing #2982: they raised *before* that fix too, because
`as_VultronPerson` is not an `as_VultronObject` and has no `from_core()`, so the
`issubclass` guard rejected it even when `WIRE_TYPE_MAP["VultronPerson"]` existed.
So removing the class-name keys changed nothing here — but it removed the last
thing that made the lookup look like it was working.

## Why it is not fixed in #3569

The correct fix is an explicit core↔wire pairing, and the mechanism ARCH-23-001
specified for that (the declarative pairing registry, ADR-0082 / #2937) is
**cancelled by ADR-0099**, which deletes the paired `as_*` domain classes instead.
Under one object model there is no counterpart to resolve and this lookup
disappears entirely.

So the real disposition question is whether ADR-0099 (currently *provisional*)
lands. If it does, this resolves itself. If it stalls, this lookup needs an
explicit index rather than a string coincidence, and ARCH-20-003's error message
should distinguish "no counterpart declared" from "cannot be resolved this way".

## Pointers

- `vultron/adapters/driven/wire_render/as2.py` (the lookup and its docstring)
- `specs/architecture.yaml` ARCH-23-001, ARCH-20-003
- `specs/vocabulary-model.yaml` VM-01-004, VM-01-007
[...]

Issue #2982, opened 2026-09-01
fix: WIRE_TYPE_MAP actor keys diverge from wire type_ values (VultronPerson vs Person)


sei-ahouseholder:
## Summary

`as_VultronPerson` and its 4 sibling classes register in `WIRE_TYPE_MAP` under **two** keys each:

1. `WIRE_TYPE_MAP["VultronPerson"]` — auto-registered by `__init_subclass__` (class name stripped of `as_`)
2. `WIRE_TYPE_MAP["Person"]` — manually added in `vultron_actor.py`

Only key 2 is the correct wire `type_` value; key 1 is stale. The stale keys exist because the class name (`as_VultronPerson` → `VultronPerson`) differs from the wire `type_` value (`Person`).

## Why it matters

The ratchet test `test_every_normalized_type_actually_shadows_a_core_type` relies on the stale `"VultronPerson"` etc. keys to detect that these wire classes shadow core classes. Attempting to remove the stale keys (by using the Literal type arg instead of the stripped class name in `__init_subclass__`) breaks the ratchet test and reveals that `_normalize_wire_to_core()` in `db_record.py` never actually fires for actor types, because `obj.type_` is `"Person"` (not `"VultronPerson"`) and `"Person"` is not in `_NORMALIZE_WIRE_TO_CORE`.

## Root cause

1. `__init_subclass__` uses stripped class name for the WIRE_TYPE_MAP key, which diverges from the actual `type_` value for `as_VultronXxx` classes.
2. `_normalize_wire_to_core()` in `db_record.py` checks `obj.type_ in _NORMALIZE_WIRE_TO_CORE`, but `_NORMALIZE_WIRE_TO_CORE` uses core class names (`VultronPerson`) rather than wire type_ values (`Person`), so normalisation silently skips actor objects.

## Desired end state

- `WIRE_TYPE_MAP` keyed exclusively by wire `type_` values (no stale class-name keys)
- `_normalize_wire_to_core()` correctly normalises `as_VultronPerson` → `VultronPerson`
- Ratchet test updated to detect actor shadowing via a method other than WIRE_TYPE_MAP key collision (e.g. by walking `to_core()` return type)

## Related

- ARCH-23-002 (VOCABULARY/CORE_VOCABULARY keys disjoint)
- `vultron/wire/as2/vocab/base/base.py` `__init_subclass__` (line 54)
- `vultron/adapters/driven/db_record.py` `_normalize_wire_to_core()` (line 298)
[...]

PR #3575, opened 2026-09-23
fix: carry object timestamps as received, never minted by the receiver

### LINKED ISSUE #2553: Wire render mints published/updated instead of carrying them, so a rendered object is not reproducible
## Problem

`as_Base` declares both timestamp fields with a factory, not a carried value:

```python
# vultron/wire/as2/vocab/base/objects/base.py:70
published: datetime | None = Field(default_factory=now_utc, ...)
updated:   datetime | None = Field(default_factory=now_utc, ...)
```

Core status objects (`ParticipantStatus`, `CaseStatus` — both `CoreObject`) hold
no timestamp of their own, so there is nothing for the wire render to carry.
The consequence is that **rendering one stored object twice yields two different
objects**: `published` and `updated` are stamped at render time, not read from
the thing being rendered. `now_utc()` truncates to whole seconds
(`vultron/wire/as2/vocab/base/dt_utils.py:41`), so the value changes once per
second.

This makes a "snapshot" not a snapshot. Two observable consequences:

1. **Ledger dedup was a race against the clock.** `_find_equivalent_recorded_entry`
   (`vultron/core/sync_helpers.py`) treats a retry as the same assertion when its
   `payload_snapshot` matches an already-recorded one. Because
   `build_add_participant_status_snapshot` / `build_add_case_status_snapshot`
   embed a freshly *re-rendered* object (and, in `target`, a whole re-rendered
   participant whose nested `participantStatuses` each carry their own pair), a
   retry one second later differed in nothing but timestamps — so the dedup
   missed and appended duplicate ledger indices, which is exactly the
   ledger-index instability ADR-0041 forbids.

   This was found as a flaky CI failure on PR #2484
   (`TestADR0041Idempotency::test_duplicate_proposal_no_duplicate_participants_or_entries`,
   `before=8, after=9`): green locally, red on slower CI, purely on whether the
   two deliveries landed in the same clock second. **Already mitigated** in that
   PR by comparing snapshots with build-time timestamps normalised out, plus a
   regression test that advances the clock deterministically. That fixes the
   dedup; it does not fix the cause below.

2. **Timestamp-ordered reads are meaningless.** `as_VulnerabilityCase` sorts case
   statuses by `cs.updated or cs.published or cs.id_`
   (`vultron/wire/as2/vocab/objects/vulnerability_case.py:284`). When those values
   are minted at render time they order by *when the list happened to be
   rendered*, not by when the statuses were recorded. A closely related bug was
   already worked around once for exactly this reason — see the `[-1]` comment on
   `as_CaseParticipant.participant_status` (issue #659), which abandoned
   timestamp ordering because "wire-layer timestamps on appended statuses ... may
   be *earlier* than locally-constructed defaults".

Anything else that compares, hashes, or orders rendered objects inherits the same
nondeterminism — including cross-replica comparison of ledger content, where two
actors recording the same event necessarily produce different snapshots.

## Why this is a separate issue

Surfaced while root-causing a blocking CI failure on PR #2484 (ADR-0072, per-actor
DataLayer isolation). The choke-point fix landed there because the failure blocked
the PR; the cause is a wire/core model question — *should a core status object
carry its own creation and modification time, and should the render carry it
rather than mint it?* — with a schema change and a stored-data implication. That
is not the per-actor-storage PR's family.

## Acceptance criteria

- [ ] Decide and record where object time lives: core models carrying
      `published`/`updated` (or equivalents) that the render carries through, vs.
      the render continuing to mint them and every consumer treating them as
      non-semantic.
- [ ] Rendering the same stored object twice produces equal dicts — asserted by a
      test that advances the clock between the two renders.
- [ ] `as_VulnerabilityCase.case_status` ordering is either backed by carried
      timestamps or changed to an ordering that does not depend on them (matching
      what #659 did for `participant_status`).
- [ ] With the cause fixed, revisit whether the `_semantic_payload` normalisation
      in `vultron/core/sync_helpers.py` is still needed, or can be narrowed/removed.
- [ ] Existing stored objects without carried timestamps continue to load
      (rehydration must not fail on absent values).

### LINKED ISSUE #3257: wire: a nested object with no claimed published time gets the receiver's clock as its published
## Summary

A nested (inline) object that carries no claimed `published` time ends up with the
**receiver's wall clock** as its `published`, presented downstream as the sender's
claim. This contradicts the rule that a snapshot carries that actor's claimed time
via `claimed_published_iso()`, never `now_utc()` (see root `AGENTS.md`, case-ledger
row, and `notes/case-ledger-authority.md`).

The top-level guard in `parse_activity` refuses an activity whose own `published` is
absent or blank (CLP-15-006, MV-03-002). That guard is deliberately **not** applied to
nested objects, which legitimately omit `published`. Nothing replaces it, so the
fabrication happens silently one level down.

## Two independent fabrication paths

1. **`default_factory=now_utc`** on `as_Object.published` / `as_Object.updated`
   (`vultron/wire/as2/vocab/base/objects/base.py`). An *omitted* nested `published`
   is filled with the receiver's clock at construction. After `model_validate`, a
   fabricated value is indistinguishable from a sender-supplied one — the exact
   reason the top-level guard runs on the raw body instead.

2. **`_get_timestamp`'s fallback** in `vultron/wire/as2/extractor/_builders.py`:

   ```python
   val = getattr(obj, field, None)
   return val if isinstance(val, datetime) else _core_now_utc()
   ```

   Called on nested objects in `_build_report_object` and `_build_case_object`. Any
   `None` becomes the receiver's clock.

## Measured behavior

`Create` carrying an inline `VulnerabilityReport`, varying only its `published`
(`model` = value on the parsed wire object, `extracted` = value on the built domain
object):

| nested `published` | model | extracted |
|---|---|---|
| omitted | `2026-09-15T15:17:08Z` (receiver clock) | same |
| `""` (blank) | `None` | `2026-09-15T15:17:08Z` (receiver clock) |
| `null` | `None` | `2026-09-15T15:17:08Z` (receiver clock) |
| real value | preserved | preserved |

Every no-claimed-time spelling yields a receiver clock at the extractor boundary.

## Not introduced by #3233

Path 1 is present on `main` and predates #3233. PR #3233 (ADR-0090) makes a blank
nested timestamp read as `None`, which routes blank to the same extracted outcome as
omitted by path 2 — the intended "blank is absence" semantics, not a regression.
Verify against `main` before scoping a fix.

## Why it matters

`CaseLedgerEntry.payloadSnapshot` comparisons (CLP-14-006/007/008) assume `published`
is the asserting participant's claim. A fabricated value leaves those checks comparing
the receiver's clock against itself, and puts a foreign clock into an actor's claimed
stream. `notes/case-ledger-authority.md` names this trap explicitly for the top-level
case; the nested case has the same shape and no guard.

## Suggested direction

Decide what a nested object with no claimed time should mean, then make one thing true
everywhere:

- Distinguish "sender supplied no time" from "we filled one in" — most likely by
  dropping `default_factory=now_utc` on the **inbound** path so absence stays `None`,
  and having `_get_timestamp` return `datetime | None` so callers must handle it.
- The outbound path legitimately needs the default (the same classes author outbound
  activities), so the two directions probably need separating rather than one flag.

Filed per BW-07-009 — surfaced during PR triage of #3233 in files that PR did not own.

### LINKED ISSUE #3592: test: the WIRE_TYPE_MAP collision ratchet cannot see a test-module squatter
## Summary

On `origin/main` at `b90cbc1f`, a test-local subclass silently takes over a production `WIRE_TYPE_MAP` key for the rest of the pytest process, and **66 tests fail in a full unit run**. Reproduced on a pristine `origin/main` worktree with no PR code:

```
$ uv run pytest test/adapters/driven/trigger_activity_adapter/test_cases.py \
                test/wire/as2/vocab/base/test_registry.py \
                test/wire/as2/test_rehydration_materialisation.py
FAILED test/adapters/driven/trigger_activity_adapter/test_cases.py::TestToWireObject::test_core_object_converted_via_from_core
FAILED test/wire/as2/vocab/base/test_registry.py::TestDynamicDiscovery::test_vocab_bug_26040902_regression
FAILED test/wire/as2/vocab/base/test_registry.py::TestDynamicDiscovery::test_case_stub_does_not_claim_the_case_wire_type_key
FAILED test/wire/as2/vocab/base/test_registry.py::TestWireTypeValues::test_critical_type_values_resolve_to_expected_wire_class[VulnerabilityCase-as_VulnerabilityCase]
FAILED test/wire/as2/test_rehydration_materialisation.py::test_rehydrate_materialises_object_slots
FAILED test/wire/as2/test_rehydration_materialisation.py::test_rehydrate_materialises_from_a_bare_case_id
FAILED test/wire/as2/test_rehydration_materialisation.py::test_rehydrate_refuses_an_unresolvable_object_slot
```

The assertion names the hijacker directly:

```
assert WIRE_TYPE_MAP["VulnerabilityCase"] is as_VulnerabilityCase
E  AssertionError: assert <class 'test.adapters.driven.trigger_activity_adapter.test_cases.
   TestAddObjectToCaseConversionBranch.test_vultron_error_from_from_core_is_not_rewrapped.
   <locals>._NotFoundWireClass'> is <class 'vultron.wire.as2.vocab.objects.vulnerability_case.as_VulnerabilityCase'>
```

Each failure is order-dependent: every one of them passes in isolation. `test/adapters/` sorts before `test/architecture/`, `test/core/` and `test/wire/`, so the polluter always runs first in a full run and every later consumer of the key sees `_NotFoundWireClass`.

## Root cause

VM-01-008 (#2982) made `as_Base.__init_subclass__` derive the `WIRE_TYPE_MAP` key from the **emitted `type` value** rather than the class name:

```python
if not is_wire_type_alias(cls):
    WIRE_TYPE_MAP[wire_type_value(cls)] = cls
```

That is the right key form, but it changed the blast radius of defining a throwaway wire subclass in a test. `test_cases.py:263` defines:

```python
class _NotFoundWireClass(as_VultronObject):
    type_: str = "VulnerabilityCase"
```

Before #2982 the derived key was `cls.__name__.removeprefix("as_")` → `"_NotFoundWireClass"`, which collided with nothing. After #2982 it is `"VulnerabilityCase"` — the real key — and the assignment is an unconditional overwrite. `WIRE_TYPE_MAP` then holds a strong reference, so the test class outlives the function that defined it and owns the key for the whole session.

This is a known hazard in the same file: `_BrokenWireClass` at `test_cases.py:224`, two tests earlier, already declares `_wire_type_alias: ClassVar[bool] = True` for exactly this reason. `_NotFoundWireClass` was simply missed.

**Why no ratchet caught it.** `test_no_unsanctioned_wire_type_collisions` is the ratchet for precisely this failure mode, and its docstring describes it exactly ("a new class that declares an existing value silently takes the key for the whole process, decided by import order"). It cannot see this one, because `_registrable_wire_classes()` scopes the walk to production modules:

```python
if sub.__module__.startswith("vultron.wire.as2.vocab") and ...
```

A `test.*` class is filtered out, so the collision is invisible to the one check written to find it.

**Why CI is green.** There is no test workflow on `main` pushes — tests run on `pull_request` only, against the PR's merge commit. `main`'s own tree is never exercised as a unit, so this landed green and stays green until a PR merges main and runs the full suite.

## Fix

Two parts, both small:

1. `_NotFoundWireClass` declares `_wire_type_alias: ClassVar[bool] = True`, matching `_BrokenWireClass`. This is not a workaround — the flag means "shares another class's wire `type` value and is not what that value should deserialize to", which is exactly true of it.
2. Extend the ratchet so the next one is caught rather than rediscovered: assert that every class owning a `WIRE_TYPE_MAP` key is defined in `vultron.*`. A `test.*` class holding a production key is always a bug, and this check sees what the subclass walk structurally cannot.

The runtime hardening option — having `__init_subclass__` refuse to overwrite a key held by a different class — is deliberately *not* proposed here: the five sanctioned actor shadows (`as_VultronPerson` et al.) rely on overwrite-by-import-order, and the sanctioned list lives in the test, not in production code. Moving that policy into `__init_subclass__` is a VM-01-008 design change and wants its own issue if desired.

Found while syncing #3575 with `main`; fixed there.

sei-ahouseholder:
- Closes #2553
- Closes #3257
- Closes #3592

## Summary

An object's time is now carried, never minted by whoever receives or renders it.
Nested inbound objects keep the sender's `published`/`updated` as received, and
an omission stays `None`. A message is refused at parse only when a decision
needs a time that is missing (new spec item CLP-15-007).

## Changes

- **`vultron/wire/as2/parser.py`** (#3257): `_absent_times_as_none` reads an
  omitted clock-defaulted timestamp as `None` before validation, at the top
  level and at every nested inline object. Where the time is required, the
  message is refused at parse with a 422:
  - a case with no `genesisHash` and no `published` (CLP-08-002);
  - an embargo with no `endTime`;
  - a ledger entry with no `published`/`receivedAt`.
- **`vultron/wire/as2/extractor/_builders.py`** (#3257):
  - `_get_timestamp` returns `None` instead of the receiver's clock.
  - Every builder now passes timestamps through. Before, four builders
    (participant, note, case status, participant status) dropped even real
    sender values.
  - The case builder keeps the sender's `genesisHash`. Before, core recomputed
    it from a minted time, so each receiver got a different genesis.
  - The ledger-entry builder keeps the CaseActor's `received_at`/`published`.
    `received_at` is hashed content (CLP-02-008), so a replica must not
    re-stamp it.
- **`vultron/core/models/base.py`, `embargo_event.py`, `case.py`,
  `case_ledger_entry.py`**:
  - `CoreObject.published`/`updated` and embargo `start_time` accept `None`.
    Locally authored objects still get the local clock.
  - Case genesis is minted only when `published` is omitted, i.e. local
    authoring.
  - The ledger entry envelope's `published` stays required (CLP-14-002).
- **`vultron/core/models/_helpers.py`**: new `most_recent_status`, shared by the
  core and wire `current_status`.
  - Ties now go to the status appended last. A bare `max` kept the first one, so
    a status carrying the same time (or none) as the one before it never became
    current.
  - This made one #2979 test use the real append order: the seeded status
    first, the received status later.
- **`vultron/core/sync_helpers.py`** (#2553): ledger dedup now ignores only the
  snapshot's top-level build stamp.
  - Embedded objects carry their own time now, so a nested timestamp that
    differs is a different assertion.
  - Measured before the change: only the synthetic restamp-everything test
    depended on the recursive drop.
- **`test/architecture/test_vocab_registry_keys.py`** (#3592): new
  `test_no_test_class_owns_a_wire_type_key`. A test-body stand-in declaring
  `type_ = "VulnerabilityCase"` takes the real `WIRE_TYPE_MAP` key for the whole
  process; `main` guarded the two known stand-ins (#3590), but the ratchet meant
  to catch a third cannot see them — its subclass walk is scoped to
  `vultron.wire.as2.vocab` modules. This one reads the live registry instead.
- **`specs/case-ledger-processing.yaml`**: new CLP-15-007, "nested timestamps
  are carried as received; a decision that needs an absent one refuses at
  validation".
- **`notes/case-ledger-authority.md`, `AGENTS.md`**: trap 4 and the case-ledger
  row describe the new rule.

## Verification

- New failing-first tests:
  - `test/wire/as2/test_nested_timestamps.py`: 45 tests, covering every nested
    kind × omitted/blank/null, genesis, embargo, and ledger entry.
  - Tie-break tests in `test/core/models/test_case.py` and
    `test/wire/as2/vocab/test_vulnerability_case.py`.
  - `TestEquivalentRecordedEntry::test_a_different_carried_time_is_a_new_assertion`.
- Regression pin (already passed on main): render across a clock tick is
  deterministic (#2553).
- Full suite including demo and integration (`-m ""`): 11103 passed. After
  syncing main, the unit suite passed again (9820).
- Black, flake8, mypy, pyright and spec-lint are clean.


sei-ahouseholder (commented):
## PR Triage: #3575 — fix: carry object timestamps as received, never minted by the receiver

**Linked issues**: #2553 (wire render mints published/updated instead of carrying them), #2992 (AGENTS.md WIRE_TYPE_MAP key naming), #3257 (nested object with no claimed published time gets the receiver's clock)
**Changed files**: 24 files — wire/as2, wire/as2/extractor, core/models, core/sync, specs, notes
**CI status**: ✅ passing
**Base branch**: main

---

### Findings

|---|---|---|---|---|
| phase12-merge-conflict-0 | merge-state | ❌ FAIL | Branch conflicts with `main`; execute must sync and resolve | fix-now |
| phase8-replica-rebuild-mints-timestamps-0 | code-review | ❌ FAIL | `exclude_none=True` drops a carried `None` time from the snapshot, so a replica rebuilding the object mints its own clock — violates this PR's own CLP-15-007 | fix-now |
| phase6-missing-adr-time-is-carried-0 | adr-check | ⚠️ IMPROVE | No ADR records the carry-vs-mint decision that #2553 AC1 asks to "decide and record" | fix-now |
| phase8-inline-dict-without-type-mints-1 | code-review | ⚠️ IMPROVE | An inline dict with no `type` skips `_absent_times_as_none` and gets minted times | fix-now |
| phase8-extractor-fallbacks-mint-2 | code-review | ⚠️ IMPROVE | Three extractor fallbacks build core objects without passing timestamps | fix-now |

**Total**: 2 FAIL · 3 IMPROVE · 0 NEW-ISSUE

---

### Notes on the two FAILs

**Replica rebuild still mints (`phase8-...-0`).** The parser path is fixed, but the
storage path is not. `build_activity_payload_snapshot` and
`_inline_snapshot_reference_value` dump with `exclude_none=True`, so a carried
`published=None` leaves no key at all; the replica effect nodes
(`participant_status_effect.py:241`, `note_effect.py:139`,
`offer_report_effect.py:173`) rebuild with `model_validate`, and
`default_factory=now_utc` fills the gap with *that replica's* clock. Verified
locally: a `CaseStatus(published=None)` dumped with `exclude_none` and
re-validated comes back stamped with the current time.

The consequence is the same divergence the PR set out to remove: the CaseActor
stores a received status with no time (sorts last), a replica replaying the entry
stamps its own time (sorts newest), and the two disagree on current status. The
new CLP-15-007 says the receiver must not substitute its own clock "whether at
parsing, extraction, storage or rendering" — so this is the PR's own requirement
going unmet on the storage leg, not an excursion.

**Missing ADR (`phase6-...-0`).** Per `notes/specs-vs-adrs.md`, an ADR is
warranted when a meaningful alternative was evaluated and rejected *and* the
decision produces recurring testable requirements — both hold here. #2553's first
acceptance criterion is "Decide and record where object time lives," naming the
rejected option explicitly (the render keeps minting and every consumer treats
the values as non-semantic). Suggested title: *An Object's Time Is Carried, Never
Minted by the Receiver*.

### Checked and clean

- **None-safety across `vultron/`**: every use of `.published`, `.updated`,
  `.start_time` and `received_at` is None-safe — `append_case_status` guards,
  `as_EmbargoEvent.set_name` calls `isoformat()` only behind a truthiness check,
  genesis hashing refuses a `None` published, and the ledger entry's
  `received_at`/`published` stay required so the hashed-content path is safe.
- **Parser depth and aliases**: `_absent_times_as_none` runs at every depth for
  typed inline objects and honours `receivedAt`/`startTime`/`endTime` aliases.
- **Tie-break**: `max` over `(recency_key, index)` does return the last-appended
  status among equals.
- **Local authoring** still takes the local clock; `default_factory` is retained.
- **AGENTS.md** is 400 lines on both this branch and `main` — no ratchet risk.
- **#2992** doc fix is correct in `notes/vocabulary-registry.md` and `notes/README.md`.

---

*Triage artifact: `.claude/pr-3575-triage.json`*

sei-ahouseholder (commented):
## ⛔ pr-execute Gate 2 — inversion-halt

Syncing this branch with `main` surfaced a **premise inversion**, not a textual conflict. Per `pr-execute` Phase 3, I have not acted on the new premise and have not filed anything. The merge itself is resolved and committed locally; **nothing has been pushed.**

### The overturned premise

This PR's third closing reference, `Closes #2992`, rests on a factual claim:

> `AGENTS.md` line 302 is wrong — `WIRE_TYPE_MAP` is keyed by the **class name minus `as_`**, not by the emitted `type_` value. The AS2 actor `type_` keys exist only because `vultron_actor.py` registers them explicitly.

That was true of the code when this branch was cut. It is no longer true. `main` has since landed **VM-01-008 / #2982**, which decided the *code* was the thing that was wrong and changed `as_Base.__init_subclass__` to key `WIRE_TYPE_MAP` by the **emitted `type` value** (`wire_type_value(cls)`), with `_wire_type_alias` suppressing a duplicate claim:

```python
# WIRE_TYPE_MAP answers "which class does this inbound `type` value
# deserialize to?", so its key is the emitted `type` value — not the
# class name ... (VM-01-008, issue #2982)
if not is_wire_type_alias(cls):
    WIRE_TYPE_MAP[wire_type_value(cls)] = cls
```

So the two branches fixed the *same* discrepancy in *opposite directions*: this PR corrected the documentation to match the code; `main` corrected the code to match the documentation. `main`'s `AGENTS.md` line now reads "`WIRE_TYPE_MAP` on the emitted `type` *value* … (VM-01-004/008)" — i.e. the sentence #2992 asked to change is now accurate as originally written.

### What this breaks, concretely

On the merged tree, this PR's own test asserting the old key derivation fails:

```
FAILED test_wire_type_map_key_is_class_name_not_type_value[VultronPerson-…]       KeyError: 'VultronPerson'
FAILED test_wire_type_map_key_is_class_name_not_type_value[VultronOrganization-…] KeyError: 'VultronOrganization'
FAILED test_wire_type_map_key_is_class_name_not_type_value[VulnerabilityCaseStub-…] KeyError: 'VulnerabilityCaseStub'
```

The `#2992` footprint that is now false is three files — `test/wire/as2/vocab/base/test_registry.py` (the pinning test above and a reworded docstring), `vultron/wire/as2/vocab/base/registry.py` (a `_wire_class_for_type` docstring), and `notes/README.md` (one paragraph). The conflicted `notes/vocabulary-registry.md` block was taken from `main`, which now documents VM-01-008. **None of this touches the ADR-0103 / #2553 / #3257 timestamp work**, which is self-contained and passing.

### The ask

1. Is this a real inversion — should this PR **drop `Closes #2992`** and revert those three files to `main`, on the grounds that #2982 already resolved #2992 the other way (and #2992 should be closed against #2982, not here)?
2. Or does #2992 still describe something outstanding after VM-01-008 that this PR should carry?
3. Either way: should I file anything to record that two issues (#2982, #2992) named the same discrepancy and prescribed opposite fixes?

I am not acting on the new premise unreviewed. On silence I will record `outcome: halted`, mark this PR as a draft, and stop — the timestamp work will wait rather than ship alongside three assertions `main` has falsified.

sei-ahouseholder:
## Merge-order note — this should land before #3540

Non-blocking, and nothing here asks for a change to this PR. Raising it because the interaction is invisible from inside either PR's CI.

This PR and #3540 both touch `vultron/wire/as2/parser.py` and `vultron/wire/as2/extractor/_builders.py`, but in **different functions**, so `git merge-tree` merges them cleanly:

```
Auto-merging vultron/wire/as2/parser.py
Auto-merging vultron/wire/as2/extractor/_builders.py
Auto-merging vultron/core/models/case.py
Auto-merging vultron/core/models/case_ledger_entry.py
Auto-merging vultron/core/models/embargo_event.py
Auto-merging vultron/wire/as2/vocab/base/objects/base.py
```

A clean textual merge is not evidence of behavioural compatibility. `_absent_times_as_none` applies absence handling "at the top level and at every nested inline object", and it was written when every nested inline class was a wire class. #3540 changes `_inline_vocab_class` so a nested inline object may resolve to a **core** class instead:

```diff
-    return cls if issubclass(cls, as_Base) else None
+    if issubclass(cls, as_Base) or obj_type in WIRE_TYPE_MAP:
+        return cls
+    return None
```

…backed by registrations like `WIRE_TYPE_MAP["VulnerabilityCase"] = VulnerabilityCase`. Whether absence handling still lands correctly on those is exercised by neither PR. Same shape in `_builders.py`, where this PR's pass-through builders meet #3540's changed input types. Filed as #3589 so it survives both merges rather than becoming either author's problem.

**Recommended order: this PR first.** Purely mechanical grounds — it has 1 conflict against `main`, where #3540 has 10 and is 51 commits behind (including a `modify/delete` on a test file `main` has already deleted). #3540 is landing as-is by decision and is not being gated on any of this.

One other thing worth flagging, since it lands in this PR's neighbourhood: I posted a comment on #3571 citing `ARCH-23-001` name-coincidence resolution against `WIRE_TYPE_MAP.get(type(obj).__name__)`. That citation was **wrong** and I have retracted it on the issue. `vultron/wire/as2/vocab/base/base.py:60` keys the map as `cls.__name__.removeprefix("as_")`, so a core class name is the intended key, not a coincidental collision. Noting it here because this PR is the one that made me go read the registration code.

sei-ahouseholder:
## Re: merge-order note — agreed on order, and your #3571 retraction was premature

Thanks for #3589 — the `_inline_vocab_class` interaction is real and neither PR exercises it. Agreed on the order; this PR is now merged with `main` (1 conflict, resolved) and is holding at a gate for an unrelated reason (see the inversion-halt comment above), not on anything in your note.

One correction, and it cuts the other way from your retraction:

> I posted a comment on #3571 citing `ARCH-23-001` name-coincidence resolution against `WIRE_TYPE_MAP.get(type(obj).__name__)`. That citation was **wrong** and I have retracted it on the issue. `vultron/wire/as2/vocab/base/base.py:60` keys the map as `cls.__name__.removeprefix("as_")`.

That line no longer exists. `main` landed **VM-01-008 / #2982**, and `as_Base.__init_subclass__` now keys the map from `wire_type_value(cls)` — the emitted `type` value, not the class name. `main`'s own `notes/vocabulary-registry.md` states the consequence in as many words:

> **Never resolve a core type's wire counterpart by name coincidence.** … `As2WireRenderAdapter.render()` still does the forbidden thing — `WIRE_TYPE_MAP.get(type(obj).__name__)`, a *core* class name against a wire `type`-value index — and it only ever resolves where the two strings happen to coincide.

So your original `ARCH-23-001` citation was correct and the retraction should itself be retracted. Worth un-retracting on #3571 before it becomes the cited answer there.

Two notes on why the stale read was easy to make, since both matter for #3589:

- `main` records that the five actor classes registered under their class names (`"VultronPerson"`) **until #2982**, and that removing those keys changed no behaviour — `as_VultronPerson` has no `from_core`, so the render guard rejected it either way. The lookup was already dead for those types; #2982 just stopped it from looking live.
- `#3540`'s `WIRE_TYPE_MAP["VulnerabilityCase"] = VulnerabilityCase` registers a **core** class under an emitted-`type` key by hand. Post-VM-01-008 the AGENTS.md rule in `vultron/wire/as2/vocab/AGENTS.md` is "do not hand-assign a `WIRE_TYPE_MAP` key" (derived from `wire_type_value()`, sole exception `as_Actor`). That hand-assignment may want a look as part of #3589 rather than after both land.

sei-ahouseholder (commented):
## PR Execute: #3575 — 7 findings, all resolved · CI 27/27 · MERGEABLE (CLEAN)

All 5 triage findings fixed, plus 2 findings that only existed after syncing with `main`. 2 CI-loop iterations.

### Triage findings

| Finding | Sev | Outcome | Commit |
|---|---|---|---|
| `phase8-replica-rebuild-mints-timestamps-0` | FAIL | fixed | `3448c39`, `be15630` |
| `phase12-merge-conflict-0` | FAIL | fixed | `e72ae6c`, `5219837` |
| `phase6-missing-adr-time-is-carried-0` | IMPROVE | fixed | `be15630` |
| `phase8-inline-dict-without-type-mints-1` | IMPROVE | fixed | `be15630` |
| `phase8-extractor-fallbacks-mint-2` | IMPROVE | fixed | `be15630` |

The replica FAIL was the substantive one: `exclude_none=True` drops a carried `None`, so a replica rebuilding a status from `payload_snapshot` minted its own clock and disagreed with the asserting actor about which status is current — this PR's own CLP-15-007, violated. Fixed with one shared helper (`absent_times_as_none`) called from `project_wire_snapshot_to_core` **after** the camelCase remap; calling it before would read a present `receivedAt` as an absent `received_at` and then raise `DuplicateKeySpellingError` on the disagreement.

The inline-dict IMPROVE drove an architecture change rather than a patch: the absent→`None` read moved out of `parse_activity` and onto `as_Base` as a context-gated `mode="before"` validator. The parser can only pre-treat a dict whose `type` it resolved, so an inline object omitting `type` slipped past it at any depth; validating on the class the parent actually chose fixes it everywhere and leaves timestamp-less classes untouched instead of handing them keys they refuse.

### Gate 2 — inversion, resolved by decision

`main` landed **VM-01-008 / #2982** mid-pipeline, keying `WIRE_TYPE_MAP` by the emitted `type` value — the inverse of this PR's `Closes #2992` doc fix, which asserted the key was the class name minus `as_`. Not a textual conflict: the PR's pinning test failed on the merged tree. [Ask posted](https://github.com/CERTCC/Vultron/pull/3575#pullrequestreview-5295696656); resolved as **drop #2992**.

`notes/README.md`, `vocab/base/registry.py` and `vocab/base/test_registry.py` reverted to `main`; `notes/vocabulary-registry.md` taken from `main`; `Closes #2992` and its body entries removed. #2992 is left open — it should be closed against #2982, which resolved it in the other direction. The timestamp work was untouched by any of this (`816b259`).

### Pre-existing failure, filed and fixed — #3592

The first post-sync full run had **66 failures, every one passing in isolation**. A test-body stand-in declaring `type_ = "VulnerabilityCase"` registers itself under the real `WIRE_TYPE_MAP` key, and the registry's own reference keeps it alive for the process — so later tests four directories away resolved that type to a class defined inside a test function. Reproduced on a pristine `origin/main` worktree, so pre-existing (BW-07-009); filed as #3592.

`main` fixed the stand-in independently in `9c5aab9b2` (#3590) about 20 minutes later; that half came in on the second sync. What this PR retains is the half `main` doesn't have: `test_no_test_class_owns_a_wire_type_key`. The ratchet written for this exact failure mode — `test_no_unsanctioned_wire_type_collisions` — structurally cannot see it, because `_registrable_wire_classes()` scopes its subclass walk to `vultron.wire.as2.vocab` modules and filters test classes out. The new check reads the live registry instead, so module of origin cannot hide a squatter. Verified it fails, naming the offending class, with `9c5aab9b2` reverted.

Two claims in that issue's body were wrong and are [corrected in-thread](https://github.com/CERTCC/Vultron/issues/3592#issuecomment-5802618853): the stand-in fix was already on `main` when I filed (my ref was stale), and "no test workflow runs on `main` pushes" is false — `python-app.yml` does run the full suite on `main`, it failed at 18:50:39Z, and that is what surfaced this. The issue is rescoped to the ratchet gap only.

### Verification

- Unit: **9840 passed**, 446 skipped, 21 xfailed (exit 0)
- Full incl. integration + demo (`-m ""`): **11173 passed**, 446 skipped, 22 xfailed (exit 0)
- black, mypy (1457 files), pyright (0 errors), spec-lint, `mkdocs build --strict` — all clean. flake8 clean on changed files; the 3 `scripts/` findings are pre-existing and present on `main`.
- **CI: 27/27 passing.** Merge state: **MERGEABLE (CLEAN)** against `main`.

Artifact: `.claude/pr-3575-execute.json`

sei-ahouseholder (commented):
## PR Verify: #3575 — READY-TO-MERGE

**Merge state: `MERGEABLE` (`CLEAN`) against `main`** — live re-check, not the artifact's reading. Not a draft, no `needs-rebase` label. No conflict markers anywhere in the tree (all 3 resolved paths swept).

**CI: 27/27 checks SUCCESS**, nothing non-skipped outside SUCCESS. `integration_tests_run: true` is satisfied by green jobs, not just unit tests: 4 Demo Integration, 4 Invariant Harness, Spec Corpus Tests and Tests (pytest) are all SUCCESS.

Artifact integrity: 5 results for 5 triage findings, ids matching.

### Triage findings — verified at HEAD, not just at the commit

| Finding | Sev | Outcome | Verdict | Evidence at HEAD |
|---|---|---|---|---|
| `phase8-replica-rebuild-mints-timestamps-0` | FAIL | fixed | **CONFIRMED** | `_helpers.py:241` defines `absent_times_as_none`; called at `:311`/`:327`; `participant_status_effect.py:248` rebuilds through `project_wire_snapshot_to_core`; pin present at `test_participant_status_effect.py:439` |
| `phase12-merge-conflict-0` | FAIL | fixed | **CONFIRMED** | `notes/vocabulary-registry.md` byte-identical to `main`; `base.py:17` import block reconciled with `main`'s `__init_subclass__` at `:70` **and** this PR's `INBOUND_CONTEXT_KEY` at `:93`; `test_cases.py:230`/`:267` both guarded |
| `phase6-missing-adr-time-is-carried-0` | IMPROVE | fixed | **CONFIRMED** | ADR-0103 present, listed in `docs/adr/index.md` and `mkdocs.yml` nav; cited at `case-ledger-processing.yaml:1562` |
| `phase8-inline-dict-without-type-mints-1` | IMPROVE | fixed | **CONFIRMED** | `parser.py:50` `_INBOUND_CONTEXT`, passed at both `model_validate` sites (`:138`, `:254`); validator on `as_Base` |
| `phase8-extractor-fallbacks-mint-2` | IMPROVE | fixed | **CONFIRMED** | 9 `_get_timestamp(obj, …)` call sites in `_builders.py` |

Every commit ref (`be15630`, `816b259`, `5219837`) is an ancestor of HEAD.

### Post-sync findings — outside the triage set, verified the same way

Neither existed when triage closed its list; both were produced by merging `main` mid-pipeline. Recorded under `post_sync_findings` so the integrity count above stays honest.

| Finding | Gate | Outcome | Verdict | Evidence at HEAD |
|---|---|---|---|---|
| `sync-wire-type-map-premise-inverted-0` | Gate 2 inversion | fixed (per your call) | **CONFIRMED** | `notes/README.md`, `vocab/base/registry.py`, `vocab/base/test_registry.py` all byte-identical to `main`; no `#2992` reference remains in `vultron/`, `test/`, `notes/`, `specs/` or `docs/adr`; `Closes #2992` removed from the body |
| `sync-test-class-hijacks-wire-type-key-1` | `fix-now-file` → #3592 | fixed | **CONFIRMED** | `test_vocab_registry_keys.py:234` `test_no_test_class_owns_a_wire_type_key`; stand-in guard taken from `main`'s `9c5aab9b2` |

### Open items for you — neither blocks the merge

- **#2992 is still OPEN**, deliberately. This PR no longer claims it. It was resolved in the opposite direction by **#2982 / VM-01-008**, so it should be closed against that, not this. Worth a glance at whether anything else cited the old premise.
- **#3592 is still OPEN** and is closed by this PR's body. It was rescoped after filing: the stand-in guard half landed on `main` independently (#3590), so what remains is the ratchet blind spot. Two claims in its original body were wrong and are corrected in-thread rather than edited away — notably "no test workflow runs on `main` pushes", which is false.

### Verdict

```
Overall verdict: READY-TO-MERGE
Merge state:     MERGEABLE (CLEAN) — base main
CI status:       passing (27/27)
```

Session artifacts cleaned up.

